PR #1006, opened 2014-01-15
Make sure CSS box-sizing: border-box has -moz version too

unwiredben:
In testing Moonstone on Firefox, I noticed a lot of buttons with the
wrong size.  This turned out to be a problem where Firefox hasn't yet
implemented a non-prefixed version of the CSS box-sizing rule.
We had some places where we had the -moz version, but the usage wasn't
consistent in Moonstone. I noticed that we had a .box-sizing CSS
selector, so I used Less's mixin support to replace all the explicit
border-box uses with that rule.

Enyo-DCO-1.1-Signed-Off-By: Ben Combee (ben.combee@lge.com)

### DIFF
--- a/css/ContextualPopup.less
+++ b/css/ContextualPopup.less
@@ -10,9 +10,7 @@
 	background-clip: padding-box;
 	-webkit-background-clip: padding-box;
 	-moz-background-clip: padding-box;
-	box-sizing: border-box;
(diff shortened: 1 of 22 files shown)


webOS101:
Thanks for the work, sadly, so out of date we can't merge easily!

PR #1333, opened 2014-05-26
BHV-4132 : Perf: Analyze create/render performance of moon.ExpandablePic...

otisnate:
...ker

Approach:
- canCreateItems flag is added to decide when picker creates child components
- addItem function should be called instead of createComponent

DCO-1.1-Signed-Off-By: SoonGil Choi soongil.choi@lge.com


lishammel:
@otisnate  - the jira this is tied to is closed.  Is this PR needed anymore or should it be closed also? thanks.

lishammel:
@otisnate  - is this PR still needed? Thanks.

PR #1500, opened 2014-08-01
Completed changes from ENYO-40

clinuz:
Second attempt from @webOS101 for ENYO-40.


lishammel:
@webOS101  and @clinuz  - the Jira this is tied to is closed. Do these still need to be merged? thanks.

webOS101:
I think we merged a different commit for it.

PR #1525, opened 2014-08-11
BHV-11870: Support moon.Item Icon Overlay

KunmyonChoi:
Adding moon.ItemOverlay kind to support common item pattern.

Item is composed of label and icons
Label should have marquee
Icons can be on the left/right side of label
Icons can be shown on focus and auto hide
We can avoid use of fittable inside of item by using this new kind.

### DIFF
--- a/css/Item.less
+++ b/css/Item.less
@@ -23,3 +23,27 @@
 .enyo-locale-non-latin .moon-item {
 	.enyo-locale-non-latin .moon-sub-header-text;
 }
+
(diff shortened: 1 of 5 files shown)


lishammel:
@KunmyonChoi  - this Jira has been closed. Should this PR be closed out also? Thanks.

KunmyonChoi:
Hi @lishammel , I don't think we need merge this in upkeep. Thanks for pointing this out. :)

PR #1553, opened 2014-08-26
ENYO-493: Duration Picker (hour, minute, second)

anishlg:
Duration Picker Widget is completed and uploaded. countdown feature is implemented.
please review the code


anishlg:
Duration Picker Widget has updated with the review comments.
1. Used single value property than using hour, minute and seconds values separately.
2. Min/max values are removed as it's always constant.
3. Combined DurationPickerBase and DurationPcker into a single file as there won't be any future sub kinds for DurationPcker 
4.Made the template property published.
5. Removed DurationPickerBase.openChanged method as it doesn't do anything
6. Updated the sample.

please review the code again

anishlg:
new PR request is generated 
https://github.com/enyojs/moonstone/pull/1978

hence closing this request

PR #1559, opened 2014-08-29
BHV-13593: Undefined Error in moon.Checkbox.

srinivasv123:
Issue:
If we create ''moon.FormCheckBox" in the group control. the dynamic
operations like create and destroy is causing exception.
Fix:
performing a check in 'decorateActivateEvent' for the current
'moon.Checkbox' component.

Enyo-DCO-1.1-Signed-off-by: Srinivas V srinivas.v@lge.com

### DIFF
--- a/source/CheckboxItem.js
+++ b/source/CheckboxItem.js
@@ -236,7 +236,8 @@
 		*/
 		decorateActivateEvent: function (inSender, inEvent) {
 			inEvent.toggledControl = this;
-			this.setChecked(this.$.input.getChecked());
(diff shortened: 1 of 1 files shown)


lishammel:
@srinivasv123  - the Jira this is tied to is closed. Should this PR be closed also? thanks.

lishammel:
@srinivasv123  - I think this can be closed out since the jira was resolved another way? Thanks.

anishramesan:
Closed as discussed with @srinivasv123

PR #1622, opened 2014-09-17
BHV-14524: MoonTooltip-multiple tooltips showing issue

srinivasv123:
Issue: when focus shifts from one tooltip decorator to other, previous
tooltip doesn't hide. this is causing multiple tooltip show.

Fix: Before showing any tooltip, handling the previous tooltips if there
are any.

DCO-1.1-Signed-Off-By: Srinivas V  srinivas.v@lge.com

### DIFF
--- a/source/Tooltip.js
+++ b/source/Tooltip.js
@@ -94,6 +94,16 @@
 			*/
 			contentUpperCase: true
 		},
+		statics: {
+			lastTooltip: null,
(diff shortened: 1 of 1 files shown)


lishammel:
@srinivasv123  - should this be closed out?  I think the jira is resolved another way? Thanks.

anishramesan:
Closed as discussed with @srinivasv123.

PR #1760, opened 2014-11-08
ENYO-442: Utilize method for temporarily suppressing events.

aarontam:
### Issue
We are currently using a pair of `silence` and `unsilence` calls to temporarily suppress events in situations where it is either required (to prevent undefined object errors) or helpful (to prevent unnecessary execution of change handlers). We've added a `muzzle` function that can be utilized now.

### Fix
`moon.DatePicker` has been updated to use the `muzzle` function, while we've added the use of `muzzle` for `moon.TimePicker` to prevent unnecessary execution of change handlers.

### Notes
This PR is dependent on https://github.com/enyojs/enyo/pull/922.

### DIFF
--- a/source/DatePicker.js
+++ b/source/DatePicker.js
@@ -183,42 +183,42 @@
 			}
 
 			values = this.calcPickerValues();
-			this.silence();
 
-			for(f = 0, l = doneArr.length; f < l; f++) {
-				o = doneArr[f];
+			this.muzzle(function () {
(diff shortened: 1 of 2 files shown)


lishammel:
@aarontam  - should we close this PR since the Jira is closed?  Thanks

aarontam:
@lishammel Yep let's do that. Closing for now as this is not needed.

PR #1849, opened 2014-12-19
ENHYO-698: ContextualPopup: Spotlight Does Not Spot Parent upon 'Select Dismiss On'

dmikey:
Issue.

When using Dismiss on Select, the popup activator was not re-spotted because `hide` was used instead of `closePopup`

Fix.

Switch the method being used to close popup.

### DIFF
--- a/samples/ContextualPopupSample.js
+++ b/samples/ContextualPopupSample.js
@@ -224,7 +224,7 @@ enyo.kind({
 		this.$.buttonPopup.setAutoDismiss(!inSender.getActive());
 	},
 	dismissRadioSelection: function(){
-		if(this.$.nestedRadioDismissButton.value) this.$.nestedRadioPopup.hide();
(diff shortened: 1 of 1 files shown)


samples/ContextualPopupSample.js
```diff
@@ -224,7 +224,7 @@ enyo.kind({
 		this.$.buttonPopup.setAutoDismiss(!inSender.getActive());
 	},
 	dismissRadioSelection: function(){
-		if(this.$.nestedRadioDismissButton.value) this.$.nestedRadioPopup.hide();
+		if(this.$.nestedRadioDismissButton.value) this.$.nestedRadioPopup.closePopup();
```
dmikey: Makes sense, if UX deems that the activator should always be hidden, we could make the change to `hide`. I do wonder if any app developers have already required use of the internal closePopup though, and if changing the default behavior for `hide` could have affects on apps currently running (possibly not?).

dmikey:
Closed in favor of https://github.com/enyojs/moonstone/pull/1895

PR #1873, opened 2014-12-29
ENYO-720: Popup in video slider has wrong height

KunmyonChoi:
### Issue: 
Slider popup canvas and label are using different unit of height. This make round canvas and label had different height.

### Fix:
Remove use of enyo.dom.unit on popupLabel, popup and popup offset height value.


lishammel:
@KunmyonChoi  - Jira ENYO-720 is closed. Is this PR still needed? Thanks.

KunmyonChoi:
@lishammel I don't think this one is needed any more.

PR #1887, opened 2015-01-02
Fixes ENYO-719

ryanjduffy:
## Issue
moon.Panel.transitionFinished is not called by moon.Panels

## Fix
Ensure that events and moon.Panel method calls occur at the same time
in the lifecycle regardless of moon.Panels animate state.

## Notes
Depends upon enyojs/layout#110

Issue: ENYO-719

### DIFF
--- a/source/Panels.js
+++ b/source/Panels.js
@@ -162,7 +162,7 @@
 				{name: 'client', tag: null}
 			]},
 			{name: 'showHideHandle', kind: 'moon.PanelsHandle', classes: 'hidden', canGenerate: false, ontap: 'handleTap', onSpotlightLeft: 'handleSpotLeft', onSpotlightRight: 'handleSpotRight', onSpotlightFocused: 'handleFocused', onSpotlightBlur: 'handleBlur'},
-			{name: 'showHideAnimator', kind: 'enyo.StyleAnimator', onComplete: 'animationComplete'}
+			{name: 'showHideAnimator', kind: 'enyo.StyleAnimator', onComplete: 'showHideAnimationComplete'}
 		],
 
 
@@ -214,13 +214,6 @@
 		_initialTransition: true,
 
 		/**
-		* Flag for panel transition.
-		*
-		* @private
-		*/
-		transitionInProgress: false,
-
-		/**
 		* Flag for blocking consecutive push/pop/replace panel actions to protect
 		* create/render/destroy time.
 		*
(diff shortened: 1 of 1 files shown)


source/Panels.js
```diff
@@ -673,19 +666,6 @@
 		},
 
 		/**
-		* Sets the index of the active panel, skips animation.
-		*
-		* @param {number} index - Index of the panel to make active.
-		* @public
-		*/
-		setIndexDirect: function(inIndex) {
```
otisnate: *setIndexDirect* is public API and should not be removed

ryanjduffy: It's not removed; it's inherited from enyo.Panels

KunmyonChoi:
Hi, we are want to minimize naming change while in production phase. Even internal function and variable, developers are referencing these to implement specific features and avoid production problems by their own. So, we encourage refactoring after production phase. :)

graynorton:
Hi, @KunmyonChoi, @otisnate -

The problem in this specific case is that we couldn't find a way of addressing ENYO-719 that didn't involve refactoring.

My suggestion is that we take this fix into master and 2.5-upkeep, but that we stop releasing from 2.5-upkeep into the webOS branch that is currently being used for production. We should discuss this in the context of our overall branching strategy.

KunmyonChoi:
@graynorton , I found that smartshare is referencing transitionInProgress property directly from app.
I'll guide app to use inTransition. I agree your suggestion. In addition to that, I also see many cases app is overriding moonstone component for specific use case. This overrides make our release hard to work because it can introduce many chance to broken. So, I want to make a chance to get all the requirements from those overriding (workarounds) and covers them from framework side. How do you think?

graynorton:
@KunmyonChoi, are you suggesting that we do something like the following?

* Take an inventory of the cases where apps are currently overriding framework components
* Analyze those cases (asking app developers questions as needed)
* Identify framework feature / API enhancements that would make those overrides unnecessary
* In cases where there don't seem to be good framework enhancements, try to suggest less "dangerous" alternatives to the app developers

If that's what you mean, I think it's a great idea.

ryanjduffy:
@graynorton @KunmyonChoi :+1:

KunmyonChoi:
@graynorton , Yes Exactly. 
You already explained detailed work items. :+1:

DavidUM:
That's very nice thought! 
I wanna give two thumbs up!! :+1:  :+1:

PR #1895, opened 2015-01-06
ENYO-698: ContextualPopup: Spotlight Does Not Spot Parent upon 'Select Dismiss On' #1849

dmikey:
Issue.

When using Dismiss on Select, the popup activator was not re-spotted because hide was used instead of closePopup

Fix.

Switch the method being used to close popup.

Enyo-DCO-1.1-Signed-off-by: Derek Anderson derek.anderson@lge.com

### DIFF
--- a/source/ContextualPopup.js
+++ b/source/ContextualPopup.js
@@ -542,6 +542,10 @@
 			this.inherited(arguments);
 			this.alterDirection();
 			this.showHideScrim(this.showing);
+
+			if(this.spotActivatorOnClose) {
+				enyo.Spotlight.spot(this.activator);
+			}
 		},
 
 		/**
--- a/source/ContextualPopup.js
+++ b/source/ContextualPopup.js
@@ -543,7 +543,7 @@
 			this.alterDirection();
 			this.showHideScrim(this.showing);
 
-			if(this.spotActivatorOnClose) {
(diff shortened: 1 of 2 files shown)


source/ContextualPopup.js
```diff
@@ -542,6 +542,10 @@
 			this.inherited(arguments);
 			this.alterDirection();
 			this.showHideScrim(this.showing);
+
+			if(this.spotActivatorOnClose) {
+				enyo.Spotlight.spot(this.activator);
```
dmikey: @KunmyonChoi can you elaborate a little more on what we need to accomplish by checking focus inside the popup? I'm not understanding what the current short coming is with this solution.

Thanks!

KunmyonChoi: @dmikey , Sorry for the confusion.
When I see the popup showingChanged code, popup is checking that current focus is on popup before actually set focus. The popup showing property can be set in programming way (from luna call response handler or setTimeout handler), If focus can be on outside of popup when showingChange is called then we need to filter out these cases before change focus.

dmikey: Good catch Kunmyon. Added check for `this.showing` so spotlight isn't sent unless the popup was dismissed.

PR #1898, opened 2015-01-08
ENYO-763: Account for transformational rounding issues in scroller thumb.

aarontam:
### Issue
The size of the scroller thumb can be incorrectly transformed when its non-transformed measurements consist of fractional pixels. This can cause the visual appearance of the scroller thumb to either appear too small or too large.

### Fix
This is another case of the web engine utilizing the rounded value of an element's measurement when applying transforms. To circumvent this, we set the value of the to-be-transformed dimension (width or height) to its rounded value at render time, so that subsequent computations will align with the value that the transformation will ultimately utilize. Additionally, we preemptively scale several pixel values to their appropriate values via `moon.riScale`. Note that simply retrieving and utilizing the actual, fractional values in the `generateMatrix` function (using `getBoundingClientRect` instead of `offsetHeight` and `offsetWidth`) to generate the transformation matrix is not sufficient as the rounded value will be utilized when the transformation is applied.

### DIFF
--- a/source/Thumb.js
+++ b/source/Thumb.js
@@ -66,6 +66,21 @@
 		/**
 		* @private
 		*/
+		rendered: enyo.inherit(function (sup) {
+			return function () {
+				sup.apply(this, arguments);
+
+				// Rounding to the nearest whole pixel in the dimension that will be transformed, to
+				// prevent rounding issues when transforms are applied.
+				var node = this.hasNode(),
+					rect = node && node.getBoundingClientRect();
+				if (rect) this.applyStyle(this.dimension, Math.round(rect[this.dimension]) + 'px');
+			};
+		}),
+
+		/**
+		* @private
+		*/
 		update: function (inStrategy) {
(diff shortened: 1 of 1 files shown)


source/Thumb.js
```diff
@@ -83,11 +98,11 @@
 				}
 				var sbo = inStrategy[this.positionMethod]() - over;
 				// calc size & position
-				var bdc = bd - this.cornerSize;
+				var bdc = bd - moon.riScale(this.cornerSize);
```
ryanjduffy: Do these (`cornerSize` and `minSize`) need to happen on `update()` or can they be precalculated on `create()`?

aarontam: Good call - these are not changed internally and are private, so they shouldn't be changed by the user after `create` time - I'll make an update.

aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/1908.

PR #1900, opened 2015-01-08
Fixes ENYO-736

ryanjduffy:
## Issue
Since the base size is 1080p, values converted to 720p are multiplied
by 2/3 which results in sub-pixel values for many properties. The
effect is that the browser may alter the dimensions of a node based
on the accumulation of the fractional part of box model property
values.

## Fix
This mixin rounds the value to the lower multiple of 3 so the
conversion results in a round integer value thereby avoiding the
fractional accumulation issue

### DIFF
--- a/css/Item.less
+++ b/css/Item.less
@@ -2,7 +2,7 @@
 .moon-item {
 	.moon-sub-header-text;
 	line-height: @moon-item-line-height;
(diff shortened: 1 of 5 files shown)


ryanjduffy:
Closing in favor of a future PR converting the `px` values to multiples of 3 thereby eliminating the need for the mixin

PR #1907, opened 2015-01-13
Patches ENYO-800

ryanjduffy:
## Issue
SimplePicker fires its onChange event on it rendered instead of when, you know, it changes.

## Fix
Since the purpose of calling it was to adjust the position of the client based on the calculated size of the picker controls, we can refactor that into another method and call that instead of selectedIndexChanged.

## Notes
This is a patch for the issue which is resolved correctly by ENYO-736 and #1906. If those are merged, this can be closed.


ryanjduffy:
Not required

PR #1913, opened 2015-01-15
ENYO-811: Use multi-res hash for image sources, update classname for moon.ImageMultiRes.

aarontam:
### Issue
`moon.ImageMultiRes` had the `moon-image` class applied, which resulted in the class being applied to both this control and its parent control `moon.Image`, resulting in the margin amount being compounded.

### Fix
We update the images to use the multi-res hash and change the class name to `moon-image-multires`.

### DIFF
--- a/samples/ImageBadgeSample.js
+++ b/samples/ImageBadgeSample.js
@@ -5,25 +5,51 @@ enyo.kind({
 	components: [
 		{kind: "moon.Scroller", fit: true, components: [
 			{kind: "moon.Divider", content: "Image Badges:"},
-			{kind: "moon.Image", src: "http://placehold.it/340x360&text=Image+One", alt: "Image One", components: [
+			{kind: "moon.Image", src: "http://placehold.it/342x360&text=Image+One", alt: "Image One", components: [
+				{kind: "moon.Icon", icon: "skipbackward"},
+				{kind: "moon.Icon", icon: "play"},
+				{kind: "moon.Icon", icon: "skipforward"},
+				{kind: "moon.Icon", icon: "search", classes: "float-right"}
+			]},
+			{kind: "moon.Image", src: {
+				"hd" : "http://placehold.it/228x240&text=Image+Two",
+				"fhd": "http://placehold.it/342x360&text=Image+Two"
+			}, alt: "Image Two", components: [
 				{kind: "moon.Icon", icon: "check"},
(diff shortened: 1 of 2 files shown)


samples/ImageBadgeSample.js
```diff
@@ -5,25 +5,37 @@ enyo.kind({
...
+					"fhd": "http://placehold.it/342x360&text=Image+One"
+				}, alt: "Image One", showBadgesOnSpotlight: true, components: [
 					{kind: "moon.Icon", icon: "check"},
 					{kind: "moon.Icon", icon: "closex"},
 					{kind: "moon.Icon", icon: "drawer", classes: "float-right"}
 				]}
 			]},
 			{kind: "moon.Item", components: [
-				{kind: "moon.Image", src: "http://placehold.it/180x240&text=Image+Two", alt: "Image Two", showBadgesOnSpotlight: true, components: [
+				{kind: "moon.Image", src: {
```
otisnate: @aarontam I think we can remain at least one sample to use string in src property instead of array

aarontam: Hi @otisnate, I think a better place to demonstrate utilizing a string for the `src` property of `moon.Image` is the `ImageSample`, which we're currently implementing/updating. @Djspaceg, what do you think?

Djspaceg: To help get the point across about the new functionality, I'm not against the idea, and it may even be prudent to include an additional image in this sample that uses just a single string src. I'd hate for developers to think that you could *only* use the hash form.

aarontam: Agreed - I guess I'm kind of torn as we had made changes to other samples to utilize a hash for the `src` property and I'm wondering if we should also re-add demonstrations of the single string `src` to those samples. We had made changes in these samples (including this PR) specifically to address issues with image scaling at 720p, as there are other samples with instances of images where we're using the single string `src`. I think that was kind of the idea of having all of the variations demonstrated in the `ImageSample`, as that sample is specifically concerning `moon.Image`, rather than trying to exhaustively demonstrate various aspects of secondary controls in a given sample. I don't think I'm necessarily opposed to re-adding a single `src` demonstration to just this sample, though, as it is closely related to `moon.Image`.

Djspaceg: 100% agree with what you said.

PR #1920, opened 2015-01-16
ENYO-828: Adjust maximum button width, other measurement tweaks.

aarontam:
### Issue
The maximum width of `moon.Button` is currently 302px, which is not a multiple of 3.

### Fix
We adjust the maximum width of `moon.Button`, along with some other measurement tweaks aimed to promote saner measurement values.

### DIFF
--- a/css/ChannelInfo.less
+++ b/css/ChannelInfo.less
@@ -33,7 +33,7 @@
 	border-radius: 6px;
 	text-align: center;
 	white-space: nowrap;
-	padding: 3px 9px;
+	padding: 3px 12px;
 	display: inline-block;
 }
 .enyo-locale-non-latin .moon-video-player-info-icon {
--- a/css/ChannelInfo.less
+++ b/css/ChannelInfo.less
@@ -33,7 +33,7 @@
 	border-radius: 6px;
 	text-align: center;
 	white-space: nowrap;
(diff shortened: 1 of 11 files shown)


ryanjduffy:
Why were the other values changed? They all appear to be %3.

aarontam:
@ryanjduffy restored a few measurements to their original %3 values, wanted to use saner values for `moon.ImageItem`.

css/Icon.less
```diff
@@ -7,7 +7,7 @@
 	background-repeat: no-repeat;
 	display: inline-block;
 	vertical-align: middle;
-	margin: 9px 15px;
+	margin: 12px 12px;
```
Djspaceg: This change makes me happy.

css/ChannelInfo.less
```diff
@@ -33,7 +33,7 @@
 	border-radius: 6px;
 	text-align: center;
 	white-space: nowrap;
-	padding: 3px 9px;
+	padding: 3px 12px;
```
Djspaceg: Can't convert this one to 12px. it's too wide to allow the first two badges to fit on the same line, like in 1080.

aarontam: Good catch! Fixed.

PR #1922, opened 2015-01-16
ENYO-792: Adding ItemOverlay and ItemOverlaySupport (2.5-upkeep)

KunmyonChoi:
Issue:
Adding new kind to support item overlay layout.

### DIFF
--- a/source/ItemOverlay.js
+++ b/source/ItemOverlay.js
@@ -1,0 +1,163 @@
+(function (enyo, scope) {
+	/**
+	* {@link moon.ItemOverlay} is a supplementary control that helps to manage
+	* layout within a {@link moon.Item}.
+	* 
+	* ```
+	* {kind: "moon.Item", components: [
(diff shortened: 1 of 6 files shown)


source/ItemOverlay.js
```diff
@@ -0,0 +1,163 @@
...
+		}),
+
+		/**
+		* Update auto-hide class on left and right overlay.
+		*
+		* @private
+		*/
+		autoHideChanged: function() {
+			if (this.$.beginning) {
+				this.$.beginning.addRemoveClass('auto-hide', this.autoHide);
```
Djspaceg: These should be something more like `this.$.beginning.set('autoHide', this.autoHide);` to hook into the existing technique used by ItemOverlay. Same with below too.

lishammel:
@KunmyonChoi  - this Jira has been closed.  Is this PR still needed? Thanks.

KunmyonChoi:
Hi @lishammel , I don't think we need merge this in upkeep. Thanks for pointing this out. :)

PR #1926, opened 2015-01-16
ENYO-781: Add a simplified automated accessor to all samples

Djspaceg:
* Former Sample.html is now simply called "All".
* Added a barebones sample loader to ultimately replace all of the HTML files in the samples directory.
* Added a .htaccess file to tell Apache servers how to redirect requests for the (deprecated) HTML files to the new universal Sample.html
 * Included support for directory indexing and index.html detection for super simple accessing.


ryanjduffy:
Thoughts on making 'All' the default value rather than the 'No Sample Selected' message?

Djspaceg:
@ryanjduffy Yeah, we could do that. Seems friendly.

aarontam:
@Djspaceg I think we'll also need to update package.js to reflect the renaming of Sample.css to All.css.

Djspaceg:
@aarontam Whoops! I'll add/fix that!

PR #1927, opened 2015-01-16
ENYO-805: Prevent incorrect marquee distance calculation.

aarontam:
### Issue
The `scrollWidth` of a `moon.Marquee` can have an incorrect value of 0.

### Fix
This is due to the `display:inline` rule being applied to the `moon.Marquee` control. We add a comment about overriding this rule. Credit to @Djspaceg for the fix.

### DIFF
--- a/css/Marquee.less
+++ b/css/Marquee.less
@@ -1,9 +1,11 @@
 .moon-marquee {
-    width: auto;
-    text-overflow: ellipsis;
-    white-space: pre !important;
-    overflow: hidden;
-    text-align: left;
+	width: auto;
+	text-overflow: ellipsis;
+	white-space: pre !important;
+	overflow: hidden;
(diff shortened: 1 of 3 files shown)


samples/PopupSample.js
```diff
@@ -15,7 +15,11 @@ enyo.kind({
 			{kind: "moon.Button", content: "Panels in Popup", ontap: "showPopup", popup: "panelsPopup"}
 		]},
 
-		{name: "basicPopup", kind: "moon.Popup", content: "Popup..."},
```
Djspaceg: I don't think you'll want to make this change. ;)

aarontam: Haha good catch - unfortunately we'll have to wait for Travis again (facepalm).

PR #1931, opened 2015-01-20
ENYO-865: Integrate acceleration config into init.js

aarontam:
### Issue
We want to enable acceleration by default for standard builds of Moonstone.

### Fix
We integrate the existing `moon.config` logic into `init.js`.


aarontam:
Closing as this change has been cherry-picked into `2.5-upkeep` and `master`.

PR #1936, opened 2015-01-21
ENYO-875: moon.VideoPlayer remoteKey 'stop' is not stop playing

KunmyonChoi:
### Issue:
moon.VideoPlayer has a remoteKeyHandler which is calling this.jumpToStart() when 'stop' button is pressed.
But, jumpToStart is calling play when it is called while it is playing. So, it's play again after change current time to 0 and paused.

### Fix:
- Set this._isPlaying as false before calling  this.jumpToStart() to block calling play.
- Call this.sendFeedback('Stop'); after calling this.jumpToStart() to change feedback text from 'paused' to 'stop'.

### DIFF
--- a/source/VideoPlayer.js
+++ b/source/VideoPlayer.js
@@ -2115,7 +2115,9 @@
 					}
 					break;
 				case 'stop':
+					this._isPlaying = false;
 					this.jumpToStart();
+					this.sendFeedback('Stop');
 					showControls = true;
 					break;
 				}
--- a/source/VideoPlayer.js
+++ b/source/VideoPlayer.js
@@ -624,11 +624,11 @@
 						{name: 'controlsContainer', kind: 'Panels', arrangerKind: 'CarouselArranger', fit: true, draggable: false, classes: 'moon-video-player-controls-container', components: [
 							{name: 'trickPlay', ontap:'playbackControlsTapped', components: [
 								{name: 'playbackControls', classes: 'moon-video-player-control-buttons', components: [
-									{name: 'jumpBack',		kind: 'moon.IconButton', small: false, onholdpulse: 'onHoldPulseBackHandler', ontap: 'onjumpBackward'},
+									{name: 'jumpBack',		kind: 'moon.IconButton', small: false, onholdpulse: 'onHoldPulseBackHandler', ontap: 'onjumpBackward', onrelease: 'onReleaseHandler'},
 									{name: 'rewind',		kind: 'moon.IconButton', small: false, ontap: 'rewind'},
 									{name: 'fsPlayPause',	kind: 'moon.IconButton', small: false, ontap: 'playPause'},
 									{name: 'fastForward',	kind: 'moon.IconButton', small: false, ontap: 'fastForward'},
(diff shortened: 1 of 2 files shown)


aarontam:
Hi @KunmyonChoi, can you fix the Travis error? Thanks!

source/VideoTransportSlider.js
```diff
@@ -646,10 +646,12 @@
 		* @private
 		*/
 		setValue: function(val) {
-			if(Math.abs(this.calcVariationPercent(val)) > this.smallVariation) {
+			var v = this.clampValue(this.min, this.max, val);
```
aarontam: Should this be a call to `transformToVideo` instead?

source/VideoTransportSlider.js
```diff
@@ -646,10 +646,12 @@
...
+			this.currentTime = v;
```
aarontam: Please disregard the previous comment, forgot we weren't using the knob value. :)

Could we do away with the explicit clamping and move a modified version of this line to the end of the function:
    `this.currentTime = `this.value`

It seems that all code paths in this function will eventually clamp and set `this.value`.

source/VideoPlayer.js
```diff
@@ -1873,7 +1877,7 @@
 		*/
 		timeUpdate: function(sender, e) {
 			//* Update _this.duration_ and _this.currentTime_
-			if (!e && e.srcElement) {
+			if (!e && e.srcElement || e.currentTime === undefined) {
```
aarontam: Do we care about short-circuiting this function if `e.currentTime` is `null` here as well - we could relax the conditional and make it `e.currentTime == null`.

source/VideoTransportSlider.js
KunmyonChoi: No objection :) Changed.

source/VideoPlayer.js
KunmyonChoi: I found that event is sometimes comes with 'undefined' currentTime. I think null should be including typechecking because we can have time as 0.

aarontam: Ah yes, `e.currentTime == null` should evaluate to true when `e.currentTime` is either `undefined` or `null`

source/VideoTransportSlider.js
aarontam: I meant `this.value` - maybe you can verify/double-check this, but it seemed that all of the code pathways originating from `setValue` result in `this.value` being set to the clamped value of the initial value `val`, so I was thinking we could instead just set `this.currentTime = this.value` at the end of the `setValue` function. Let me know what you think, thanks!

KunmyonChoi: setValue of TransportSlider calls _setValue with value.
_setValue calls updateKnobPosition with value.
_updateKnobPosition is updating slider position with value.
But, the time is updated by currentTime which is not synced with value.
So, assigning currentTime before calling this._setValue(val); in setValue to sync.

aarontam: Thanks for the clarification - that makes sense!

PR #1941, opened 2015-01-22
ENYO-820 Implement back key support for Popup

DavidUM:
Requirement is
Close (MODAL) popup

ENYO-518-davidum has common lib for back key support.
All of sub task branches are generated from it.

### DIFF
--- a/source/keymap.js
+++ b/source/keymap.js
@@ -23,115 +23,171 @@
 		66 : 'b'
 	});
 
-})(enyo, this);
+	enyo.singleton({
+		name: 'moon.History',
+		/**
+		* Target object of back key.
+		* It is also the last object which history.pushState() is executed.
+		* When we press back key or call history.back(), it is target object.
+		*
+		* @private
+		*/
+		_currentObj: null,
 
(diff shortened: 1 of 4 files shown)


source/keymap.js
```diff
@@ -23,115 +23,184 @@
...
+		/**
+		* If "disableBackHistoryAPI" in AppInfo.json is set to true, this property
+		* should be true
+		* It should be binded with AppInfo.json. Will be replaced using enyo.LunaService
+		*
+		* @type {Bollean}
+		* @default false
+		* @public
+		*/
+		disableBackHistoryAPI: false,
```
otisnate: How about changing property name as ```enableBackHistoryAPI``` instead of ```disableBackHistoryAPI```

DavidUM: @otisnate Ya, for better understanding of developers, I agree with your opinion.
Let me fix it when I bind it with `disableBackHistoryAPI` property in AppInfo.json

DavidUM:
I will maintain a single branch for back key support

PR #1943, opened 2015-01-22
Enyo 518 Provide back key common library

DavidUM:
To support back key feature, I added a common library for moon UI controls.
It deals physical back key and history.back()

Note: There are some sub task which should be merged into Enyo-518-davidum branch before its review.
Plz review those first then see this.


source/keymap.js
```diff
@@ -11,8 +11,327 @@
...
+
+			this._dePushBackQueue();
+			return;
+		},
+
+		/**
+		* @private
+		*/
+		backKeyHandler: function (inSender, inEvent) {
+			switch (inEvent.keySymbol) {
```
aarontam: Do we anticipate handling more than one `keySymbol` value here? If not, it seems that we could remove the switch and simplify to a single conditional such as

```
if (inEvent.keySymbol == 'back' && this._currentObj && this._currentObj.getShowing())
```

source/keymap.js
```diff
@@ -11,8 +11,327 @@
...
+				item;
+			for (var i = 0; i < length; i++) {
+				item = queue.pop();
+				this._pushBackHistory(item.currentObj, item.handler);
+			}
+		},
+
+		/**
+		* Decide whether this popstate event should call handler or shouldn't.
+		*
```
aarontam: Hi @DavidUM, I think it would be helpful to document all of the various pop states with some details about what they represent.

source/keymap.js
```diff
@@ -11,8 +11,327 @@
...
+		published: {
+			/**
+			* Flag for back key handler progress
+			*
+			* @public
+			*/
+			isBackInProgress: false
+		},
+
+		events: {
```
aarontam: Do we need to fire and handle the `onPopBackHistory` event or can we just directly call the `popBackHistory` method where needed?

DavidUM: Yes, you are right. In current use case, we do not need to use event-handler model.

DavidUM:
Based on comment, I updated this PR.

DavidUM:
Currently it has conflict in moon.Panels. 
When this PR ready to merge, I will rebase it.

source/History.js
```diff
@@ -0,0 +1,389 @@
...
+		handlers: {
+			/**
+			* Hanlder for back key input.
+			* To use custom behavior for back key, you can modify this handler.
+			*/
+			onPushBackHistory: 'pushBackHistory'
+		},
+
+		/**
+		* @private
```
aarontam: In your original implementation, developers could override this as needed, per control. Do you think it would be useful to maintain that feature? if so, I think we might want to mark this as `@public`.

source/History.js
```diff
@@ -0,0 +1,389 @@
...
+		*/
+		pushBackHistory: function () {
```
aarontam: If we want to allow developers to override this in a given control (see above comment), I think we need to modify this to something like the following:
```
pushBackHistory: enyo.inherit(function (sup) {
			return function () {
				if (!sup.apply(this, arguments)) {
					moon.History.pushBackHistory(this, this.backKeyHandler || moon.History.backKeyHandler);
					return true;
				}
			};
		})
```

source/ListActions.js
```diff
@@ -678,6 +705,9 @@
 			// Skip animation before render time
 			if (!this.$.client.hasNode()) { return; }
 			this.$.client.addRemoveClass('open', this.open);
+			if (this.allowBackKey && this.open) {
```
aarontam: Instead of handling this in `moon.ListActionsDrawer`, can we handle this in `moon.ListActions`'s `openChanged` instead? It seems like we're spreading parts the back-key support between two kinds here. This would allow `moon.ListActionsDrawer` to not require any knowledge of back-key support (and the resulting one-off declaration of related properties and events), as its open/close state is effectively controlled by `moon.ListActions`.

source/History.js
```diff
@@ -0,0 +1,389 @@
...
+			return true;
+		},
+
+		/**
+		* Abstract back key handler function.
+		* Each control which mixin moon.History should implement this method.
+		*
+		* @private
+		*/
+		backKeyHandler: function () {
```
aarontam: Should this be moved into the mixin, instead of the singleton, following a form similar to `pushBackHistory`?

source/History.js
DavidUM: That's what I find last day!
Thank you so much!

source/History.js
DavidUM: Right! 
I would like to make developers to override this method but I did't think a fancy way.
Thank you for your example code.

source/ListActions.js
```diff
@@ -216,7 +221,8 @@
 			{from: 'open', to: '$.drawer.open'},
 			{from: 'iconSrc', to: '$.activator.src'},
 			{from: 'icon', to: '$.activator.icon'},
-			{from: 'disabled', to: '$.activator.disabled', oneWay: false}
+			{from: 'disabled', to: '$.activator.disabled', oneWay: false},
+			{from: 'allowBackKey', to: '$.drawer.allowBackKey', oneWay: false}
```
aarontam: Hi @DavidUM, thanks for all the updates! I think because `moon.ListActions` fully handles the back key support now, we can remove this binding, as well as the `onPushBackHistory` event and the `allowBackKey` property from `moon.ListActionsDrawer`.

DavidUM:
Thanks @aarontam 
Rebasing makes more conflicts than merging to master.
I think if you try merge it to master, you may see two conflicts from moon.Panels and VideoPlayer.
Could you handle it?

aarontam:
Hi @DavidUM, yep no problem, I can take care of the conflcits.

aarontam:
This was squashed and merged via this commit: https://github.com/enyojs/moonstone/commit/9475ecc158a5546aa9818b5dc82d75f393b6f933. In the future I'll make sure to create a separate branch with the squashed commit and open a PR for the change, for easier tracking.

PR #1945, opened 2015-01-24
Added Firefox Support for Moonstone

Mortivor:
Added some unprefixed CSS (like "transform") in order to correctly display moonstone on Firefox (and other) Browsers. Some issues:
* "line-clamp" seems to be a webkit-only-thing, no support for Firefox (ExpandableText.less)
* customising scrollbar seems to be a webkit-only-thing, no support for Firefox (TextArea.less)
* Always-Viewing-Panels: Couldn't find where the translateX happens, so couldn't fix this :(

Enyo-DCO-1.1-Signed-off-by: Christian Lange <christian.lange.81@googlemail.com>


aarontam:
Thanks for the contribution @Mortivor! We'll take a look over this. For the panels, were you looking for something like this: https://github.com/enyojs/moonstone/blob/master/source/Panels.js#L1099?

Mortivor:
Hi Aaron,

yes, these lines caught my attention too. I tried to simply add an
"this.$.clientWrapper.applyStyle('transition', direct ? null : 'transform
0.5s cubic-bezier(0.25, 0.1, 0.25, 1)');" and an
"this.$.clientWrapper.applyStyle('transition', direct ? null : 'transform
0.5s cubic-bezier(0.25, 0.1, 0.25, 1)');" but it doesn't work. If I click
on "Item One", the panel correctly shrinks, but it doesn't move to the left.

I will have to check deeper.

css/InputHeader.less
```diff
@@ -24,15 +24,15 @@
 .moon-input-header .moon-input.moon-header-title {
 	position: static;
 }
-.moon-input-header .moon-input.moon-header-title::-webkit-input-placeholder {
+.moon-input-header .moon-input.moon-header-title::-webkit-input-placeholder, .moon-input-header .moon-input.moon-header-title::-moz-placeholder {
```
Djspaceg: On these 4 additions, could you put a newline between the comma and the period? When the lines get long they are harder to read, and not conveniently line up for comparisons. Thanks!

Djspaceg:
>  If I click on "Item One", the panel correctly shrinks, but it doesn't move to the left.

It's interesting you say this, because this is the behavior I've been seeing in my Chrome for the last 4 months, and can't reproduce anywhere else, or sometimes, not even in my Chrome. Curious...

Also, it appears the Travis failed because the moonstone CSS files were compiled without the `--ri` flag. (That's why there's so many additions/deletions in those files.)
```
$ enyo/tools/lessc.sh moonstone/css/all-package.js --ri
```

Nice work converting all of these!

Mortivor:
I added these linebreaks. I hope everything ist o.k. with my pull request, I had some trouble with merging. Although we use git at my company, this github workflow is new to me. :)

Djspaceg:
Your PR looks just fine. Thanks again. We'll be reviewing/integrating this soon.

css/Panels.less
```diff
@@ -95,7 +95,9 @@
 		content: @moon-icon-arrowlargeleft;
 		color: @moon-spotlight-text-color;
 		-webkit-transform: translate3d(0, 0, 0);
+		transform: translate3d(0, 0, 0);
 		transition: -webkit-transform 0.2s linear, opacity 0.2s linear;
```
Djspaceg: This line should likely be the following:
```
-webkit-transition: -webkit-transform 0.2s linear, opacity 0.2s linear;
```
with the prefix on the value but not the rule, it would only get picked up by the rare case where the browser supports both the prefix and non prefixed versions. (This would be an existing bug, not one you introduced.)

css/InputHeader.less
```diff
@@ -24,15 +24,18 @@
...
-.moon-input-header .moon-input.moon-header-title::-webkit-input-placeholder {
+.moon-input-header .moon-input.moon-header-title::-webkit-input-placeholder,
+.moon-input-header .moon-input.moon-header-title::-moz-placeholder {
 	color: @moon-input-header-placeholder-color;
 	margin-top:12px;
 	line-height: 1.25em;
 }
-.enyo-locale-non-latin .moon-input-header .moon-input.moon-header-title::-webkit-input-placeholder {
+.enyo-locale-non-latin .moon-input-header .moon-input.moon-header-title::-webkit-input-placeholder,
+.enyo-locale-non-latin .moon-input-header .moon-input.moon-header-title::-moz-input-placeholder{
```
Djspaceg: Petty little change. Could you add a space between the "r" and the "{": `placeholder {` and the following 2 additions as well.

Mortivor:
Incorporated some changes mentioned by Blake.

css/InputHeader.less
Mortivor: Hi Blake,

i have incorporated these, Travis CI build passed.

aarontam:
Hi @Mortivor, thanks for all the updates! I was wondering if you had explored augmenting some of the styles applied in the JS that are `webkit` prefixed i.e. https://github.com/enyojs/moonstone/blob/master/source/Panels.js#L1138. Currently the panel does not animate when showing/hiding, in Firefox, due to this. If you ran into some issues with this, no worries, just wanted to verify, otherwise we can take a quick look at this and make the requisite changes.

Mortivor:
I did notice that it does not animate in Firefox, but sadly was not able to
figure out why.
Am 09.03.2015 19:01 schrieb "Aaron Tam" <notifications@github.com>:

aarontam:
@Mortivor no problem! We made a commit to address this and have merged in all of the changes. Thanks for your contribution!

PR #1947, opened 2015-01-27
ENYO-891: Allow popup to be sized to the natural height of contents.

aarontam:
### Issue
Due to some measurement tweaks made to minimize rounding issues, the previously set height of the "Selection" popup in the DataGridListSample (via the `moon-6v` class) caused the last item to be partially obscured.

### Fix
We remove the `moon-6v` class and allow the popup to be sized to the natural height of its contents. Additionally, we correct the height of the grid rows to more closely match the adjustments we have made throughout.

### DIFF
--- a/samples/DataGridListSample.js
+++ b/samples/DataGridListSample.js
@@ -16,7 +16,7 @@ enyo.kind({
 			{kind: "moon.ToggleButton", content:"Selection", name:"selectionToggle"},
 			{kind: "moon.ContextualPopupDecorator", components: [
 				{kind: "moon.ContextualPopupButton", content:"Selection Type"},
-				{kind: "moon.ContextualPopup", classes:"moon-4h moon-6v", components: [
+				{kind: "moon.ContextualPopup", classes:"moon-4h", components: [
 					{kind: "moon.RadioItemGroup", name: "selectionTypeGroup", components: [
 						{content: "Single", value: "single", selected: true},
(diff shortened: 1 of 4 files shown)


Djspaceg:
I think this may actually indicate an issue with our `moon-Xv` set of classes. For a longer term fix, instead of this sample-specific fix, we should correct the row height to match the size of the moon.Items by setting `@moon-grid-row-height: 42px;`. This change may resolve other sizing issues as well.

aarontam:
@Djspaceg good call - for some reason I thought we hadn't updated the grid sizes and this was only due to the spotlight outset changing, but that's incorrect. It's been updated, thanks for the catch.

samples/DataGridListSample.js
```diff
@@ -16,7 +16,7 @@ enyo.kind({
 			{kind: "moon.ToggleButton", content:"Selection", name:"selectionToggle"},
 			{kind: "moon.ContextualPopupDecorator", components: [
 				{kind: "moon.ContextualPopupButton", content:"Selection Type"},
-				{kind: "moon.ContextualPopup", classes:"moon-4h moon-6v", components: [
```
Djspaceg: Are you sure you want to still drop this class? (Since it fits now.)

aarontam: Yeah it didn't seem necessary as there's no scroller in the popup - I'm open to leaving it in though.

PR #1954, opened 2015-01-30
ENYO-792: Support moon.Item Icon Overlay

aarontam:
Adding a new kind "moon.ItemOverlay" and "moon.ItemOverlaySupport" mixin with sample.

### DIFF
--- a/css/ItemOverlay.less
+++ b/css/ItemOverlay.less
@@ -1,0 +1,29 @@
+/* ItemOverlay.css */
+.moon-item .moon-item-overlay {
+	float: left;
+}
+.moon-item .moon-item-overlay.moon-item-overlay-beginning .moon-icon:first-child {
+	margin-left: 0px;
+}
+.moon-item .moon-item-overlay.moon-item-overlay-ending .moon-icon:last-child {
+	margin-right: 0px;
+}
+.moon-item:not(.spotlight) .moon-item-overlay.auto-hide {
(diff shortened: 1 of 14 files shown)


css/ItemOverlay.less
```diff
@@ -0,0 +1,31 @@
...
+		float: left;
+
+		&.right {
+			float: right;
+		}
+		&.beginning .moon-icon:first-child {
+			margin-left: 0;
+		}
+		&.ending .moon-icon:last-child {
+			margin-right: 0;
```
aarontam: Hi @KunmyonChoi, I think we need to account for this when in RTL.

source/ItemOverlay.js
```diff
@@ -0,0 +1,174 @@
...
+						this.components[0].mixins = ['moon.MarqueeItem'];
+					}
+				}
+				sup.apply(this, arguments);
+				this.createComponent({
+					name: 'beginning', kind: 'moon.ItemOverlay',
+					classes: 'beginning', right: false, addBefore: this.controlParentName,
+					components: this.beginningComponents
+				});
+				this.createComponent({
```
aarontam: Hi @KunmyonChoi, is there any reason to not use `createComponents` in this case?

KunmyonChoi: Not really. I can use createComponents here. Thanks!

PR #1955, opened 2015-01-30
ENYO-923: Stop marquee when window throw blur event

KunmyonChoi:
### Issue:
On app change state between BG and FG, div which is under cursor is not getting onleave event.
While marquee is flow on hover, it is not stop because it is not getting onleave event when app  goes to BG.

### Fix:
- Add blur on windowEvents array to handle on dispatcher feature. (Core)
- Remember last hovered marqueeSupport control on global variable. (Marquee)
- Call onleave event handler of last hovered marqueeSupport control when app is getting onblur event. (Marquee)

### DIFF
--- a/source/Marquee.js
+++ b/source/Marquee.js
@@ -296,6 +296,9 @@
 			this._marquee_isHovered = true;
 			if ((this.marqueeOnHover && !this.marqueeOnSpotlight) ||
 			(this.disabled && this.marqueeOnSpotlight)) {
+				if (this.marqueeOnHover) {
+					moon.Marquee.setMarqueeOnHoverControl(this);
+				}
 				this.startMarquee();
 			}
 		},
@@ -306,6 +309,9 @@
 		_marquee_leave: function (sender, ev) {
 			this._marquee_isHovered = false;
 			if ((this.marqueeOnHover && !this.marqueeOnSpotlight) || (this.disabled && this.marqueeOnSpotlight)) {
+				if (this.marqueeOnHover) {
+					moon.Marquee.setMarqueeOnHoverControl(null);
+				}
 				this.stopMarquee();
 			}
 		},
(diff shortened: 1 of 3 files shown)


source/Marquee.js
```diff
@@ -1108,4 +1114,26 @@
...
+		},
+
+		getMarqueeOnHoverControl: function() {
+			return this._marqueeOnHoverControl;
+		}
+	};
+
+	enyo.dispatcher.features.push(
+		function (e) {
+			if ("blur" === e.type) {
```
clinuz: You don't test to see where the event came from but blur can happen in other cases so this isn't a safe assumption that any blur would be from the window.

source/Marquee.js
```diff
@@ -1108,4 +1114,26 @@
 		style: 'overflow: hidden;'
 	});
 
+
+	moon.Marquee = {
+		_marqueeOnHoverControl: null,
+
+		setMarqueeOnHoverControl: function(oControl) {
+			this._marqueeOnHoverControl = oControl;
```
clinuz: You store a reference to a control but what happens if that control is destroyed? This would inherently keep it from being released to the GC and potentially cause other issues if you're calling methods on it after the fact.

KunmyonChoi: Aha, thanks to pointing out!

source/Marquee.js
KunmyonChoi: Yes, that's true. Maybe it&s better to get more specific event name for app state change between FG and BG.

source/Marquee.js
```diff
@@ -232,6 +232,17 @@
 		* @method
 		* @private
 		*/
+		destroy: enyo.inherit(function (sup) {
+			return function () {
+				moon.Marquee.setMarqueeOnHoverControl(null);
```
clinuz: Shouldn't it only clear the marquee control if it is the one that was set? Otherwise it is clearing it unnecessarily.

clinuz: As in, unlike when we detect 'leave' and we know we were the control, being destroyed doesn't imply the same.

source/Marquee.js
```diff
@@ -1108,4 +1125,29 @@
 		style: 'overflow: hidden;'
 	});
 
+
+	enyo.singleton({
```
clinuz: This doesn't need to be an instance calling enyo.singleton.

source/Marquee.js
KunmyonChoi: Hi @clinuz ,
I'm not sure about completely understood or not on second comment. 
The sentence structure is not familiar to me @.@
Anyway, I think I can add a condition check before setting back to null on destroy.
Do you want to add handling on a leave event also?

source/Marquee.js
KunmyonChoi: Roger!

source/Marquee.js
```diff
@@ -232,6 +243,18 @@
 		* @method
 		* @private
 		*/
+		destroy: enyo.inherit(function (sup) {
+			return function () {
+				if (_getMarqueeOnHoverControl())
```
clinuz: This absolutely must be a test -> `if (this === _getMarqueeOnHoverControl())` we can't just set it to null because it was destroyed and some control existed the `this` may not be the one that was set!

KunmyonChoi: Aha, that's absolutely right! Thanks Cole :)

PR #1961, opened 2015-02-02
ENYO-519 : add moon.DayPicker component as a Dreadlocks UX requirements

sbcho:
### Issue
There is a requirement for supporting string overlay on Expandable picker in Dreadlocks.
(Especially, it target expandable picker which can select day of the week and shows representative value )

### Solution
For resolving the requirement, make a new kind which inherited from Expandable Picker and add features which are described in ENYO-519

### DIFF
--- a/source/DayPicker.js
+++ b/source/DayPicker.js
@@ -1,7 +1,7 @@
 (function (enyo, scope) {
 	/**
 	* {@link moon.DayPicker}, which extends {@link moon.ExpandablePicker}, is
-	* a drop-down picker menu that solicits day of the week from the user. 
+	* a drop-down picker menu that solicits day of the week from the user.
 	*
 	* ```
 	* {kind: 'moon.DayPicker'}
@@ -13,7 +13,7 @@
 	*
 	* The content of representative value can be changed.
 	* ```
-	* {kind: 'moon.DayPicker', everyWeekday:'Weekdays', everyWeekend:'Weekends', everyday:'Daily'}
+	* {kind: 'moon.DayPicker', everyWeekdayText:'Weekdays', everyWeekendText:'Weekends', everyDayText:'Daily'}
 	* ```
 	*
 	* @class moon.DayPicker
@@ -47,7 +47,7 @@
 		/**
 		* @private
 		*/
-		content: moon.$L('Select days'), 
+		content: moon.$L('Select days'),
 
 		/**
 		* @private
@@ -59,10 +59,10 @@
 			* Text to be displayed when all of the day are selected.
 			*
 			* @type {String}
-			* @default 'Everyday'
+			* @default 'Every Day'
 			* @public
 			*/
-			everydayText: moon.$L('Everyday'),
+			everyDayText: moon.$L('Every Day'),
 
 			/**
 			* Text to be displayed when all of the weekday are selected.
@@ -125,7 +125,7 @@
 			var joinIndex = this.selectedIndex.join();
 
 			if (indexLength === 7) {
-				return this.everydayText;
+				return this.everyDayText;
 			} else if (indexLength === 5 && joinIndex == '0,1,2,3,4') {
 				return this.everyWeekdayText;
 			} else if (indexLength === 2 && joinIndex == '5,6') {
--- a/source/DayPicker.js
+++ b/source/DayPicker.js
@@ -1,0 +1,153 @@
+(function (enyo, scope) {
+	/**
+	* {@link moon.DayPicker}, which extends {@link moon.ExpandablePicker}, is
+	* a drop-down picker menu that solicits day of the week from the user. 
+	*
+	* ```
+	* {kind: 'moon.DayPicker'}
+	* ```
+	*
+	* When the picker is minimized, the currently selected day are
+	* displayed as subtext below the picker label. And if the picker select every weekday,
+	* subtext will be changed 'Every Weekday' automatically.
+	*
+	* The content of representative value can be changed.
+	* ```
+	* {kind: 'moon.DayPicker', everyWeekday:'Weekdays', everyWeekend:'Weekends', everyday:'Daily'}
+	* ```
+	*
+	* @class moon.DayPicker
+	* @extends moon.ExpandablePicker
+	* @ui
+	* @public
+	*/
+	enyo.kind(
+		/** @lends moon.DayPicker.prototype */ {
+
+		/**
+		* @private
+		*/
+		name: 'moon.DayPicker',
+
+		/**
+		* @private
+		*/
+		kind: 'moon.ExpandablePicker',
+
+		/**
+		* @private
+		*/
+		autoCollapseOnSelect: false,
+
+		/**
+		* @private
+		*/
+		multipleSelection: true,
+
+		/**
+		* @private
+		*/
+		content: 'Select days', 
+
+		/**
+		* @private
+		* @lends moon.DayPicker.prototype
+		*/
+		published: {
+
+			/**
+			* Text to be displayed when all of the day are selected.
+			*
+			* @type {String}
+			* @default 'Everyday'
+			* @public
+			*/
+			everyday: 'Everyday',
+
+			/**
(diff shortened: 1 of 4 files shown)


source/DayPicker.js
```diff
@@ -0,0 +1,153 @@
...
+
+		/**
+		* @private
+		*/
+		multipleSelection: true,
+
+		/**
+		* @private
+		*/
+		content: 'Select days',
```
aarontam: Hi @sbcho, I think we want to wrap this in a call to `$L()`.

source/DayPicker.js
```diff
@@ -0,0 +1,153 @@
...
+		published: {
+
+			/**
+			* Text to be displayed when all of the day are selected.
+			*
+			* @type {String}
+			* @default 'Everyday'
+			* @public
+			*/
+			everyday: 'Everyday',
```
aarontam: This and the following string properties should also be wrapped in a call to `$L()`.

source/DayPicker.js
```diff
@@ -0,0 +1,153 @@
...
+			* @public
+			*/
+			noneText: 'Nothing selected',
+		},
+
+		/**
+		* @private
+		*/
+		dayOfTheWeek: [
+			{content: "Monday"},
```
aarontam: Minor nitpicky thing - can we be consistent about double vs single quotes (preference for single quotes).

source/DayPicker.js
```diff
@@ -0,0 +1,153 @@
...
+	* {kind: 'moon.DayPicker'}
+	* ```
+	*
+	* When the picker is minimized, the currently selected day are
+	* displayed as subtext below the picker label. And if the picker select every weekday,
+	* subtext will be changed 'Every Weekday' automatically.
+	*
+	* The content of representative value can be changed.
+	* ```
+	* {kind: 'moon.DayPicker', everyWeekday:'Weekdays', everyWeekend:'Weekends', everyday:'Daily'}
```
aarontam: In the example, I think we will want to encourage developers (particularly app developers) to wrap their overridden strings with `moon.$L()` so that these strings are identified and properly localized.

source/DayPicker.js
```diff
@@ -0,0 +1,153 @@
...
+			var str = '';
+			this.selectedIndex.sort();
+			for (var i=0; i < this.selectedIndex.length; i++) {
+				if (!str) {
+					str = controls[this.selectedIndex[i]].getContent().substring(0,3);
+				} else {
+					str = str + ', ' + controls[this.selectedIndex[i]].getContent().substring(0,3);
+				}
+			}
+			if (!str) {
```
aarontam: I think we can simplify L145-L148 to:

```
return str || this.getNoneText();
```

source/DayPicker.js
```diff
@@ -0,0 +1,153 @@
...
+		multiSelectCurrentValue: function () {
+			var indexLength = this.selectedIndex.length;
+			var joinIndex = this.selectedIndex.join();
+
+			if (indexLength === 7) {
+				return this.everyday;
+			} else if (indexLength === 5 && joinIndex == "0,1,2,3,4") {
+				return this.everyWeekday;
+			} else if (indexLength === 2 && joinIndex == "5,6") {
+				return this.everyweekend;
```
aarontam: Did you mean `this.everyWeekend` here?

source/DayPicker.js
aarontam: I think we will want to add a "text" suffix to these propert names, i.e. "everydayText".

source/DayPicker.js
```diff
@@ -0,0 +1,153 @@
...
+			}
+
+			var controls = this.getCheckboxControls();
+			var str = '';
+			this.selectedIndex.sort();
```
aarontam: I *think* this can be removed as the building of the indices for `this.selectedIndex` is done by iterating through the checkbox controls sequentially. Are there any cases you can think of where the order of the checkbox controls array would be different than their visual order? I realize the implementation of `multiSelectCurrentValue` in the parent kind also sorts, so it may not be necessary there, either.

source/DayPicker.js
```diff
@@ -0,0 +1,153 @@
...
+			return function() {
+				// super initialization
+				sup.apply(this, arguments);
+
+				this.createComponents(this.dayOfTheWeek);
+			};
+		}),
+
+		/**
+		* @private
```
aarontam: Can we add a comment to the documentation for this function in `moon.ExpandablePicker` communicating that this is effectively "protected" and can be overridden by subkinds, such as `moon.DayPicker`, as it is currently documented as "private."

source/DayPicker.js
sbcho: Thanks, @aarontam! I misunderstood that I should commit a new PR for supporting multilanguage next time. I'll change it now

source/DayPicker.js
sbcho: Thanks, 'everydayText' seems to have more obvious meaning :)

source/DayPicker.js
sbcho: I see. I'll add it in example after WP-461 job

source/DayPicker.js
sbcho: It looks more clear and simple :) Thanks!

source/DayPicker.js
aarontam: Ah yes, it didn't seem to hurt to use `moon.$L()` now, and if apps begin to use this control, the text strings will be surfaced as strings to be localized. Thanks!

source/DayPicker.js
sbcho: I added it since multiSelectCurrentValue has a line {{this.selectedIndex.sort()}}(Indeed, I'm not sure what  this line want to prevent) and currently I don't call it through {{this.inherited}} or {{sup.apply}}. If this line is for preventing wrong calculation when the picker add/remove components dynamically, I'll remove it

source/DayPicker.js
```diff
@@ -0,0 +1,149 @@
...
+		published: {
+
+			/**
+			* Text to be displayed when all of the day are selected.
+			*
+			* @type {String}
+			* @default 'Everyday'
+			* @public
+			*/
+			everydayText: moon.$L('Monday'),
```
aarontam: Hi @sbcho, did you mean "Everyday" here? :smile:

source/DayPicker.js
```diff
@@ -0,0 +1,149 @@
...
+			* @default 'Nothing selected'
+			* @public
+			*/
+			noneText: moon.$L('Nothing selected')
+		},
+
+		/**
+		* @private
+		*/
+		dayOfTheWeek: [
```
ryanjduffy: Can we localize the order and content of these? We'd need to update the logic in `multiSelectCurrentValue` to account for the change in first day of week as well

aarontam: Yeah I'm wondering if, at this point, we shouldn't just make the requisite changes in this PR that we had broke out into ENYO-930 since this PR is still open and it might be better to fully complete the work in one cohesive PR. We had originally broken it into separate tickets because we had wanted to resolve ENYO-519 for the sprint, but that didn't end up happening.

source/DayPicker.js
```diff
@@ -0,0 +1,149 @@
...
+		published: {
+
+			/**
+			* Text to be displayed when all of the day are selected.
+			*
+			* @type {String}
+			* @default 'Everyday'
+			* @public
+			*/
+			everydayText: moon.$L('Everyday'),
```
graynorton: Did these strings come from UX?

This string should really be "Every Day", not "Everyday".

http://www.elearnenglishlanguage.com/blog/english-mistakes/everyday-vs-every-day/

aarontam: Hi @graynorton, I had the same question, but these were the strings specified in the ticket. I can pose this as a question there to verify that this was their intention.

source/DayPicker.js
sbcho: @ryanjduffy > Yep, I didn't consider first day of week at first time. I'll implement it after confirming requirement.

PR #1969, opened 2015-02-05
ENYO-890: Prevent running animation when no arrangement is needed.

aarontam:
### Issue
There was a delay when moving Spotlight focus between joined panels.

### Fix
We ensure that we only execute the extensive animation cycle (involving `enyo.Animator`) when an arrangement is required, so the case of joined panels should be excluded. Additionally, we store the value of `shouldAnimate` once we have calculated it for a given index change, so that it does not need to be recalculated when the `indexChanged` method of `enyo.Panels` is run.

### Note
This PR is dependent on https://github.com/enyojs/layout/pull/117

### DIFF
--- a/source/Panels.js
+++ b/source/Panels.js
@@ -699,12 +699,13 @@
 			this.toIndex = index;
 
 			this.queuedIndex = null;
+			this._willMove = null;
 
 			// Ensure any VKB is closed when transitioning panels
 			this.blurActiveElementIfHiding(index);
 
 			// If panels will move for this index change, kickoff animation. Otherwise skip it.
-			if (this.shouldArrange() && this.animate) {
+			if (this.shouldAnimate()) {
 				enyo.Spotlight.mute(this);
 				this.startTransition();
 				this.triggerPreTransitions();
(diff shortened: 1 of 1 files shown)


source/Panels.js
```diff
@@ -737,11 +738,21 @@
...
-		* Returns `true` if any panels will move in the transition from `fromIndex` to
+		* Returns `true` if the panels should animate in the transition from `fromIndex` to
 		* `toIndex`.
 		*
 		* @private
 		*/
+		shouldAnimate: function () {
+			this._willMove = this._willMove || (this._willMove == null && (this.shouldArrange() && this.animate));
+			return this._willMove;
+		},
```
graynorton: @aarontam, what you have here should work just fine, but how would you feel about this instead:

```javascript
if (this._willMove == null) {
	return (this._willMove = this.animate && this.shouldArrange());
}
else {
	return this._willMove;
}
```

It's a bit more verbose, but I think it makes the logic a bit more transparent and shouldn't be any slower to execute (in fact I think should be faster, only because we check the simple boolean this.animate before running the more expensive this.shouldArrange()).

aarontam: That sounds good to me! Especially checking `this.animate` first. :)

PR #1971, opened 2015-02-05
ENYO-958: Added aspect ratio detection and classes.

Djspaceg:
ENYO-958: Added aspect ratio detection and classes.
Caching and minor speed improvement in screen lookup functions.

### DIFF
--- a/source/moon-resolution.js
+++ b/source/moon-resolution.js
@@ -5,13 +5,14 @@
 		screenType,
 		screenTypeObject;
 
-	var getScreenTypeObject = function (name) {
-		if (name == screenType && screenTypeObject) {
+	var getScreenTypeObject = function (type) {
+		type = type || screenType;
+		if (type == screenType && screenTypeObject) {
 			return screenTypeObject;
 		}
 		var types = scope.moon.ri.screenTypes;
 		return types.filter(function (elem) {
-			return (name == elem.name);
+			return (type == elem.name);
 		})[0];
 	};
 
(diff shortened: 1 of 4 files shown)


source/moon-resolution.js
```diff
@@ -69,6 +71,10 @@
 			type = type || screenType;
 			if (type) {
 				enyo.dom.addBodyClass('moon-res-' + type.toLowerCase());
+				var scrObj = getScreenTypeObject(type);
+				if (scrObj.aspectRatioName) {
+					enyo.dom.addBodyClass('enyo-aspect-ratio-' + scrObj.aspectRatioName.toLowerCase());
```
Djspaceg: This is intentionally different than the `moon-` namespace above it, to more-future-proof the developers' work that uses this class.

graynorton: Should we consider adding enyo-res above (in addition to the moon-res)? That way devs who wanted to could move to the full future-proof version right away.

Djspaceg: Great idea. I'll add that.

source/moon-resolution.js
```diff
@@ -69,6 +71,11 @@
 			type = type || screenType;
 			if (type) {
 				enyo.dom.addBodyClass('moon-res-' + type.toLowerCase());
+				enyo.dom.addBodyClass('enyo-res-' + type.toLowerCase());
```
KunmyonChoi: Is this needs to be moon-res in upkeep branch?

Djspaceg: `moon-res-*` should remain in 2.5-upkeep because app code may already be using it, so we can't take it away. We _can_ add the future name of the class, so apps which are able to update, can update now to protect them from future changes.

PR #1978, opened 2015-02-09
ENYO-493: Duration Picker(hour, minute, second)

anishlg:
updated the old duration picker
https://github.com/enyojs/moonstone/pull/1553

review comments updated
1. Used single value property than using hour, minute and seconds values separately.
2. Min/max values are removed as it's always constant.
3. Combined DurationPickerBase and DurationPcker into a single file as there won't be any future sub kinds for DurationPcker 
4.Made the template property published.
5. Removed DurationPickerBase.openChanged method as it doesn't do anything
6. Updated the sample.


samples/DurationPickerSample.html
```diff
@@ -0,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+<head>
+	<meta charset="UTF-8">
+	<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+	<title>Duration Picker Sample</title>
+	<!-- -->
+	<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+	<!-- Less.js (uncomment for client-side rendering of less stylesheets; leave commented to use only CSS)
+	<script src="../../../enyo/tools/minifier/node_modules/less/dist/less-1.7.0.min.js"></script> -->
```
otisnate: Please update less related routine. You need to refer https://github.com/enyojs/moonstone/blob/master/samples/ButtonSample.html

samples/DurationPickerSample.html
```diff
@@ -0,0 +1,27 @@
...
+	<script src="../../../enyo/enyo.js" type="text/javascript"></script>
+	<script src="../../enyo-ilib/package.js" type="text/javascript"></script>
+	<script src="../../enyo-cordova/package.js" type="text/javascript"></script>
```
otisnate: Do you need cordova APIs? If not, you can remove this line

otisnate:
Please update manifest.json and manifest-light.json file. Then you need to verify if Sampler shows Duration Picker sample

samples/DurationPickerSample.js
```diff
@@ -0,0 +1,51 @@
...
+			{classes: 'moon-7h moon-vspacing-s', components: [
+				{kind: 'moon.DurationPicker', name: 'pickerDuration', noneText: 'Pick Duration', content: 'Duration', countdown: true, value: '', template: 'h:m:s', onDurationChange: 'durationChanged', onCountdownExpired: 'countdownExpired'},
+				{kind: 'moon.Button', name: 'buttonReset', content: 'Reset Duration', small: true, ontap: 'resetTapped'},
+				{kind: 'moon.Button', name: 'countdownToggle', content: 'Toggle Coutdown', small: true, ontap: 'countdownTapped'}
+			]}
+		]},		
+		{kind: 'moon.InputDecorator', components: [
+			{kind: 'moon.Input', placeholder: 'h:m:s', name: 'newTemplate'}			
+		]},
+		{kind: 'moon.Button', name: 'setTemplate', content: 'Set Template', ontap: 'setNewTemplate'},
```
Djspaceg: Lets set these buttons to `small: true` so their height matches the inputs.

source/DurationPicker.js
```diff
@@ -0,0 +1,565 @@
...
+	* `moon.DurationPickerBase` is a base kind implementing fuctionality
+	* not intended to be used directly by  {@link moon.DurationPicker}
+	*
+	* @class moon.DurationPickerBase
+	* @extends moon.ExpandableListItem
+	* @protected
+	*/
+
+	enyo.kind(
+		/** @lends moon.DurationPickerBase.prototype */ {
```
Djspaceg: Docs still mention DurationPickerBase.

anishlg:
Duration picker is updated
1. removed the duration picker base component.
2. updated with review comments. 
3. added 'disablepicker' option to duration picker.
4. updated the duration picker sample with full, disabled and non-editable duration pickers

source/DurationPicker.js
```diff
@@ -0,0 +1,516 @@
...
+				{name: 'client', kind: 'enyo.Control', classes: 'enyo-tool-decorator moon-date-picker-client', onSpotlightLeft: 'closePicker', onSpotlightSelect: 'closePicker'}
+			]}
+		],
+
+		/**
+		* @private
+		*/
+		bindings: [
+			{from: '.disabled', to: '.$.headerWrapper.disabled'},
+			{from: '.disabledPicker', to: '.$.hour.disabled'},
```
otisnate: Is there requirement for ```disabledPicker``` option?

source/DurationPicker.js
```diff
@@ -0,0 +1,516 @@
...
+			{from: '.disabledPicker', to: '.$.minute.disabled'},
+			{from: '.disabledPicker', to: '.$.second.disabled'}
+		],
+
+		/**
+		* @private
+		*/
+		constructor: function() {
+			this.inherited(arguments);
```
otisnate: Please use ```enyo.inherit``` instead of ```this.inherited``` for every functon

source/DurationPicker.js
```diff
@@ -0,0 +1,516 @@
...
+					{name: 'secondLabel', content: this.secondText, classes: 'moon-date-picker-label moon-divider-text'}
+				]}
+			);
+		},
+
+		/**
+		* @fires moon.DurationePicker#onChange
+		* @private
+		*/
+		handleChangeEvent: function (sender, ev) {
```
otisnate: Please change ```(inSender, inEvent)``` instead of ```(sender, ev)```

samples/DurationPickerSample.js
```diff
@@ -0,0 +1,73 @@
...
+		if (this.$.result && ev.value) {
+			this.$.result.set('content', ev.name + ' changed to ' + ev.value);
+		}
+	},
+	countdownExpired: function(sender, ev) {
+		if (this.$.result && ev.name) {
+			this.$.result.set('content', ev.name + ' expired');
+		}
+	},
+	resetTapped: function(sender, ev) {
```
otisnate: Please add ```this.$.pickerDuration.set("open", false);```

source/DurationPicker.js
Djspaceg: `(inSender, inEvent)` is the old naming convention we no longer use. `(sender, ev)` is the current, shorter names.

samples/DurationPickerSample.js
Djspaceg: Why? Shouldn't you be able to reset the picker with it still open? This is a sample.

source/DurationPicker.js
Djspaceg: Yes, most controls should have a disabled state.

source/DurationPicker.js
Djspaceg: While `enyo.inherit` is technically faster, `this.inherited` makes the code easier to read. We could, during minification for example, automatically upgrade each `this.inherited` to a faster `enyo.inherit` method, because it's so consistent and easily readable. There won't be hundreds of these stamped out at a time on a single view, so the speed benefit is minimal. Either of these options is technically fine though.

source/DurationPicker.js
aarontam: What @Djspaceg said is correct, and just to add on to that, as a general change we have been removing the `in` prefix from all of our parameter names in favor of shorter names.

samples/DurationPickerSample.js
aarontam: I *think* there is a (known?) issue where resetting a picker while it is still open results in the child picker values not being updated until the picker is closed and opened again. You can see this in the `DatePickerSample` as well, which has the same behavior and does not contain an explicit call to close the picker. It's possible that the explicit call to close the picker was an app workaround for this, but I'm unclear on how frequent the case is to reset the picker while it is open. In any case, this might be something to dig into to fix, as a separate issue, and we probably wouldn't want the explicit call to close the picker since it would just be a workaround and not ideal.

Djspaceg: This behavior doesn't seem to be the case with this picker in my local tests. Are you seeing otherwise?

aarontam: Sorry just saw this (didn't get a notification for some reason)! I hadn't had a chance to actually try out the DurationPicker at the time - it appears to be working correctly for the case of resetting, so it was a workaround that was specific to `moon.DateTimePickerBase`, which is probably context for @otisnate's request/comment. In any case, it doesn't look like we need to add the call here.

source/DurationPicker.js
otisnate: Yes, you are right. But ```disabledPicker``` is new flag and is not same with ```disabled```

source/DurationPicker.js
otisnate: When can we check current shorter name's rule?

samples/DurationPickerSample.js
otisnate: I just want to sync up with TimePickerSample - https://github.com/enyojs/moonstone/blob/master/samples/TimePickerSample.js#L77

anishlg:
The duration picker has updated
1. add start/cancel and pause/resume buttons.
2. hide/show button option for application developer.
3. the application developer can set any order for hour, minute and second
4. the picker values are removed 0 from single digits

source/DurationPicker.js
Djspaceg: This change of standards was made in July of 2014, during the docs migration and code cleanup.

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+					{kind: 'moon.Button', name: 'pauseButton', content: this.pauseConent, small: true, ontap: 'pauseTapped'}
+				]}
+			]}
+		],
+
+		/**
+		* @private
+		*/
+		bindings: [
+			{from: '.disabled', to: '.$.headerWrapper.disabled'},
```
aarontam: Hi @anishlg, per @Djspaceg's previous comment, we can and should remove the leading "." before each of these binding paths as they are no longer necessary, and the removal of the "." prefix will speed things up.

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+			this.valueArray['second'] = 0;
+			if (this.$.second) this.$.second.set('value',0);
+		},
+
+		/**
+		* @private
+		*/
+		closePicker: function (sender, ev) {
+			/**
+			* If select/enter is pressed on any date picker item or the left key is pressed on the
```
aarontam: Pressing 5-way left on the first item does not seem to close the drawer.

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+				}
+			} else {
+				this.set('active', true);
+			}
+		},
+
+		/**
+		* @private
+		*/
+		hourTextChanged: function (inOldvalue, inNewValue) {
```
aarontam: Can we update the parameter names so that they do not use the "in" prefix here? Similarly for the other two change handlers that follow.

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+		pickerExpired: function () {
+			this.cancelPicker();
+			this.doPickerExpired({name:this.name});
+		},
+
+		/**
+		* @private
+		*/
+		changeCountdown: function () {
+			if (this.countdown) {
```
aarontam: Can we simplify this entire method to:

```javascript
if (this.$.hour)
	this.$.hour.set('disabled', this.countdown);
if (this.$.minute)
	this.$.minute.set('disabled', this.countdown);
if (this.$.second)
	this.$.second.set('disabled' , this.countdown);
```

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+		templateArray: [],
+
+		/**
+		* Optional label for Start Button
+		*
+		* @type {String}
+		* @default moon.$L('Start')
+		* @private
+		*/
+		startConent: moon.$L('Start'),
```
aarontam: This should probably be `startContent`.

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+
+		/**
+		* Optional label for Pause Button
+		*
+		* @type {String}
+		* @default moon.$L('Pause')
+		* @private
+		*/
+		pauseConent: moon.$L('Pause'),
```
aarontam: This should probably be `pauseContent`.

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+		* @default true
+		* @private
+		*/
+		pauseDisabled: true,
+
+		/**
+		* @private
+		*/
+		observers: {
+			'changeCountdown': ['countdown']
```
aarontam: Do we need to explicitly specify this observer, or could we instead just implement a `countdownChanged` method?

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+		* @private
+		*/
+		timer: null,
+
+		/**
+		* @type {Object}
+		* @default {}
+		* @private
+		*/
+		valueArray: {},
```
aarontam: This is on the nitpicky side, but this is an object and not an array. :)

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+			{from: '.pauseDisabled', to: '.$.pauseButton.disabled'},
+			{from: '.enableTimer', to: '.$.buttonPlaceHolder.showing'}
+		],
+
+		/**
+		* @private
+		*/
+		constructor: function () {
+			this.inherited(arguments);
+			this.valueArray = {
```
aarontam: Is there a specific reason we need to initialize these here, as opposed to doing so in `create`?

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+					]}
+				);
+			}
+			this.render();
+		},
+
+		/**
+		* @private
+		*/
+		currentTemplate: function () {
```
aarontam: On the nitpicky side again, but this sounds more like a property name than a method - how about something like 'initTemplate` or `parseTemplate`?

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+			this.setupPickers();
+			this.setupArray();
+			this.setValue(this.formatValue());
+		},
+
+		/**
+		* @private
+		*/
+		setupArray: function () {
+			if (this.template.search('h') < 0)
```
aarontam: I think using `indexOf` would be more performant here: http://jsperf.com/exec-vs-match-vs-test-vs-search/11.

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+		* @private
+		*/
+		formatText: function () {
+			var item, idx, text = '';
+			if (this.valueArray) {
+				for (idx = 0; idx < this.templateArray.length; idx++) {
+					item = this.templateArray[idx];
+					switch (item) {
+						case 'h':
+							text = text + ' ' + this.valueArray['hour'] + ' ' + moon.$L('Hours');
```
aarontam: We need to take into account number agreement between the value and the units (i.e. 1 Hour).

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+				if (this.$.second)
+					this.$.second.set('disabled' , false);
+			}
+		},
+
+		/**
+		* @private
+		*/
+		startTapped: function () {
+			if (this.startConent == moon.$L('Start')) {
```
aarontam: It seems like a bad idea to have the string value for `startContent` effectively defined in two locations as we're using it in the conditional here. Can we instead check to see if we are currently counting down, to determine which path to take?

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+			} else {
+				this.set('value', '');
+			}
+		},
+
+		/**
+		* @private
+		*/
+		pauseTapped: function () {
+			if (this.pauseConent == moon.$L('Pause')) {
```
aarontam: Same as above - makes me uncomfortable to be comparing against an effectively static string as a way of knowing our current state.

source/DurationPicker.js
```diff
@@ -0,0 +1,733 @@
...
+			this.set('startConent', moon.$L('Cancel'));
+			this.set('pauseDisabled', false);
+			this.startTimer();
+		},
+
+		/**
+		* @private
+		*/
+		cancelPicker: function () {
+			this.set('startConent', moon.$L('Start'));
```
aarontam: In fact, we might want to consider storing these strings in a single location, possibly as private properties of this control, so we can update them all in one place. As it stands now, we would have to remember to update the text content for "Pause" in at least 4 locations, if we ever wanted to change it.

Djspaceg: If you add these as locally scoped private variables, on around line#2, we'll get all of the benefits of the strings stored in one place, but none of the memory or loading overhead of a published property. Ex:
```
var pauseStr = $L('Pause),
    startStr = $L('Start');
```

aarontam: :+1:

anishlg:
The duration picker updated with review comments.
1. removed the '.' prefix from binding
2. removed 'in' prefix from 'valueChanged()' functions
3. updated the 'countdownChanged()' fucntion
4. 'startContent' and 'pauseContent' text updated
5. removed the observer
6. replaced 'valueArray' to 'values'
7. replaced 'currentTemplate()' to 'initTemplate()'. 
8. moved the 'constructor()' code to 'initDefault()' 
9. replaced 'search()' array with 'indexOf()'
10.fixed the number agreement between the value and the units
11. created private variables for button text with ilib support and removed the direct string checking used in condition fucntions
12. resolved 1 second delay issue.

source/DurationPicker.js
```diff
@@ -0,0 +1,781 @@
...
+		* @private
+		*/
+		formatText: function () {
+			var item, idx, text = '';
+			if (this.values) {
+				for (idx = 0; idx < this.templateArray.length; idx++) {
+					item = this.templateArray[idx];
+					switch (item) {
+						case 'h':
+							text = text + ' ' + this.values['hour'] + ' ' + (this.values['hour'] == 1 ? moon.$L('Hour') : moon.$L('Hours'));
```
aarontam: Hi @anishlg, instead of using `moon.$L('Hour')` and `moon.$L('Hours')` here, can we instead utilize our `hourText` property? Similarly, we would want to do the same thing for minutes and seconds.

source/DurationPicker.js
```diff
@@ -0,0 +1,781 @@
...
+		templateArray: [],
+
+		/**
+		* Optional label for Start Button
+		*
+		* @type {String}
+		* @default moon.$L('Start')
+		* @private
+		*/
+		startText: moon.$L('Start'),			// i18n 'Start' label in moon.DurationPicker buttons
```
aarontam: Having these as properties on the kind definitely works - I was curious if you had tried @Djspaceg's approach as a more efficient alternative: https://github.com/enyojs/moonstone/pull/1978#discussion_r25743288.

anishlg:
updated the duration picker to reuse the text property

anishlg:
Duration picker updated
1. the optimization suggested is implemented. all the texts are made to local private variables.
2. implemented an interval feature than using window.setInterval() as this causes time variation in javascript.

anishlg:
the tickInterval() renamed as moon.tickInterval() to solve the travis error

anishlg:
updated the duration picker
1. used fixed LTR.
2. used ilib lib for capitalized and the non-capitalized strings
3. used the ilib.DurFmt to get the formatted string

aarontam:
@anishlg when you next make an update, can you also fix the merge conflict? Thanks!

anishlg:
created new PR request 
https://github.com/enyojs/moonstone/pull/2031

hence closing this request

PR #1983, opened 2015-02-11
ENYO-980: Lite: Fixed size was set in moon.ContextualPopup

yeram:
## Issue
Some private properties are set by fixed number

## Fix
Set private properties as resolution independent by using moon.riScale()


graynorton:
Looks like these values also need to be tweaked to make them multiples of three...

PR #1991, opened 2015-02-16
Enyo 951 yeramchoi

yeram:


source/PanelWithAnimator.js
```diff
@@ -0,0 +1,137 @@
...
+			return true;
+		}
+	},
+	drawPanelFrame: function(sender, ev) {
+		this.$.viewport.applyStyle('-webkit-transform', 'translate3d(0, -' + sender.value + '%, 0)');
+		this.$.contentWrapper.applyStyle('-webkit-transform', 'translate3d(0, ' + sender.value + '%, 0)');
+	},
+	drawBreadFrame: function(sender, ev) {
+		this.$.breadcrumbViewport.applyStyle('-webkit-transform', 'translate3d(0, ' + sender.value + '%, 0)');
+		this.$.breadcrumbBackground.applyStyle('-webkit-transform', 'translate3d(0, -' + sender.value + '%, 0)');
```
Djspaceg: We should convert these to use `enyo.dom.transform({translate3d: '0, -' + sender.value + '%, 0'})`
There's probably also a way to get the value without having to refer to `sender`. We've recently rarely been using sender because it's unreliable. Often `ev.originator` is what you're looking for and is more reliable.

PR #1999, opened 2015-02-24
ENYO-1013: Set background image when sizing is specified.

aarontam:
### Issue
This seems to have been broken for a long time, but setting the `sizing` property of `moon.Image` does not properly set the background-image as there is an ordering issue where the setting of the background-image is gated by the value of `sizing`, but in the current ordering of the bindings, the value of `sizing` would be set after the value of `src` at create time.

### Fix
We modify the order of the bindings such that the binding for `sizing` occurs before the binding for `src`. Additionally, we update the sample to demonstrate the use of the `sizing` property.

### DIFF
--- a/source/Image.js
+++ b/source/Image.js
@@ -113,9 +113,9 @@
 		* @private
 		*/
 		bindings: [
+			{from: '.sizing', to: '.$.image.sizing'},
 			{from: '.src', to: '.$.image.src'},
 			{from: '.alt', to: '.$.image.alt'},
-			{from: '.sizing', to: '.$.image.sizing'},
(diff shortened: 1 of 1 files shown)


ryanjduffy:
Having timing-dependent bindings seems like a bad place to be and prone to mysterious errors. Would moving `srcChanged` out of the `this.generated` check in `sizingChanged` also solve the problem. Doesn't appear to be a reason to guard `srcChanged` here.

aarontam:
@ryanjduffy yep that's a great call - will close this in favor of making a change to `sizingChanged`: https://github.com/enyojs/enyo/pull/1051.

PR #2000, opened 2015-02-25
ENYO-964 Prevent to back previous page in browser

DavidUM:
Issue
------
Pressing back key make it back to previous page in browser

Cause
-------
history.back() call without pushed history state will back to previous page. Natural behavior of history object.

Fix
----
Add a guidance state to prevent to back to previous page


source/History.js
```diff
@@ -353,6 +366,9 @@
 
 			switch (state) {
 			case 'empty':
+				if (history.state.currentObjId == 'historyMaster') {
```
aarontam: Hi @DavidUM, it seems that `history.state` can be `null` here, if we do not have any previous history i.e. this is the first page we load in the browser. I think we should add logic to handle this case properly, which may involve adding guard code. Thanks!

DavidUM: Hi @aarontam , thank you for your point. 
For this reason I added an {{_initHistoryState}} method.
But I observed that there are some cases which above method was not called before I pressed back key.
For safety, I add one more condition. Thank you.

aarontam:
Hi @DavidUM, I think there are a number of extra (and duplicate) commits you didn't intend to include in this branch/PR - maybe it might be best to open a new PR and apply your work there? Thanks!

DavidUM:
I see. I rebased this branch but it show above. 
Let me open another PR. Thanks

PR #2002, opened 2015-02-25
ENYO-965 Restore spotlight focus to contextual popup

DavidUM:
Issue
-------
After pressing back key in ContexualPopup, spotlight focus does not return to origin.

Cause
--------
When popup is opened, contents of popup had spotlight focus. 
But back key hide popup without spotlight focusing, so spotlight focus lost its way and back to container.

Fix
----
When popup is closed with back key, trigger onRequestSpot event to focus contextualPopup button.

### DIFF
--- a/source/ContextualPopupDecorator.js
+++ b/source/ContextualPopupDecorator.js
@@ -136,6 +136,7 @@
 		*/
 		requestSpot: function (inSender, inEvent) {
 			enyo.Spotlight.spot(this);
+			return true;
 		}
 	});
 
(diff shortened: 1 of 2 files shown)


source/ContextualPopupDecorator.js
```diff
@@ -126,6 +127,15 @@
...
 			this.waterfallDown('onRequestHidePopup');
+		},
+
+		/**
+		* When back key is pressed, return spotlight focus to popup button
+		*
+		* @private
+		*/
+		requestSpot: function (inSender, inEvent) {
+			enyo.Spotlight.spot(this);
```
clinuz: Shouldn't this `return true;`?

DavidUM: Yes, you are right. I missed it!
I updated it and testified its functionality on PC.
Thanks @clinuz

PR #2003, opened 2015-02-25
ENYO-1091: Marquee (when set marqueeOnHover as true) is not stop when fo...

KunmyonChoi:
...cus moves to other control by 5way

### Issue:
Header title marquee is not stop after focus move to 'switch mode' in 5way.

### Fix:
Stops marquee if other control is focused only when a marquee control which is having marqueeOnHover option as true is flow.


lishammel:
Closing this PR out as the jira was fixed another way

PR #2009, opened 2015-03-03
ENYO-513: Overlay play icon for video/audio files

anishramesan:
### Approach:
Added mixin for GridListImageItem to expose properties like;
type - to identify the type of poster(i.e. video/audio, image).
position - top/bottom alignment.
icon - to customize video play icon.

Enyo-DCO-1.1-Signed-off-by: Anish Ramesan <anish.ramesan@lge.com>


css/BadgeOverlaySupport.less
```diff
@@ -0,0 +1,8 @@
+.moon-badge-overlay-support-scrim {
+	display: block;
+	z-index: 1000;
+	width: 100%;
+	height: 55px;
```
Djspaceg: We made a decision recently to make all measurements a multiple of 3, so they can be converted to rem units "smoothly". The easiest way is to +/- 1 to any number. Although it works out even better if it's a multiple of 6, 12 or even 24.

source/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,144 @@
...
+					this.alignChanged();
+				}
+			};
+		}),
+
+		/**
+		* @private
+		*/
+		bindings: [
+			{from: '.badgeScrimIcon', to: '.$.badgeScrimIcon.src'}
```
Djspaceg: We no longer need to prefix these strings with a ".", and in fact, the code runs faster if the "." is removed.
```
{from: 'badgeScrimIcon', to: '$.badgeScrimIcon.src'}
```

source/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,144 @@
...
+		_badgeScrim: [
+			{name:'badgeScrim', classes: 'enyo-fit moon-badge-overlay-support-scrim', components: [
+				{name:'badgeScrimIcon', kind: 'moon.Icon', small: false, icon: 'play', spotlight: false}
+			]}
+		],
+
+		alignChanged: function() {
+			if(this.$.badgeScrim) {
+				// TODO: identify the proper value based on height of image.
+				this.$.badgeScrim.applyStyle("top", this.align=='top'?0:'125px');
```
Djspaceg: We should avoid using hard-coded values in the JavaScript. This same effect could very easily be achieved by adding a class with the static rules written into the CSS files. It would also be preferable to have this not be a static value, something like `top: 50%` would work out better in the case of supporting multiple different sizes.

source/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,144 @@
...
+		* @private
+		*/
+		create: enyo.inherit(function (sup) {
+			return function () {
+				sup.apply(this, arguments);
+				if(this.type && this.type != "image") {
+					this.createChrome(this._badgeScrim);
+					// Allow the icon to be modified by user
+					if (this.badgeScrimIcon) {
+						this.$.badgeScrimIcon.set('icon','');
```
Djspaceg: moon.Icons support both "icon" and "src" property. The below we should also have an icon binding to support the font-icon cases. All of the moonstone icons are currently and many new additional icons are icon-font and no longer rasterized images.

source/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,144 @@
...
+		* @method
+		* @private
+		*/
+		create: enyo.inherit(function (sup) {
+			return function () {
+				sup.apply(this, arguments);
+				if(this.type && this.type != "image") {
+					this.createChrome(this._badgeScrim);
+					// Allow the icon to be modified by user
+					if (this.badgeScrimIcon) {
```
otisnate: Should we still check ```if (this.badgeScrimIcon)``` after running ```this.createChrome(this._badgeScrim)``` ?

samples/DataGridListSample.js
```diff
@@ -87,8 +87,8 @@ enyo.kind({
 enyo.kind({
 	name: "moon.sample.GridSampleItem",
 	kind: "moon.GridListImageItem",
-	mixins: ["moon.SelectionOverlaySupport"],
-	selectionOverlayVerticalOffset: 35,
```
otisnate: I think we still need sample code for ```moon.SelectionOverlaySupport```
How about creating ```moon.sample.GridSampleVideoItem``` inherited from ```moon.BadgeOverlaySupport```
And you can add ToggleButton in DataGridList sample to change between ```moon.SelectionOverlaySupport``` and ```moon.BadgeOverlaySupport```

source/BadgeOverlaySupport.js
otisnate: Please refer to http://nebula.lgsvl.com/dev-guide/release-notes/migration-2.3-to-2.5.html

css/BadgeOverlaySupport.less
```diff
@@ -0,0 +1,8 @@
...
+	color: white;
+	background-color: rgba(0, 0, 0, 0.5);
+}
```
otisnate: Please extract each value and add those values to ```moonstone-variables.less``` (or ```moonstone-variables-dark.less``` and ```moonstone-variables-light.less```)
You can refer to SelectionOverlaySupport.less file

source/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,144 @@
...
+		* Identifies the MIME type of badge. 
+		* Possible values are "image", "audio" or "video".
+		* Defaulted as 'image'.
+		*
+		* @name moon.BadgeOverlaySupport#type
+		* @type {String}
+		* @default undefined
+		* @public
+		*/
+
```
Djspaceg: Could we add the property names under these comments, for consistency and readability. They can be assigned to `null`, but they should be present.

Same for `badgeScrimIcon` below.

source/BadgeOverlaySupport.js
Djspaceg: Yes. The `badgeScrimIcon` may still be undefined/null. If it is, we need to blank-out the icon value.

samples/DataGridListSample.js
Djspaceg: I agree. It would be nice to see the examples two approaches, instead of just one.

css/BadgeOverlaySupport.less
```diff
@@ -0,0 +1,17 @@
+.moon-badge-overlay-support .moon-image-client {
+	color: white;
+	padding: 0;
+	background-color: rgba(0, 0, 0, 0.5);
+	.moon-icon {
+		width: 36px;
+		height: 36px;
```
Djspaceg: If you're going to change the size of an icon, you'll also need to change the size of the line-height, so the font-icons stay vertically centered in their frame:
```
line-height: 36px;
```

css/BadgeOverlaySupport.less
```diff
@@ -0,0 +1,17 @@
...
+	}
+}
+.moon-badge-overlay-support .moon-image {
+	display: block;
+	margin: 0;
+}
+.moon-badge-overlay-support.top .moon-image-client {
+	bottom: initial;
+	top: 0;
+}
```
Djspaceg: Lets rearrange this file like this:
```
.moon-badge-overlay-support {
	.moon-image-client {
		color: #fff;
		padding: 0;
		background-color: rgba(0, 0, 0, 0.5);

		.moon-icon {
			width: 36px;
			height: 36px;
			line-height: 36px;
		}
	}
	&.top .moon-image-client {
		bottom: initial;
		top: 0;
	}
	.moon-image {
		display: block;
		margin: 0;
	}
}
```
It will visually group the rules and the CSS will be just the same. (changed the color from "white" to "#fff" for consistent color naming across the framework, added `line-height`, and swapped the order of the `.top` class.

anishramesan:
@Djspaceg Sorry if I misunderstand you. If we consider small:true(default case), invisible tap area is added around moon.icon which makes it smaller than 36px.

Djspaceg:
@anishramesan I just tried removing the `small:false` property in BadgeOverlaySupport, and removed the custom sizing, then added the same font-sizing rule, it looks like the right size and position, with less overrides. Here's the one override I changed:
```
.moon-icon {
    width: @moon-icon-small-size;
    height: @moon-icon-small-size;
    line-height: @moon-icon-small-size;
}
```
to
```
.moon-icon.small {
    font-size: (@moon-icon-size * 2);
}
```

Also, please convert any/all files to use tab indenting (instead of spaces), and all files from double-quotes `"` to single quotes `'`.

css/lessc.sh
```diff
@@ -1 +1 @@
-../../enyo/tools/lessc.sh ./package.js
+../../enyo/tools/lessc.sh ./package.js --ri
```
ryanjduffy: This is somewhat minor but I think we should pull this out of this PR. If we want to make this kind of global change, it should be in its own JIRA & PR with appropriate justification.

samples/DataGridListSample.js
ryanjduffy: Since another sample was created for BadgeOverlaySupport, this change should be reverted. As it stands now, this sample is broken because it has buttons for adjusting the selection options but no way to select an item.

source/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,220 @@
...
+			};
+		}),
+
+		/**
+		* @method
+		* @private
+		*/
+		badgeTap: function (inSender, inEvent) {
+			if (inEvent && inEvent.originator) {
+				switch(inEvent.originator.name) {
```
ryanjduffy: I think we should revisit the event handling as this seems like a brittle approach. We're assuming a particular DOM structure that the mixin doesn't control. I couldn't get the onBadgeIconTap to fire at all because the tap area is the originator, not the icon. Further, we're breaking encapsulation by assuming the internal structure of moon.Image in referencing the `client` control.

source/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,220 @@
...
+	*		kind: 'moon.ImageItem', bindings: [
+	*			{from: '.model.title', to: '.label'},
+	*			{from: '.model.description', to: '.text'},
+	*			{from: '.model.coverSource', to: '.source'}
+	*		]
+	*	}
+	* ]}
+	*
+	* @mixin moon.BadgeOverlaySupport
+	* @protected
```
ryanjduffy: If there are public properties then the mixin has to be `@public` too so devs know that the properties exist. Furthermore, this should probably be public anyway because it isn't a framework-internal feature.

source/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,220 @@
...
+	*			{from: '.model.coverSource', to: '.source'}
+	*		]
+	*	}
+	* ]}
+	* ```
+	*
+	* The item may define a
+	* [position]{@link moon.BadgeOverlaySupport#position} to
+	* specify position of badge. Possible values are 'top' or 'bottom'. Defaulted as 'bottom'.
+	* ```javascript
```
ryanjduffy: Need an extra line before this for the code block to be picked up by the doc parser

source/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,220 @@
...
+		* @default undefined
+		* @public
+		*/
+		badgeType: null,
+
+		/**
+		* Font icon to be used in place of default icon.
+		*
+		* @name moon.BadgeOverlaySupport#badgeIcon
+		* @type {String}
```
ryanjduffy: For all three of these public properties, the documented default value (`undefined`) doesn't match the actual default value (`null`).

source/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,220 @@
...
+		*/
+		position: null,
+
+		/**
+		* @private
+		*/
+		events: {
+
+			/**
+			* {@link moon.BadgeOverlaySupport#onBadgeTap}
```
ryanjduffy: These don't link anywhere because the events haven't been documented correctly. See https://github.com/enyojs/moonstone/blob/master/source/Calendar.js#L2 for an example of documenting an event.

source/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,220 @@
...
+		bindings: [
+			{from: 'badgeSrc', to: '$.badgeScrimIcon.src'},
+			{from: 'badgeIcon', to: '$.badgeScrimIcon.icon', transform: function(val){ return val ? val : 'play';}}
+		],
+
+		/**
+		* @private
+		*/
+		_badgeOverride: {
+			image: { kind:"moon.Image", components:[{
```
ryanjduffy: Please take a quick pass to convert double quotes to single and add spaces after `function`, `if`, and `switch` statements like so:

`method: function () {`, `if (something) {` and `switch (condition) {`

PR #2010, opened 2015-03-04
ENYO-510 - Modified DataGridList Sample so to pass custom placeholder to GridListImageItem kind

rakesh25:
Issue:

DataGridListSample needs to pass custom placeholder(enyo.Image.placeholder) to GridListImage kind

Fix:

Though binding, placeholder value being passed to kind.

Enyo-DCO-1.1-Signed-off-by: Rakesh kumar rakesh25.kumar@lge.com

### DIFF
--- a/samples/DataGridListSample.js
+++ b/samples/DataGridListSample.js
@@ -43,9 +43,9 @@ enyo.kind({
 		]}
 	],
 	bindings: [
-		{from: ".collection", to: ".$.dataList.collection"},
-		{from: ".collection", to: ".$.gridList.collection"},
-		{from: ".$.selectionTypeGroup.active", to: ".$.gridList.selectionType",
+		{from: "collection", to: "$.dataList.collection"},
+		{from: "collection", to: "$.gridList.collection"},
+		{from: "$.selectionTypeGroup.active", to: "$.gridList.selectionType",
 			transform: function (selected) {
 				return selected && selected.value;
 			}
@@ -67,8 +67,7 @@ enyo.kind({
 				selected: false,
 				text: "Item " + idx + title,
 				subText: subTitle,
-				url: "http://placehold.it/300x300/" + Math.floor(Math.random()*0x1000000).toString(16) + "/ffffff&text=Image " + idx,
-				placeholder: enyo.Image.placeholder
+				url: "http://placehold.it/300x300/" + Math.floor(Math.random()*0x1000000).toString(16) + "/ffffff&text=Image " + idx
 			});
 		}
 		// update our internal index so it will always generate unique values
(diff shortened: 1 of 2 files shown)


samples/DataGridListSample.js
```diff
@@ -67,7 +67,8 @@ enyo.kind({
 				selected: false,
 				text: "Item " + idx + title,
 				subText: subTitle,
-				url: "http://placehold.it/300x300/" + Math.floor(Math.random()*0x1000000).toString(16) + "/ffffff&text=Image " + idx
+				url: "http://placehold.it/300x300/" + Math.floor(Math.random()*0x1000000).toString(16) + "/ffffff&text=Image " + idx,
+				placeholder: enyo.Image.placeholder
```
aarontam: I think we can probably remove this as the default value of `placeholder` is `enyo.Image.placeholder`

source/GridListImageItem.js
```diff
@@ -33,6 +33,20 @@
 
 		/**
 		* @private
+		* @lends moon.Button.prototype
+		*/
+		published: {
+			/**
+			* @type {String}
+			* @default ''
```
aarontam: The default value in the documentation should be updated to reflect the current value.

source/GridListImageItem.js
```diff
@@ -33,6 +33,20 @@
 
 		/**
 		* @private
+		* @lends moon.Button.prototype
```
aarontam: I'm guessing this is a copy/paste remnant? :)

samples/DataGridListSample.js
```diff
@@ -90,9 +91,11 @@ enyo.kind({
 	mixins: ["moon.SelectionOverlaySupport"],
 	selectionOverlayVerticalOffset: 35,
 	subCaption: "Sub Caption",
+	imageSizing: "cover",
 	bindings: [
 		{from: ".model.text", to: ".caption"},
 		{from: ".model.subText", to: ".subCaption"},
-		{from: ".model.url", to: ".source"}
+		{from: ".model.url", to: ".source"},
+		{from: ".model.placeholder", to: ".placeholder"}
```
Djspaceg: Please remove the prefixed `.` in front of each of these bindings to/from strings. It is no longer needed.

samples/DataGridListSample.js
```diff
@@ -90,9 +91,11 @@ enyo.kind({
 	mixins: ["moon.SelectionOverlaySupport"],
 	selectionOverlayVerticalOffset: 35,
 	subCaption: "Sub Caption",
+	imageSizing: "cover",
```
Djspaceg: This new default is unnecessary. It changes the way that the sample is rendered, switching from `img` tag to `div` tag. It can stay only if we have a reason for the change.

samples/DataGridListSample.js
rakesh25: Got it!. Removing it.

samples/DataGridListSample.js
rakesh25: You are right. Removing it making img as default case.

ryanjduffy:
Closing master PRs

PR #2022, opened 2015-03-07
ENYO-23301: Start preview if necessary, when moving pointer in the slider.

aarontam:


ryanjduffy:
Now today, I can't repro the problem on my branch. Can you check it again too and see if I'm missing it? Thanks!

PR #2023, opened 2015-03-09
ENYO-1159: Clean-up cruft.

aarontam:
### Issue
Ran into this while merging another PR and wanted to clean it up while it was still fresh. Due to some refactoring and clean-up in https://github.com/enyojs/moonstone/pull/1693, there were a few additional properties, methods, and calls, that were orphaned.

### Fix
The unnecessary properties, methods, and calls have now been removed as necessary.

### DIFF
--- a/source/Popup.js
+++ b/source/Popup.js
@@ -167,16 +167,6 @@
 		activator: null,
 
 		/**
-		* @private
-		*/
-		directShowHide: false,
-
-		/**
-		* @private
-		*/
-		initialDuration: null,
-
-		/**
 		* Creates chrome components.
 		*
 		* @private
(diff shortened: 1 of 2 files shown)


source/Popup.js
```diff
@@ -192,7 +182,6 @@
 		create: function () {
 			this.inherited(arguments);
 			this.animateChanged();
-			this.initialDuration = this.getComputedStyleValue('-webkit-transition-duration', '0.4s');
```
ryanjduffy: Not that it changes the PR at all but I don't think this could have ever returned anything other the `0.4s` because `getComputedStyleValue` is guarded by `hasNode()` which would always (I think) be falsey at this point in the lifecycle.

/shrug

PR #2024, opened 2015-03-09
ENYO-1138: Ensuring requestScrollIntoView's originator is showing before...

germboy:
... triggering the scroll.

### Issue
Developers could have a need to arbitrarily hide controls at any time - including after issuing a "requestScrollIntoView" event, causing an unexpected scroll to a hidden control (or bounce-effect).

### Fix
Ensuring controls are showing before performing the scroll should eliminate these edge-cases.

Enyo-DCO-1.1-Signed-off-by: Jeremy Thomas <jeremy.thomas@lge.com>

### DIFF
--- a/source/MoonScrollStrategy.js
+++ b/source/MoonScrollStrategy.js
@@ -715,16 +715,19 @@
 		* @private
 		*/
 		requestScrollIntoView: function(sender, event) {
-			var showVertical, showHorizontal,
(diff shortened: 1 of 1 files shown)


germboy:
It seems to me that developers could have a need to arbitrarily hide controls at any time - including after issuing a "requestScrollIntoView" event. Ensuring controls are showing before performing the scroll should eliminate these edge-cases.

PR #2031, opened 2015-03-18
ENYO-493: Duration Picker (hour, minute, second)

anishlg:
updated the old duration picker
https://github.com/enyojs/moonstone/pull/1978/

### DIFF
--- a/source/DurationPicker.js
+++ b/source/DurationPicker.js
@@ -1,0 +1,690 @@
+(function (enyo, scope) {
+	/**
+	* Fires when the value changes.
+	*
+	* @event moon.DurationPicker#onChange
+	* @type {Object}
+	* @property {String} name - contains the name of this control.
+	* @public
+	*/
+
+	/**
+	* `moon.DurationPicker` is a [control]{@link enyo.Control} that can display -- or allow the
+	* selection of -- a duration expressed in hours, minutes and seconds, the selection
+	* display is configurable using template.
+	*/
+
+	var 
+		/**
+		* Optional label for hour
+		*
+		* @type {String}
+		* @default moon.$L("1#hour|#hours")
+		* @private
+		*/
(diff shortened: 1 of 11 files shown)


source/DurationPicker.js
```diff
@@ -0,0 +1,690 @@
...
+	});
+
+	(function () {
+		/* A replacement for window.setInterval() function, It use window.setTimeout() to perform the interval functionality
+		* It is based on enyo.perfNow() and use setInterval() and clearInterval() function as like with window
+		*
+		*/
+
+		/* Interval Object */
+		var Interval = function (func, delay) {
```
aarontam: Hi @anishlg, do you know why it was necessary to create all of this custom functionality? It appears to work fine, but I seem to recall we were originally utilizing `setInterval`.

anishlg: Hi @aarontam, as i mentioned earlier the window.setInterval() function always had +/- 1 ms difference in each interval (1 s). By getting added this in every interval, this will considerably get wrong duration on long run. In current the moon.interval keep the difference consistent from the starting time by using {{enyo.perfNow}}.

aarontam: Hi @anishlg, thanks for the clarification. I'm wondering if this shouldn't go into `enyo` core, maybe as part of `enyo.job` i.e. `enyo.repeatingJob`. It seems to be something that would be useful at a more general level, and could benefit from utilizing the job paradigm where a job can be named and a delay can be specified.

cc @Djspaceg @clinuz

anishlg:
Updated the Duration Picker
1. removed the setTimeout() function dependency and used the enyo core function enyo.startJob() to create the timer
2. removed the template property and used hour/minute/second hide/show properties

KunmyonChoi:
We closed the Jira because no app is planning to use this new control. Let's close the PR here.

PR #2045, opened 2015-03-27
ENYO-1251: Prevent unnecessary collapsing of input.

aarontam:
### Issue
When initiating a drag on the `moon.InputDecorator` control, with the VKB visible, the drag ends due to what seems to be a `mouseout` event being fired. This causes the `moon.ExpandableInput` to collapse in response to an `onSpotlightBlur` event being triggered by the `mouseout`, in addition to the `moon.InputDecorator` being focused in response to an `onSpotlightSelect`, also due to the `mouseout`.

### Fix
This fix attempts to address this behavior by preventing the collapse of the `moon.ExpandableInput` in this situation by guarding against collapsing when the previous event was `dragstart` or `drag`. Also took the liberty of converting parameter names to remove the `in` prefix.

### DIFF
--- a/source/ExpandableInput.js
+++ b/source/ExpandableInput.js
@@ -189,8 +189,10 @@
 		*/
 		inputBlur: function (sender, ev) {
 			var eventType = enyo.Spotlight.getLastEvent().type;
+			// TODO: adding a hacky workaround to prevent undesired collapse of the control when
+			// initiating a drag on the InputDecorator control on the TV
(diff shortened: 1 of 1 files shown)


DavidUM:
If we point out of input filed, should drawer close?
Even though I click another control, drawer is still open.

aarontam:
Hi @DavidUM yep that's a good catch - let's go the pointer-events route instead, thanks!

aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2046.

PR #2059, opened 2015-04-04
ENYO-1335: Date Picker : Passing undefined params stops application

kyo504:
### Issue ###

At creation time of Date Picker, *value* is set to null so iLib date object cannot be instantiated from it. However picker calls iLib function to calculate maximum days by passing undefined values. Inside iLib API, there exist a code trying to manipulate undefined params and it causes undefined error.

### Fix ###

Before calling iLib API to get **monthLength**, we'll check if parameters to be passed like **inYear** and **inMonth** have proper values.

Enyo-DCO-1.1-Signed-off-by: Hunkyo Jung <hunkyo.jung@lge.com>

### DIFF
--- a/source/DatePicker.js
+++ b/source/DatePicker.js
@@ -194,7 +194,7 @@
 					this.createComponent(
 						{classes: 'moon-date-picker-wrap', components:[
 							{kind:'moon.IntegerPicker', name:'day', classes:'moon-date-picker-field', wrap:true, digits:digits, min:1,
-							max:this.monthLength(values.fullYear, values.month), value: values.date, onChange: 'pickerChanged'},
+							max:values.maxDays, value: values.date, onChange: 'pickerChanged'},
 							{name: 'dayLabel', content: this.dayText, classes: 'moon-date-picker-label moon-divider-text'}
 						]});
 					break;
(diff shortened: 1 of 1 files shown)


DavidUM:
This issue is only happened in Ethiopia calendar. 
Every calendar has own guard code in i18n except Ethiopia.
So, i18n would better to add guard code for Ethiopia like others.

Anyway this guard code looks reasonable.

ryanjduffy:
I think we'd be better off skipping the call into ilib altogether by checking for invalid values in `DatePicker.monthLength()`. As it is now, that error is happening 2 functions deep in ilib and there's no value in those method calls. Even if they add a check, there's no reason to wait to hit it when we know we can't determine the month length when month and year are invalid.

kyo504:
@ryanjduffy Yes, you're right. However when I asked i18n team to add a check in ilib code, they said that the problem is to send undefined values to ilib and it should be avoided first. That's why I added guard code. (I heard i18n team also would add a check but I don't know when they are going to work on it.)

kyo504:
Previous condition to assign default values is removed. Instead monthLength function in date picker would check if inYear and inMonth have actual values before calling iLib API.

PR #2069, opened 2015-04-15
ENYO-1438: make css of videoPlayer's iconButton like REW/FF be effective only when icon is enabled.

suhlee:
## Issue
There was a Q issue about disabled imageIcon like rewind or fastFoward in videoPlayer.
If disabled rewind or fastFoward image Icon in videoPlayer is clicked, then it is affected by css and changes its color to red.

## FIx
It is  needed to make css effective only when icon is not disabled.
To do this, I added ":not(.disabled)" in VideoControlFullscreen.less.
So, when an icon is disabled, the css does not affect disabled icon.


aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2077.

PR #2074, opened 2015-04-15
ENYO-1441 : modify resolution update routine for removing old resolution classes

sbcho:
## issue
moon.ri.init() function needs a line for removing ScreenTypeOnBody
moon.ri.getScreenTypeObject always return old screenTypeObject if there is no parameter

## fix
added oldScreenType variable for for checking and removing old resolution classes.
and removed caching routine in getScreenTypeObject function.

Enyo-DCO-1.1-Signed-off-by: Sungbae Cho sb.cho@lge.com

### DIFF
--- a/source/moon-resolution.js
+++ b/source/moon-resolution.js
@@ -6,8 +6,11 @@
 		screenTypeObject,
 		oldScreenType;
 
-	var getScreenTypeObject = function (type) {
+	var getScreenTypeObject = function (type, ignoreCache) {
 		type = type || screenType;
+		if (type == screenType && !ignoreCache && screenTypeObject) {
+			return screenTypeObject;
+		}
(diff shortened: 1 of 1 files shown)


source/moon-resolution.js
```diff
@@ -3,11 +3,12 @@
 	var baseScreenType = 'fhd',
 		riRatio,
 		screenType,
-		screenTypeObject;
+		screenTypeObject,
+		oldScreenType;
 
-	var getScreenTypeObject = function (type) {
+	var getScreenTypeObject = function (type, ignoreCache) {
```
aarontam: @sbcho we can get rid of the `ignoreCache` parameter right?

sbcho: sure! I'll remove it :)

PR #2079, opened 2015-04-16
ENYO-1332: adjusting the content inside button according to the specs

rajyavardhanp:
The content inside the button extends beyond the curvature of the
button, which is a deviation from specs.
Padding of the button is increased to make content align with in the
curvature.
Enyo-DCO-1.1-Signed-off-by: Rajyavardhan P  <rajyavardhan.p@lge.com>

### DIFF
--- a/css/moonstone-variables.less
+++ b/css/moonstone-variables.less
@@ -194,7 +194,7 @@
 @moon-button-small-text-size: 24px;
 @moon-button-large-min-width: 180px;
 @moon-button-small-min-width: 132px;
-@moon-button-h-padding: 30px;
+@moon-button-h-padding: 27px;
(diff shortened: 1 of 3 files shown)


css/moonstone-variables.less
```diff
@@ -194,7 +194,7 @@
 @moon-button-small-text-size: 24px;
 @moon-button-large-min-width: 180px;
 @moon-button-small-min-width: 132px;
-@moon-button-h-padding: 18px;
+@moon-button-h-padding: 30px;
```
Djspaceg: Lets change this to {{27px}} so the small round button (one character) stays round.

lishammel:
Closing PR per Rajyavardhan; no longer needed.

PR #2080, opened 2015-04-17
ENYO-1292: ExpandableListItem: Lack of Padding displays in RTL

anishlg:
### Issue:
Lack of padding displays to the right edge of the expandable list item header when in RTL mode

### Fix:
The right padding for expandable picker header was changed from 0px to @moon-spotlight-outset 

### DIFF
--- a/css/ExpandableListItem.less
+++ b/css/ExpandableListItem.less
@@ -27,6 +27,9 @@
 .moon-expandable-list-item-client.indented {
 	padding-left: @moon-item-indent + @moon-spotlight-outset;
 }
+.enyo-locale-right-to-left .moon-expandable-list-item-header {
+	padding-right: @moon-spotlight-outset !important;
+}
 .enyo-locale-right-to-left .moon-expandable-list-item-client.indented {
 	padding-left: 0;
 	padding-right: @moon-item-indent + @moon-spotlight-outset;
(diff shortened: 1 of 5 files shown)


KunmyonChoi:
@anishlg , ExpandableListItem should not use picker CSS. I think we should move the padding to moon-expandable-list-item-header or moon-expandable-list-header.

KunmyonChoi:
It is strange because expandable list item is using picker CSS. Which is reverse direction access in component hierarchy. Please find out how we can remove these reverse reference.

anishlg:
The padding css is removed from ExpandablePicker and used ExpandableListItem for the same. This css property only required in RTL mode so the css class included the RTL in it

css/ExpandableListItem.less
```diff
@@ -27,6 +27,9 @@
 .moon-expandable-list-item-client.indented {
 	padding-left: @moon-item-indent + @moon-spotlight-outset;
 }
+.enyo-locale-right-to-left .moon-expandable-list-item-header {
```
KunmyonChoi: Hi @anishlg, 
We usually doesn't encourage to use '!important' in general. Instead, I think we can use a more specific css rule like '.enyo-locale-right-to-left .moon-expandable-list-header.moon-expandable-list-item-header'.
How do you think?

anishlg:
Updated with review comments. Please review again

anishlg:
updated the review comments
1. restored the item kind
2. rearranged the less file

webOS101:
PR against master not needed

PR #2081, opened 2015-04-17
ENYO-1285: Marquee Text is ellipsis but does not roll when content 

anishlg:
Issue:
The node.scrollWidth will return the correct value when the text inside the div is more wider than the div visible area. Else the node.scrollWidth will return the div size itself. In this case both the node.scrollWidth and node.getBoundingClientRect().width will be the same and the resultant marquee distance will be 0 so the marquee won't work. 

But the text-overflow: ellipsis is applied to the node starting from the div width and inside text content width is same. Which causes the issue.

Fix:
Remove the text-overflow : ellipsis from the node if the node.scrollWidth and node.getBoundingClientRect().width are same.

### DIFF
--- a/source/Marquee.js
+++ b/source/Marquee.js
@@ -692,6 +692,7 @@
 				// can occur with a moon.Header that is located inside a moon.Scroller which has
 				// vertical scrollbars visible.
 				this._marquee_detectAlignment();
+				this._marquee_setTextOverflow();
 			};
 		}),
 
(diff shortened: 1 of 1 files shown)


anishlg:
1.Remove the text-overflow : ellipsis from the node if the node.scrollWidth and node.getBoundingClientRect().width are same

2.  Changed the math.floor into math.ceil to upper round marquee distance value

KunmyonChoi:
@anishlg , it looks like high code operation for render and contentChanged. But, I don't have other solution in mind. Please look at little bit more on other solution.

source/Marquee.js
```diff
@@ -780,6 +794,7 @@
 		* @private
 		*/
 		_marquee_contentChanged: function () {
+			this._marquee_setTextOverflow();
```
KunmyonChoi: Hi Anish,
I think we need to test length after setting content on marquee text. How do you think?

source/Marquee.js
```diff
@@ -717,6 +718,19 @@
 			};
 		}),
 
+		_marquee_setTextOverflow: function () {
+			var node, rect;
+			node = this.$.marqueeText ? this.$.marqueeText.hasNode() : this.hasNode();
```
KunmyonChoi: I think we can use _marquee_calcDistance() instead here.

anishlg: It is hard to use the _marquee_calcDistance() function here, as the only similarity is in getting the node and rect values but the logic is different and the _marquee_calcDistance() code will unnecessarily create this._marquee_distance which will cause other issues

lishammel:
closing out per Anish - jira was solved with a different pr.

PR #2085, opened 2015-04-18
ENYO-898: Re-implement moon.Drawer using CSS transform

yeram:
To share concept of implementation.

### DIFF
--- a/css/Drawer.less
+++ b/css/Drawer.less
@@ -4,3 +4,8 @@
 .moon-drawer-client {
 	padding: 18px;
 }
+.moon-drawer {
+  position: absolute;
+  bottom: 0;
+  width: 100%;
+}
--- a/css/Drawer.less
+++ b/css/Drawer.less
@@ -5,7 +5,7 @@
 	padding: 18px;
 }
 .moon-drawer {
-  position: absolute;
-  bottom: 0;
-  width: 100%;
+	position: absolute;
+	bottom: 0;
+	width: 100%;
 }
--- a/css/Drawer.less
+++ b/css/Drawer.less
@@ -4,8 +4,3 @@
 .moon-drawer-client {
 	padding: 18px;
 }
-.moon-drawer {
-	position: absolute;
-	bottom: 0;
-	width: 100%;
-}
--- a/css/Drawer.less
+++ b/css/Drawer.less
@@ -4,3 +4,8 @@
 .moon-drawer-client {
 	padding: 18px;
 }
+.moon-drawer-accelerate {
+  position: absolute;
+  bottom: 0;
+  width: 100%;
+}
--- a/css/Drawers.less
+++ b/css/Drawers.less
@@ -64,6 +64,7 @@
 	top: 0;
 	left: 0;
 	right: 0;
+	z-index: 1;
 
 	.moon-drawers-handles {
 		text-align: center;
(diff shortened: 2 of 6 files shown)


css/Drawers.less
```diff
@@ -85,6 +86,13 @@
 	position: relative;
 	height: 100%;
 	width: 100%;
+	-webkit-transition: -webkit-transform 0.35s ease-out;
+  transition: transform 0.35s ease-out;
```
KunmyonChoi: One minor thing. But, need to use tab instead of space here for indentation.

css/Drawers.less
```diff
@@ -85,6 +86,13 @@
...
+}
+
+.moon-drawers-drawer-container {
+  -webkit-transition: -webkit-transform 0.35s ease-out;
```
KunmyonChoi: Fix indentation.

css/Drawers.less
```diff
@@ -64,6 +64,7 @@
 	top: 0;
 	left: 0;
 	right: 0;
+	z-index: 1;
```
KunmyonChoi: I would suggest reverse the order of component block instead of change z-index here.
You can switch the order of handleContainer and drawers.

css/moonstone-dark.css
```diff
@@ -3595,6 +3595,11 @@ html {
 .moon-drawer-client {
   padding: 0.75rem;
 }
+.moon-drawer {
+  position: absolute;
```
KunmyonChoi: Fix indentation.

css/Drawer.less
```diff
@@ -3,4 +3,9 @@
 }
 .moon-drawer-client {
 	padding: 18px;
-}
\ No newline at end of file
+}
+.moon-drawer {
+  position: absolute;
```
KunmyonChoi: One minor thing. But, need to use tab instead of space here for indentation.

source/Drawer.js
```diff
@@ -184,6 +184,9 @@
 			} else {
 				this.controlsOpenChanged();
 			}
+
+			if (moon.config.accelerate)
+				this.$.controlDrawer.$.client.setShowing(true);
```
KunmyonChoi: It works fine without this code. Can we remove this?

KunmyonChoi:
I tested with partial drawer only without handles just like the same scenario with the music player app.
It looks good and performs well.

source/Drawers.js
```diff
@@ -348,12 +349,42 @@
...
+						this.$.client.spotlightDisabled = true;
+					}
+					this.hideDrawers(ev.originator);
+					this.animate(ev.height);
+				}
+			} else {
+				if (ev.originator instanceof moon.Drawer) {
+					this.updateActivator(true);
+					// Hide client when fullscreen drawer is open so it is not focusable
+					if (ev.originator.getOpen()) {
```
aarontam: Hi @KunmyonChoi, with this conditional when we have set `moon.config.accelerate` to `false`, any drawer that is hidden initially will not be shown. You can see an example of this in the `DrawerSample` when trying to open the partial drawer.

cc @yeram

KunmyonChoi: Aha, thanks. good catch!

ryanjduffy:
PR to master not needed

PR #2086, opened 2015-04-20
ENYO-1225: ProgressButton: Full progress % does not display during transition

anishlg:
### Issue
In the progress button, full progress percentage (100%) never displays. Once reached the progress 90% it is unexpectedly transitions to 'launch'. But expected to show the progress 100% and then translate to 'launch'

### Fix
Called the 'updateProgressPercent()' function when the progress is greater than or equal to 100%

### DIFF
--- a/source/ProgressButton.js
+++ b/source/ProgressButton.js
@@ -216,6 +216,7 @@
 					this.$.bar.applyStyle('-webkit-transform', 'translateX(-' + (100 + offset - progress) + '%)');
 				}
 			} else if (percent >= 100) {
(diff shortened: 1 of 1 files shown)


KunmyonChoi:
It looks like minimal change for fixing this issue. Looks good. @anishlg

KunmyonChoi:
Please add sign-off statement at the bottom of the PR.

KunmyonChoi:
@anishlg , It looks okay.

webOS101:
Not accepting PR's against master

PR #2088, opened 2015-04-20
ENYO-1217: Update scroller boundaries when (re)rendering the repeater.

aarontam:
### Issue
Due to some optimizations in `ScrollMath` we are now calculating the `maxY` value from the current scroll bounds and clamping to that value for the amount of vertical scrolling. In the case of `moon.IntegerPicker`, there is only one item rendered initially, and so outdated scroll bounds are utilized when scrolling to the next value.

### Fix
We now force a recalculation of the scroll boundaries when the flyweight repeater is re-rendered.

### DIFF
--- a/source/IntegerPicker.js
+++ b/source/IntegerPicker.js
@@ -471,7 +471,7 @@
 			this.$.repeater.set('rowOffset', index);
 			this.$.repeater.set('count', count || 1);
 			this.$.repeater.render();
-			this.$.scroller.getStrategy().calcBoundaries();
+			this.$.scroller.remeasure();
 		},
 
 		/**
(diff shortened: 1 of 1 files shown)


rajyavardhanp:
hi @aarontam ,
 but calcBoundaries()  in enyo.TouchScrollStrategy is a private method.!!

aarontam:
Hi @rajyavardhanp, I was trying to perform the least amount of computation necessary here, but we could alternatively change this to call a documented public method i.e. `this.$.scroller.remeasure();`, which will also do some work with the scroll thumbs and stabilization, which we don't absolutely need in the case of `moon.IntegerPicker`.

aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2103.

PR #2089, opened 2015-04-21
Enyo-1171 Add webkit keycode of backspace for keyboard

DavidUM:
Issue
------
HistorySample does not work when we use USB keyboard

Cause
--------
keymap.js did not have keycode for back space key in keyboard

Fix
----
Add keycode into keymap.js

Note
------
'0x8' is webkit keycode of backspace key. It is not reserved from any other keys of remote controller yet. However we'd better get confirm from Architect.

### DIFF
--- a/source/keymap.js
+++ b/source/keymap.js
@@ -12,7 +12,8 @@
 			19  : 'pause',
 			412 : 'rewind',
 			417 : 'fastforward',
-			461 : 'back'
+			461 : 'back',
+			8	: 'back'	//webkit keycode of backspace key in keyboard
 		});
 	}
 })(enyo, this);
--- a/source/keymap.js
+++ b/source/keymap.js
@@ -12,8 +12,7 @@
 			19  : 'pause',
 			412 : 'rewind',
 			417 : 'fastforward',
-			461 : 'back',
-			8	: 'back'	//webkit keycode of backspace key in keyboard
+			461 : 'back'
 		});
 	}
 })(enyo, this);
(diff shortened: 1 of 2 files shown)


source/keymap.js
```diff
@@ -12,7 +12,8 @@
 			19  : 'pause',
 			412 : 'rewind',
 			417 : 'fastforward',
-			461 : 'back'
+			461 : 'back',
+			8	: 'back'	//webkit keycode of backspace key in keyboard
```
otisnate: Please check what does keycode 8 mean in general environment

DavidUM: In case of TV remote controller, there is no key which has key code 8.
As I commented, it means {{back space}} key in keyboard.
That's the everything that I can find from google, collab and some senior engineer from LSM.
How about e-mail to goldfinger-dev mailing list?

aarontam: I'm wondering if we shouldn't add this new mapping in a more general location (as keyCode 8 for "backspace" is native to JavaScript and we probably want this behavior to occur on PC as well); these mappings here are conditionalized for webOS devices. Or alternatively, just handling keyCode 8 directly in `moon.History`, similar to how we are handling keyCode 13 for "enter" throughout the framework in general.

DavidUM: Hi @aarontam  Thank you for your opinion.
I think, it is not PC only. Some user can connect usb keyboard to TV.

aarontam: Hi @DavidUM yep no disagreement there - I think my point was that this is not a webOS-specific keyCode.

lishammel:
Closing out as Jira linked to this has been closed.

PR #2091, opened 2015-04-21
ENYO-1094: Made changes so that pause and time display position correctly when RTL ...

rakesh25:
...applied

Issue:

Pause and Time display positions are swapped when RTL applied.

Fix:

Made required changes in CSS/LESS so to support correct behavior.

Enyo-DCO-1.1-Signed-off-by: Rakesh kumar rakesh25.kumar@lge.com


ryanjduffy:
PR on master isn't required

PR #2092, opened 2015-04-23
DRD-8143: Fix full screen video player layout when enters from inline pl...

KunmyonChoi:
...ayer

### Issue:
- The fullscreen video player layout is broken when enters from inline player.
- This can be reproduced only when we use inline player inside of panels. Because webkit fullscreen is affected by the layout of parent node.

### Fix:
- Clear overflow, padding and margin of all the ancestor when it is full screen.

### DIFF
--- a/css/VideoPlayer.less
+++ b/css/VideoPlayer.less
@@ -89,3 +89,8 @@
 		position: static;
 	}
 }
+:-webkit-full-screen-ancestor:not(iframe) {
+	overflow: visible;
+	padding: 0 !important;
+	margin: 0 !important;
+}
--- a/css/VideoPlayer.less
+++ b/css/VideoPlayer.less
@@ -121,3 +121,11 @@
 	width: 100% !important;
 	height: 100% !important;
 }
+:full-screen-ancestor:not(iframe) {
+	position: absolute !important;
+	overflow: visible !important;
+	padding: 0 !important;
+	margin: 0 !important;
+	width: 100% !important;
+	height: 100% !important;
+}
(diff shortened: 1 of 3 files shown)


Djspaceg:
I assume this set of rules is not combined into one set of selectors because of the bugs with combining them? If not, lets stack up these selectors into one rule-set; like below. Lets also add a "standards" rule, which isn't browser-prefixed: `:full-screen-ancestor:not(iframe)`
```
}
:-webkit-full-screen-ancestor:not(iframe),
:-moz-full-screen-ancestor:not(iframe),
:-ms-full-screen-ancestor:not(iframe),
:-o-full-screen-ancestor:not(iframe),
:full-screen-ancestor:not(iframe) {
  position: absolute !important;
  overflow: visible !important;
  padding: 0 !important;
  margin: 0 !important;
  width: 100% !important;
  height: 100% !important;
}
```

KunmyonChoi:
@Djspaceg , split out rules and add standard rule. Done :)

KunmyonChoi:
Actually stacking up as one rule is not work in this case.

Djspaceg:
That's what I was afraid of. These pseudo/fullscreen selectors are "grouchy".

aarontam:
I *think* the reason stacking doesn't work is because anytime an unrecognized selector is encountered, the browser ignores the associated declaration block.

Also, closing in favor of https://github.com/enyojs/moonstone/pull/2106.

Djspaceg:
That would make sense! Thx @aarontam

PR #2094, opened 2015-04-23
ENYO-1492: Allow for dynamic changing of the enableBackHistoryAPI property

aarontam:
### Issue
There are cases where we would like to dynamically enable/disable the back history API. Currently this is set to a default of `true` and is updated accordingly based on an application flag.

### Fix
We implement a change handler for the `enableBackHistoryAPI` property.

### DIFF
--- a/source/History.js
+++ b/source/History.js
@@ -162,7 +162,6 @@
 		* @private
 		*/
 		init: function() {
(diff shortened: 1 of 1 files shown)


aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2102.

PR #2096, opened 2015-04-26
when disabling moon history, back button should function

dmikey:
Keeps moon.History from hooking the back button when historyAPI is disabled.


DavidUM:
Thanks @dmikey 
When historyAPI is not supported, we don't need to call _initHistoryState().

PR #2100, opened 2015-04-28
Enyo-1171 Add webkit keycode of backspace for keyboard

DavidUM:
Issue
------
HistorySample does not work when we use USB keyboard

Cause
--------
keymap.js did not have keycode for back space key in keyboard

Fix
----
Add keycode into keymap.js

Note
------
'0x8' is webkit keycode of backspace key. It is not reserved from any other keys of remote controller yet. However we'd better get confirm from Architect.


lishammel:
Closing out as jira tied to this has been closed.

PR #2102, opened 2015-04-28
ENYO-1492: Allow for dynamic changing of the enableBackHistoryAPI property

aarontam:
### Issue
There are cases where we would like to dynamically enable/disable the back history API. Currently this is set to a default of `true` and is updated accordingly based on an application flag.

### Fix
We implement a change handler for the `enableBackHistoryAPI` property.


DavidUM:
It looks good. I didn't consider this one. 
Thank you @aarontam

lib/History/History.js
```diff
@@ -158,6 +159,17 @@ var History = module.exports = kind.singleton({
 	},
 
 	/**
+	* @private
+	*/
+	enableBackHistoryAPIChanged: function () {
+		if (this.enableBackHistoryAPI) {
+			window.onpopstate = this.bindSafely(function() {this.popStateHandler();});
```
ryanjduffy: I don't think we want to assign the event handler like this. `onpopstate` is already included in the enyo dispatcher so we could use Signals here. We could also use `dispatcher.listen()` to add the listener. Either would be okay, imo.

aarontam: Ah ha yes good call, updating. :)

lib/History/History.js
```diff
@@ -89,6 +96,12 @@ var History = module.exports = kind.singleton({
 		if (this.enableBackHistoryAPI) {
 			this._initHistoryState();
 		}
+
+		this._popStateHandler = this.bindSafely(function () {
```
ryanjduffy: Not a big deal but there's no need to wrap `popStateHandler` with another function. `this.bindSafely(this.popStateHandler)` should do the trick, right?

aarontam: Whoops, holdover from the original change, updated!

PR #2106, opened 2015-04-29
DRD-8143: Fix full screen video player layout when enters from inline pl...

KunmyonChoi:
...ayer

Issue:
- The fullscreen video player layout is broken when enters from inline player.
- This can be reproduced only when we use inline player inside of panels. Because webkit fullscreen is affected by the layout of parent node.

Fix:
- Clear overflow, padding and margin of all the ancestor when it is full screen.

### DIFF
--- a/lib/VideoPlayer/VideoPlayer.less
+++ b/lib/VideoPlayer/VideoPlayer.less
@@ -89,3 +89,43 @@
 		position: static;
 	}
 }
+:-webkit-full-screen-ancestor:not(iframe) {
+	position: absolute !important;
+	overflow: visible !important;
+	padding: 0 !important;
+	margin: 0 !important;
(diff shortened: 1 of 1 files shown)


aarontam:
Hi @KunmyonChoi, because we can't combine selectors, what if we threw these properties into a mixin instead, to keep it DRY?

aarontam:
Hi @KunmyonChoi, I created a branch off of your branch with a potential mixin change - let me know what you think: https://github.com/enyojs/moonstone/compare/DRD-8143-kunmyonchoi-new...DRD-8143-aarontam.

KunmyonChoi:
I see what you mean by less mixin. :) That is a good thing to know. It is clear to understand. 
Wouldyou like to explain why we need hide css?

PR #2108, opened 2015-04-29
ENYO-1310-Overlay play icon for video/audio files

rajyavardhanp:
Added mixin for GridListImageItem to expose properties like;
type - to identify the type of poster(i.e. video/audio, image).
position - top/bottom alignment.
icon - to customize video play icon.
badgeIconPosition: left/right alignment.
transparency: true/false for background color

### DIFF
--- a/lib/BadgeOverlaySupport/BadgeOverlaySupport.js
+++ b/lib/BadgeOverlaySupport/BadgeOverlaySupport.js
@@ -64,10 +64,26 @@ var
 * ]}
 * ```
 *
+* The item may define a
+* [badgeIconPosition]{@link moon.BadgeOverlaySupport#badgeIconPosition} to
+* specify position of badgeIcon. Possible values are 'left' or 'right'. Defaulted as 'left'.
+*
+* ```javascript
+* {name: 'list', kind: 'moon.DataList', components: [
+*	{mixins: ['moon.BadgeOverlaySupport'], badgeType: 'video', position: 'top', badgeIconPosition: 'right',
+*		kind: 'moon.ImageItem', bindings: [
+*			{from: '.model.title', to: '.label'},
+*			{from: '.model.description', to: '.text'},
+*			{from: '.model.coverSource', to: '.source'}
+*		]
+*	}
+* ]}
+* ```
+*
 * @mixin moon.BadgeOverlaySupport
 * @public
 */
-BadgeOverlaySupport = exports.Support = {
+module.exports = {
 
 	/**
 	* @private
@@ -82,7 +98,6 @@ BadgeOverlaySupport = exports.Support = {
 	/**
 	* Identifies the MIME type of badge.
 	* Possible values are "image", "audio" or "video".
-	* Defaulted as 'image'.
 	*
 	* @name moon.BadgeOverlaySupport#badgeType
 	* @type {String}
@@ -114,7 +129,6 @@ BadgeOverlaySupport = exports.Support = {
 	/**
 	* Sepcifies the alignment position of badge.
 	* Possible values are 'top' or 'bottom'.
-	* Defaulted as 'bottom'.
 	*
 	* @name moon.BadgeOverlaySupport#position
 	* @type {String}
@@ -126,7 +140,6 @@ BadgeOverlaySupport = exports.Support = {
 	/**
 	* Sepcifies the alignment position of badge icon.
 	* Possible values are 'left' or 'right'.
-	* Defaulted as 'left'.
 	*
 	* @name moon.BadgeOverlaySupport#badgeIconPosition
 	* @type {String}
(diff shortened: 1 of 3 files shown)


lib/BadgeOverlaySupport/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,260 @@
...
+		onBadgeIconTap: ''
+	},
+
+	/**
+	* @method
+	* @private
+	*/
+	create: kind.inherit(function (sup) {
+		return function () {
+			this.kindComponents = enyo.Component.overrideComponents(this.kindComponents, this._badgeOverride);
```
aarontam: I think we'll want to change `enyo.Component` to its module equivalent version.

lib/BadgeOverlaySupport/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,260 @@
...
+	*/
+	transparencyChanged: function() {
+		this.addRemoveClass('moon-badge-overlay-transparent', this.transparency);
+	},
+
+	/**
+	* @method
+	* @private
+	*/
+	dispatchEvent: enyo.inherit(function (sup) {
```
aarontam: We'll need to change `enyo.inherit` to `kind.inherit` here.

lib/BadgeOverlaySupport/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,260 @@
...
+		return function (sEventName, oEvent, oSender) {
```
aarontam: Minor nitpick: can we use consistent parameter names i.e. `(nom, ev, sender)`?

lib/BadgeOverlaySupport/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,260 @@
...
+			}
+			return sup.apply(this, arguments);
+		};
+	}),
+
+	/**
+	* @method
+	* @private
+	*/
+	badgeTap: function (inSender, inEvent) {
```
aarontam: Minor nitpick: Can we update the parameter names such that they are not prefixed with "in", i.e. `(sender, ev)`?

lib/BadgeOverlaySupport/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,261 @@
...
+			}
+			return sup.apply(this, arguments);
+		};
+	}),
+
+	/**
+	* @method
+	* @private
+	*/
+	badgeTap: function (nom, ev, sender) {
```
aarontam: Hi @rajyavardhanp I believe this should just be `(sender, ev)` as the standard parameters passed to an event handler.

lib/BadgeOverlaySupport/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,261 @@
...
+*			{from: '.model.coverSource', to: '.source'}
+*		]
+*	}
+* ]}
+* ```
+*
+* @mixin moon.BadgeOverlaySupport
+* @public
+*/
+BadgeOverlaySupport = exports.Support = {
```
ryanjduffy: * `BadgeOverlaySupport` is created as a global var here. You don't need a local ref (afaik) so that can be removed.
* The convention in the framework has been to export the kind or mixin directly rather than as a member on the exported object. So, rather than `exports.Support`, just set the literal to exports

`module.exports = {`

lib/BadgeOverlaySupport/BadgeOverlaySupport.js
```diff
@@ -0,0 +1,261 @@
...
+
+	/**
+	* @private
+	*/
+	classes: 'moon-badge-overlay-support',
+
+	/**
+	* Identifies the MIME type of badge.
+	* Possible values are "image", "audio" or "video".
+	* Defaulted as 'image'.
```
ryanjduffy: No need to state the default in the description. That's why we have the @default tag. Same for `position` and `badgePosition`

lib/BadgeOverlaySupport/BadgeOverlaySupport.less
```diff
@@ -0,0 +1,25 @@
+.moon-badge-overlay-support {
+	.moon-image-client {
```
ryanjduffy: Is it possible to create the same effect without going into the internal classes of .moon-image? Doing so creates a more fragile and interdependent system.

rajyavardhanp: Hi @ryanjduffy . The other way i can think of accessing the client is accessing it via component and adding class (image.$.client). But it looks like breaking the encapsulation. 'moon-image' class can be removed and can be achieved same effect via the code, but 'moon-image-client' , need some more analysis.
Thankyou.

lishammel:
Closing PR per Rajyavardhan; no longer needed.

PR #2111, opened 2015-04-29
ENYO-1194 Set HW acceleration to actual animation element in marqueeText

yeram:
Modify target element that will be accelerated from 'this' to 'this.$.marqueeText'.


yeram:
Wrong one.

PR #2114, opened 2015-04-29
ENYO-1292: ExpandableListItem: Lack of Padding displays in RTL

anishlg:
### Issue:

Lack of padding displays to the right edge of the expandable list item header when in RTL mode

### Fix:

The right padding for expandable picker header was changed from 0px to @moon-spotlight-outset

### DIFF
--- a/lib/ExpandableListItem/ExpandableListItem.less
+++ b/lib/ExpandableListItem/ExpandableListItem.less
@@ -31,7 +31,7 @@
 	}
 }
 .enyo-locale-right-to-left {
-	.moon-expandable-picker-header.moon-expandable-list-item-header {
+	.moon-expandable-list-item-header.moon-expandable-list-header {
 		padding-right: @moon-spotlight-outset;
 	}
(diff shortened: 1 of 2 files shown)


lib/ExpandableListItem/ExpandableListItem.less
```diff
@@ -27,6 +27,9 @@
 .moon-expandable-list-item-client.indented {
 	padding-left: @moon-item-indent + @moon-spotlight-outset;
 }
+.enyo-locale-right-to-left .moon-expandable-list-item-header.moon-expandable-list-item-header {
```
Djspaceg: What's the reasoning for doubling-up the selector classes here?
`.moon-expandable-list-item-header.moon-expandable-list-item-header`
Couldn't just one of these 2 classes do the same job?

anishlg:
Changed the css class name

anishlg:
used expandable list item class instead of expandable picker class

webOS101:
Corrected by ENYO-2144

PR #2120, opened 2015-04-29
[2.6.0-pre.4-dev -> 2.6.0-dev] LightPanels updates and refinement.

aarontam:
"Cherry-picking" changes to `moon.LightPanel` from `2.6.0-pre.4-dev` to `2.6.0-dev`. This PR is related to and dependent upon https://github.com/enyojs/enyo/pull/1140.

### DIFF
--- a/lib/LightPanel/LightPanel.js
+++ b/lib/LightPanel/LightPanel.js
@@ -141,26 +141,16 @@ module.exports = kind(
 	deactivated: function () {},
 
 	/**
-	* Determines whether or not we should perform post-transition work.
-	*
-	* @return {Boolean} - If `true`, we should not run the
-	*	{@link enyo.LightPanels#postTransition} method.
-	* @public
-	*/
-	shouldSkipPostTransition: function () {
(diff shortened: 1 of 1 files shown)


lib/LightPanel/LightPanel.js
```diff
@@ -141,26 +141,16 @@ module.exports = kind(
...
 	* This overridable method is called after a transition.
 	*
 	* @public
 	*/
 	postTransition: function () {
-		this.createComponents(this.clientComponents);
-		this.$.client.render();
-		this.$.client.addClass('populated');
-		this._skipPostTransition = true;
+		if (!this.$.client.children.length) {
```
aarontam: I'm wondering if there is a better or more semantically appropriate way of checking for this condition.

graynorton: Per our discussion on HipChat, we'll merge this for now and take another look later, in the broader context of conditionally creating / diffing components depending on circumstances.

PR #2127, opened 2015-05-05
ENYO-1160: Shadows Pointer Tip of ContextualPopup is Truncated

anishlg:
### Issue:
The shadow pointer tip in contextual popup is found truncated in both dark and light css theme. The issue caused by the improper usage of tip height and width css property. 

### Fix:
The height and width css property of contextual popup tip made 0 to get the proper triangle share. The margin css properties of the tip also changed to accommodate the new change.


ryanjduffy:
Master PR not required

PR #2131, opened 2015-05-06
Enyo 1144 Slider: Insert arrow indicators on tool tip when selecting knob

DavidUM:
Issue
-------
There is UX requirement to add arrow indicator for slider

Implemented
-----------------
Using iconButton which is in simplePIcker, I add arrow buttons into both sides of slider.

### DIFF
--- a/lib/Slider/Slider.js
+++ b/lib/Slider/Slider.js
@@ -340,7 +340,7 @@ module.exports = kind(
 	*/
 	jumpWrapperComponents: [
 		{name: 'buttonLeft', kind: IconButton, noBackground: true, classes: 'moon-simple-picker-button left', icon: 'arrowlargeleft', onSpotlightKeyDown: 'configureSpotlightHoldPulse', onSpotlightSelect: 'previous', ondown: 'downLeft', onholdpulse: 'holdLeft', ondragstart: 'preventDrag', defaultSpotlightDisappear: 'buttonRight'},
-		{name: 'sliderWrapper', classes: 'moon-slider', spotlight: true},
+		{name: 'slider', classes: 'moon-slider', spotlight: true},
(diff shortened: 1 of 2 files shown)


lib/Slider/Slider.js
```diff
@@ -384,12 +419,33 @@ module.exports = kind(
...
+		ProgressBar.prototype.initComponents.apply(this, arguments);
+	},
+
+	/**
+	* @private
+	*/
+	createJumpIncrementButton: function () {
+		this.createComponents(this.jumpWrapperComponents, {owner: this});
+		this.controlParent = this.$.sliderWrapper;
+		this.removeClass('moon-slider');
```
aarontam: Hi @DavidUM, this will cause some problems when the slider is disabled from a styling perspective. We also need to disable the actual slider component (rather than the wrapper) when the control disabled, so that it cannot be spotted when disabled. The semantics were becoming confusing (see https://github.com/enyojs/moonstone/pull/2131/files#diff-90bef7cb477f9b2a2191c4de159c3017R343), as that is the slider component (or at least it has the "moon-slider" class) but its name is "sliderWrapper", while the parent control has the "moon-slider-wrapper" class. I went ahead and took a stab at a change to correct the disabled behavior, as well as renaming the component, here: https://github.com/enyojs/moonstone/compare/Enyo-1144-davidum-2.6.0-dev...ENYO-1144-aarontam. Let me know what you think, thanks!

cc @ryanjduffy

DavidUM: Thank you @aarontam , during my commit, I made a confusing semantics. 
I read your suggestion and agree with it.
Let me cherry-pick your commit to this PR.

DavidUM:
Hi @aarontam I bring your commit and resolve conflict this PR to 2.6.0-dev.
Could you take a look? thank you.

aarontam:
Hi @DavidUM, thanks for the update! I had to make a small tweak so that, when spotted, the control will be scrolled into view when we have set `enableJumpIncrement:true`. I think this looks good now so I'll go ahead and merge it.

PR #2136, opened 2015-05-07
Glyph shapes updated and more international glyphs added.

Djspaceg:


Djspaceg:
This is not intended to be merged. It was for testing purposes only.

PR #2137, opened 2015-05-07
ENYO-1326: update to show scroll thumb after top of scroll content is reached

sbcho:
### Issue
The scroll thumb does not display(via PageUp) after the top of the scroll content is reached.

### Fix
Add conditional statement and call {{alertThumbs}} when it is top/bottom of scroll content.
So the scroll thumb display(via PageUp) after the top/bottom of the scroll content is reached.


lishammel:
Closing out as jira tied to this has been closed.

PR #2138, opened 2015-05-07
ENYO-1331: update PageUp/Down to support horizontal scrolling

sbcho:
### Issue
PageUp/Down Does Not Function in Horizontal Scrolling

### Fix
Update PageUp/Down key to support horizontal scrolling. It scrolls horizontally if app has only horizontal scroll contents, and it go vertically if there is horizontal and vertical scroll contents both.


lishammel:
closing out as jira tied to this has been closed as fixed with another PR.

PR #2140, opened 2015-05-08
ENYO-1443: ExpandablePicker: Pre-Expanded Picker Does Not Collapse Initially after Selection

anishlg:
### Issue:
The selectAndClose() get called if the isDrawerRendered property true. But in initially isDrawerRendered property was became false in the render function as the drawer.renderOnShow is always true. This cause the issue that in activated function this.startJob won't fire as the this.isDrawerRendered is false.

### Fix:
Instead of using  drawer.renderOnShow, use this.getOpen() to identify whether the drawer is  already opened and set isDrawerRendered property true in such cases will solve the issue.

### DIFF
--- a/source/ExpandablePicker.js
+++ b/source/ExpandablePicker.js
@@ -238,8 +238,8 @@
 		* @private
 		*/
 		rendered: function () {
-			this.inherited(arguments);
-			if (!this.$.drawer.renderOnShow) this.isDrawerRendered = true;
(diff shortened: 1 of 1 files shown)


anishlg:
Updated the review comments. 
included !this.$.drawer.renderOnShow checking in the if conditon

aarontam:
Closing in favor of http://github.com/enyojs/moonstone/pull/2141.

PR #2141, opened 2015-05-08
ENYO-1443: ExpandablePicker: Pre-Expanded Picker Does Not Collapse Initially after Selection

anishlg:
### Issue:

The selectAndClose() get called if the isDrawerRendered property true. But in initially isDrawerRendered property was became false in the render function as the drawer.renderOnShow is always true. This cause the issue that in activated function this.startJob won't fire as the this.isDrawerRendered is false.

### Fix:

Instead of using drawer.renderOnShow, use this.getOpen() to identify whether the drawer is already opened and set isDrawerRendered property true in such cases will solve the issue.

### DIFF
--- a/lib/ExpandablePicker/ExpandablePicker.js
+++ b/lib/ExpandablePicker/ExpandablePicker.js
@@ -254,7 +254,7 @@ module.exports = kind(
 	*/
 	rendered: function () {
 		ExpandableListItem.prototype.rendered.apply(this, arguments);
-		if (!this.$.drawer.renderOnShow) this.isDrawerRendered = true;
+		if (this.getOpen()) this.isDrawerRendered = true;
 	},
 
 	/**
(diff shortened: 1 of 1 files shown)


lib/ExpandablePicker/ExpandablePicker.js
```diff
@@ -254,7 +254,7 @@ module.exports = kind(
 	*/
 	rendered: function () {
 		ExpandableListItem.prototype.rendered.apply(this, arguments);
-		if (!this.$.drawer.renderOnShow) this.isDrawerRendered = true;
+		if (this.getOpen()) this.isDrawerRendered = true;
```
sbcho: I think below code seems more safe to guarantee functionality of ENYO-474
{{ if (!this.$.drawer.renderOnShow || this.getOpen()) ... }}
What do you think about it?

DavidUM: Although its default value is true, user can set it false when user instantiates it.
So, I think @sbcho 's suggestion has worth to consider.

anishlg:
Updated the review comments. 
included !this.$.drawer.renderOnShow checking in the if conditon

PR #2143, opened 2015-05-09
ENYO-1589: Fix and enhance post transition behavior

aarontam:
### Issue
We are currently attempting to spot the current panel if there is no spotlight focus, without taking into account pointer mode and/or whether or not the current item is spottable. Additionally, we want to allow for custom logic to be executed once the deferred client components have actually been created and rendered.

### Fix
I've cherry-picked the following commit from @KunmyonChoi to resolve the first issue: https://github.com/enyojs/moonstone/commit/568ef3784e50d4eeaa6284edd8702124e4edb5. And for the second issue, we've followed the pattern of `enyo.DataList` and execute the `didClientRender` method if applicable.

### DIFF
--- a/lib/LightPanel/LightPanel.js
+++ b/lib/LightPanel/LightPanel.js
@@ -146,7 +146,7 @@ module.exports = kind(
 	deactivated: function () {},
 
 	/**
-	* This overridable method is called after a transition.
+	* This overridable (or extendable) method is called after a transition.
 	*
 	* @public
 	*/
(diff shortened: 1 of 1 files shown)


lib/LightPanel/LightPanel.js
```diff
@@ -149,12 +149,23 @@ module.exports = kind(
 		if (!this.$.client.children.length) {
 			this.createComponents(this.clientComponents);
 			this.$.client.render();
-			this.$.client.addClass('populated');
+			this.didClientRender();
 		}
 
-		if (!Spotlight.getCurrent()) {
+		if (!Spotlight.getPointerMode() && !Spotlight.isSpottable(Spotlight.getCurrent(), true)) {
```
ryanjduffy: @KunmyonChoi Can you confirm this condition `!Spotlight.isSpottable(Spotlight.getCurrent(), true)` is what you intended? What scenario were you guarding against?

KunmyonChoi: Hi @ryanjduffy , Sorry for the delayed response.
I don't remember the exact scenario because it was done while I modify huge amount of code on BreadcrumbSupport.js. I think we can skip this spotlight change in this PR for now. So, we can create separate PR if it is needed.

KunmyonChoi: Hi @aarontam , it light panels have no spotlight issue without this change?

aarontam: Hi @KunmyonChoi I've gone a different direction and refactored via this PR (https://github.com/enyojs/moonstone/pull/2277) to address Spotlight issues in a more consistent way, so I don't think we will end up needing this specific change.

KunmyonChoi: Then, I think it will be better to just add didClientRender in this PR. I'll change that code.

PR #2151, opened 2015-05-15
ENYO-1648: Increase moveTolerance value from 16 to 1500 on configureH…

KunmyonChoi:
…oldPulse

Issue:
- The onHoldPulse event is canceled too early when user move mouse abound.

Fix:
- We need higher moveTolerance value to make it stay pulse more on TV.
- Increase moveTolerance value from 16 to 1500 on configureHoldPulse by testing.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,10 +11,7 @@ exports.version = '2.6.0-pre';
 // Override the default holdpulse config to account for greater delays between keydown and keyup
 // events in Moonstone with certain input devices.
 gesture.drag.configureHoldPulse({
-	frequency: 200,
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -15,7 +15,7 @@ gesture.drag.configureHoldPulse({
 	events: [{name: 'hold', time: 400}],
 	resume: false,
 	moveTolerance: 16,
```
aarontam: If we're using `onLeave`, can we remove the explicit declaration of the `moveTolerance` property here?

PR #2155, opened 2015-05-18
ENYO-1288 : Transparent button in image header and full image pattern. (rank: 27)

suhlee:
## Issue
There is a requirement to display transparent buttons in panel's header in Dreadlock

## Solution

 - in Header.less, I added 9 classes from 10-transparency to 90-transparency
 - Framework provides only [10, 20, 30, 40, 50, 60, 70, 80, 90] transparency which can be set.
 - in Header.js, I added two properties ( translucentButtons which can make buttons be transparent and translucentAmount which can set how much transparent).
 - Because class that will be applied from Header.less override spotlight related css, I added '!important' in Button.less and IconButton.less

Enyo-DCO-1.1-signed-off-by: Suhyung Lee suhyung2.lee@lge.com

### DIFF
--- a/css/moonstone-variables-dark.less
+++ b/css/moonstone-variables-dark.less
@@ -21,6 +21,8 @@
 @moon-spotlight-border-color:         @moon-accent;
 @moon-spotlight-color:                @moon-accent;
 @moon-button-background-color:        #4d4d4d;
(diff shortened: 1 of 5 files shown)


suhlee:
closed, because the implementation should be different, 
now we directly apply translucent property to button, iconButton, etc instead of header

PR #2156, opened 2015-05-18
Fixes ENYO-1641

ryanjduffy:
## Issue
We have an existing and 2 in-flight overlay implementations that need to be consolidated into a consistent API.

## Fix
Add moonstone/Overlay


lib/ClampedText/ClampedText.js
```diff
@@ -0,0 +1,166 @@
...
+	/**
+	* @private
+	*/
+	classes: 'moon-clamped-text',
+
+	/**
+	* @lends  moon.ClampedText.prototype
+	* @private
+	*/
+	published: {
```
Djspaceg: Are we still doing `published` blocks? I thought we were transitioning away from them.

ryanjduffy: Dunno, honestly.

lib/Overlay/Overlay.less
```diff
@@ -0,0 +1,88 @@
...
+		line-height: @moon-icon-button-small-size;
+		border: @moon-button-border-width solid black;
+		background-color: white;
+		color: transparent;
+		margin: 0;
+		border-radius: 50%;
+		position: relative;
+		opacity: 0.5;
+
+		.selected & {
```
Djspaceg: What does the trailing `&` do?

lib/ClampedText/ClampedText.js
graynorton: Yeah, we're moving away from them. There was a bit of wavering on this on the next-gen track when we were looking at web components, but I think we're back to "don't use them in new stuff" now. @clinuz, are you on the same page?

lib/Overlay/Overlay.less
ryanjduffy: `&` is always replaced by the active selector. In this case, it adds it after the `.selected` class:

```
/* resulting css */
.selected .moon-overlay-container.moon-overlay-selection .moon-icon { ... }
```

Djspaceg: WOW! That's convenient! I hadn't thought to use it like that before. I'll for sure be taking advantage of that in the future, especially for all of our `.enyo-locale-right-to-left` and locale classes.

KunmyonChoi: Wow! Very handy!

lib/Overlay/Overlay.js
```diff
@@ -0,0 +1,591 @@
...
+		}}
+	],
+
+	/**
+	* @private
+	*/
+	create: kind.inherit(function (sup) {
+		return function () {
+			this.overlayComponents = this.overlayComponents || [
+				{name: 'overlayIcon', kind: Icon, small: false, icon: 'check', owner: this}
```
Djspaceg: The default icon doesn't seem to be getting applied. I just have empty circles in the sample when the sample doesn't specify an `overlayIcon`.

lib/Overlay/Overlay.js
```diff
@@ -0,0 +1,591 @@
...
+	* @private
+	*/
+	bindings: [
+		// to reduce API surface area, we'll only expose overlayIcon and map that to both src and
+		// icon ensuring that only 1 is valued based on the presence of a `.` to indicate a URL
+		{from: 'overlayIcon', to: '$.overlayIcon.src', transform: function (icon) {
+			return (!icon || icon.indexOf('.') >= 0) ? null : icon;
+		}},
+		{from: 'overlayIcon', to: '$.overlayIcon.icon', transform: function (icon) {
+			return (!icon || icon.indexOf('.') == -1) ? null : icon;
```
Djspaceg: This transform seems to be interfering with the ability to set the `overlayIcon` in the sample. Removing the transform allows the icon to display as expected.

lib/Overlay/Overlay.js
```diff
@@ -0,0 +1,591 @@
...
+	*/
+
+	/**
+	* @private
+	*/
+	bindings: [
+		// to reduce API surface area, we'll only expose overlayIcon and map that to both src and
+		// icon ensuring that only 1 is valued based on the presence of a `.` to indicate a URL
+		{from: 'overlayIcon', to: '$.overlayIcon.src', transform: function (icon) {
+			return (!icon || icon.indexOf('.') >= 0) ? null : icon;
```
Djspaceg: This `.` requirement means that icons which reference "http://..." URLs, icons which are targeted using server relative URLs: "/myicon.png", and icon files in the root library directory (referenced with just "lib/badPlaceForAssets/myicon.png", which is technically poor form) will not be supported. This may or may not be a big deal, just something to note which differs from vanilla moon.Icon.

lib/Image/Image.js
```diff
@@ -134,26 +133,25 @@ module.exports = kind(
 		{from: '.position', to: '.$.image.position'}
```
Djspaceg: These should drop their `.` prefix, since you're already in here.

lib/Image/Image.js
```diff
@@ -134,26 +133,25 @@ module.exports = kind(
...
-		if (this.getClientControls().length > 0) {
-			this.$.client.canGenerate = true;
-			this.addClass('has-children');
+		// Adapting from the prior API. We'll assume that the presence of components but not
+		// overlayComponents means the consumer is expecting the former ImageBadge-style overlay
+		if (this.components && !this.overlayComponents) {
+			this.overlayComponents = this.components;
+			this.components = null;
+			this.overlayComponents.forEach(function (c) {
+				c.kind = c.kind || Icon
```
Djspaceg: Missing trailing semicolon.

lib/Image/Image.js
```diff
@@ -134,26 +133,25 @@ module.exports = kind(
...
+			});
+			this.overlayPosition = this.overlayPosition || 'bottom';
```
Djspaceg: Perhaps instead of hard-coding "bottom" here, it should be a property, so subkinds can effectively change it easily.

lib/Image/Image.js
```diff
@@ -11,24 +11,24 @@ var
...
 * {@link moon.Image} is a simple control that wraps an {@link enyo.Image} to provide proper
 * alignment with text-based controls.
 *
-* In addition, `moon.Image` accepts optional {@link moon.ImageBadge} client components
-* (ImageBadge being the default kind of Image), which are placed inside a container positioned
-* over the image.  These badges are normally persistent, but may be shown or hidden based on
-* {@glossary Spotlight} focus, using the
-* [showBadgesOnSpotlight]{@link moon.Image#showBadgesOnSpotlight} property.
+* In addition, `moon.Image` adds {@link moon.Overlay} support to show controls over the image. This
+* can be used to add action icons as in the example below.
```
Djspaceg: It should be documented that `components` of this kind are automatically assumed to be `moon.Icon`s, unless an `overlayComponents` block exists. Also, if `overlayComponents` _do_ exist, they are **not** automatically treated as `moon.Icon`s.

lib/Image/Image.js
ryanjduffy: It can be changed by specifying `overlayPosition` which is the new API. This block is just for adapting from the old API so I do want to expose anything that perpetuates it.

lib/Overlay/Overlay.js
ryanjduffy: I swapped the conditions in these transforms due to a likely copy/paste error. Will fix this up.

lib/Overlay/Overlay.js
ryanjduffy: You might confirm you aren't pulling in the new moonstone font file which changed the code point for 'check' (e.g. you installed it locally and that's what it's using). I'll check things out but I believe I noticed the same behavior and that was the cause since I was reviewing the ticket in the same window.

Djspaceg: I did. It's not font related. The icon property isn't being set at all, and the classes being applied are referring to the LG Icons font class which isn't what's applied when `check` is assigned correctly; it's assigned as the fallback when there is no icon name match. I also tested using a different symbol to rule out the potential code-point change (even though I tried both moonstone font versions) and that had the same effect; an empty content where there should be an HTML entity reference. After removing the binding transform I was able to get at least manual assignment to work, but never the default from the line above.

ryanjduffy: So my assumption was right ... but the person was wrong. I assumed that's what was going on so I didn't worry about it but turns out that because I'm using a binding but specifying the default, `null` was overwriting my value when the binding evaluated. :person_frowning:

Djspaceg:
Let me just state that I think this is an excellent, elegant implementation and replacement for so many different and disjointed controls peppering the Moonstone. We need more controls like this that can generically and effortlessly pull double-duty in the framework.

lib/Image/Image.js
ryanjduffy: Added some doc to this effect.

ryanjduffy:
Thanks, @Djspaceg!

lib/Image/Image.js
```diff
@@ -11,26 +11,30 @@ var
...
+* 		{kind: Icon, src: 'assets/icon-recommended.png'},
+* 		{kind: Icon, icon: 'star'},
+* 		{kind: Icon, src: 'assets/icon-new.png'}
+* 	]
+* }
 * ```
 *
+* For backwards compatibility, component configurations specified using the `components` block
+* within an `moon.Image` will be mapped to `overlayComponents` and will default the `kind` to be
+* {@link moon.Icon}. This behavior is considered deprecated and will be removed in a future release.
```
Djspaceg: If this is deprecated functionality, why ImageBadgeSample.js in the other PR (https://github.com/enyojs/enyo-strawman/pull/22/files#diff-10120f68294655a53b7a059825f90293R25) explicitly being changed _into_ this method, instead of using the newer method?

ryanjduffy: Good point. I was testing the conversion and didn't fix it up before committing. Thanks!

ryanjduffy: Updated!

lib/Overlay/Overlay.js
```diff
@@ -0,0 +1,621 @@
...
+			if (this.overlayPosition) classes.push(mapPosition(this.overlayPosition));
+			if (this.overlayAlign) classes.push(mapAlign(this.overlayAlign));
+			if (this.overlayTransparent) classes.push(mapTransparent(this.overlayTransparent));
+			if (this.overlayShowing) classes.push(mapShowing(this.overlayShowing));
+
+			this.createComponent({name: 'overlayContainer', classes: classes.join(' ')});
+		}
+
+		// wrapper should be owned by this but client controls should be owned by the instance owner
+		var c = this.$.overlayContainer.createComponent({
```
Djspaceg: Lint saw this, so I can't take credit. `c` is declared and never used.

lib/Overlay/Overlay.js
```diff
@@ -0,0 +1,621 @@
...
+	create: kind.inherit(function (sup) {
+		return function () {
+			sup.apply(this, arguments);
+
+			var fn = this.bindSafely(forwardPropertyChange);
+			this.overlayPropertyBindings.forEach(function (p) {
+				this.observe(p, fn);
+				if (utils.exists(this[p])) fn(null, this[p], p);
+			}, this);
+		}
```
Djspaceg: Lint wanted a semicolon to end this return statement.

lib/Overlay/Overlay.js
```diff
@@ -0,0 +1,621 @@
...
+	/**
+	* @private
+	*/
+	create: kind.inherit(function (sup) {
+		return function () {
+			this.overlayComponents = this.overlayComponents || [
+				{name: 'overlayIcon', kind: Icon, small: false, owner: this}
+			];
+			sup.apply(this, arguments);
+		}
```
Djspaceg: Lint wanted a semicolon to end this return statement too.

lib/Overlay/Overlay.js
```diff
@@ -0,0 +1,621 @@
...
+	* @private
+	*/
+	create: kind.inherit(function (sup) {
+		return function () {
+			this.overlayComponents = this.overlayComponents || [
+				{name: 'overlayTitle', classes: 'moon-overlay-text-title', owner: this},
+				{name: 'overlaySubtitle', classes: 'moon-overlay-text-subtitle', owner: this}
+			];
+			sup.apply(this, arguments);
+		}
```
Djspaceg: And finally Lint wanted a semicolon to end this return statement.

ryanjduffy:
Interesting that Travis didn't complain about those lint errors. Do you have a different local config?

Djspaceg:
I believe I do, as in I had only minimally modified my older enyo-style config to include the now globals and drop the old ones (enyo, $L etc), so it's probably more strict.

lib/Image/Image.js
```diff
@@ -114,46 +123,51 @@ module.exports = kind(
...
+	* overlayComponents means the consumer is expecting the former ImageBadge-style overlay.
+	*
+	* @deprecated Backwards compatibility for moon.ImageBadge's within components block
+	* @private
+	*/
+	adaptComponentsBlock: function () {
+		if (this.components && !this.overlayComponents) {
+			this.overlayComponents = this.components;
+			this.components = null;
+			this.overlayComponents.forEach(function (c) {
```
aarontam: From a perf standpoint, would it be faster to use a standard `for` loop, semantics aside?

lib/Overlay/Overlay.js
```diff
@@ -0,0 +1,621 @@
...
+/**
+* Maps the value of `overlayPosition` to the appropriate CSS class
+*
+* @param	{String}	value	Position value
+* @return	{String}			CSS Class
+*
+* @private
+*/
+function mapPosition (value) {
+	return value === 'centered' ? value : 'position-' + value;
```
aarontam: Should we use `==` for consistency here?

lib/Overlay/Overlay.js
```diff
@@ -0,0 +1,621 @@
...
+*
+* @param  {Object[]}		components	Component configs
+* @param  {enyo.Control}	owner		Owner of components
+* @return {Object[]}					Updated component configs
+*
+* @private
+*/
+function updateOwnership (components, owner) {
+	if (components) {
+		components.forEach(function (c) {
```
aarontam: Similar question regarding `for` performance as above.

PR #2160, opened 2015-05-19
ENYO-1649: scrollToControl makes scroller shake on rtl case

KunmyonChoi:
Issue:
- getAbsoluteBounds right value was measured from the right edge before https://github.com/enyojs/enyo/commit/c1189eef1fb245badaa3ac9d22af930cda678183 commit.
- Now right value is measured from the left edge. So, controlBounds.right - absoluteBounds.right is negative value.

Fix:
- Reverse offsetLeft calculation to it can have positive value.


graynorton:
Looks good, @KunmyonChoi! I applied the same change to the equivalent `scrollToNode()` implementation in the new enyo/Scrollable module.

PR #2161, opened 2015-05-19
ENYO-1285: Marquee Text is ellipsis but does not roll when content 

anishlg:
### Issue:
The node.scrollWidth will return the correct value when the text inside the div is more wider than the div visible area. Else the node.scrollWidth will return the div size itself. In this case both the node.scrollWidth and node.getBoundingClientRect().width will be the same and the resultant marquee distance will be 0 so the marquee won't work.

But the text-overflow: ellipsis is applied to the node starting from the div width and inside text content width is same. Which causes the issue.

### Fix:
Remove the text-overflow : ellipsis from the node if the node.scrollWidth and node.getBoundingClientRect().width are same.


lishammel:
Closing out. Per Anish: already another PR created for the same issue

PR #2162, opened 2015-05-19

anishlg:
### issue:
The components outside the contextual popup were not spottable even after the spotlightModal property set to false for the contextual popup. The spotlightModal property doesn't  use for show/hide the scrim which caused this issue. 

### Fix


anishlg:
Changed the contextual popup sample file to include the modal checkbox button to verify the modal property

lishammel:
Closing out. Per Anish: this PR for the master branch and may not require now, only 2.6.0 branch is valid.

PR #2163, opened 2015-05-19

anishlg:
### Issue:

The components outside the contextual popup were not spottable even after the spotlightModal property set to false for the contextual popup. The spotlightModal property doesn't use for show/hide the scrim which caused this issue.

### Fix

### DIFF
--- a/lib/ContextualPopup/ContextualPopup.js
+++ b/lib/ContextualPopup/ContextualPopup.js
@@ -127,6 +127,12 @@ module.exports = kind(
 		*/
 		spotlightModal: false,
 
+		/* @type {Boolean}
+		* @default false
+		* @public
+		*/
+		modal: false,
+
 		/**
 		* If `false`, the close button is hidden; if `true`, it is shown. When this
 		* property is set to `'auto'` (the default), the close button is shown when
@@ -200,6 +206,9 @@ module.exports = kind(
 	*/
 	activator: null,
 
+	bindings: [
+		{from: 'spotlightModal', to: 'modal', oneWay: false}
+	],
 	/**
 	* @private
 	*/
@@ -471,6 +480,13 @@ module.exports = kind(
 	},
 
 	/**
+	* @private
+	*/
+	modalChanged: function() {
+		this.showHideScrim(this.showing);
+	},
+
+	/**
 	* Called when [showCloseButton]{@link moon.ContextualPopup#showCloseButton} changes.
 	*
 	* @private
@@ -533,7 +549,7 @@ module.exports = kind(
 	* @private
 	*/
 	showHideScrim: function (inShow) {
-		if (this.floating && (this.scrim || (this.modal && this.scrimWhenModal))) {
+		if (this.floating && (this.scrim || this.modal || this.scrimWhenModal)) {
 			var scrim = this.getScrim();
 			if (inShow && this.modal && this.scrimWhenModal) {
 				// move scrim to just under the popup to obscure rest of screen
@@ -562,7 +578,7 @@ module.exports = kind(
 		// show a transparent scrim for modal popups if
 		// {@link moon.ContextualPopup#scrimWhenModal} is `true`, else show a
 		// regular scrim.
-		if (this.modal && this.scrimWhenModal) {
+		if (this.scrimWhenModal) {
 			return Scrim.scrimTransparent.make();
 		}
 		return Scrim.make();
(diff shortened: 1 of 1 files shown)


lib/ContextualPopup/ContextualPopup.js
```diff
@@ -471,6 +480,13 @@ module.exports = kind(
 	},
 
 	/**
+	* @private
+	*/
+	modalChanged: function() {
```
KunmyonChoi: Please add space between function and left parenthesis.

lib/ContextualPopup/ContextualPopup.js
```diff
@@ -533,7 +549,7 @@ module.exports = kind(
 	* @private
 	*/
 	showHideScrim: function (inShow) {
-		if (this.floating && (this.scrim || (this.modal && this.scrimWhenModal))) {
+		if (this.floating && (this.scrim || this.modal || this.scrimWhenModal)) {
```
KunmyonChoi: Would you like to explain why we need this fix?
It looks like scrimWhenModal is meaning it should show scrim only when this.modal is true.

lib/ContextualPopup/ContextualPopup.js
```diff
@@ -562,7 +578,7 @@ module.exports = kind(
 		// show a transparent scrim for modal popups if
 		// {@link moon.ContextualPopup#scrimWhenModal} is `true`, else show a
 		// regular scrim.
-		if (this.modal && this.scrimWhenModal) {
+		if (this.scrimWhenModal) {
```
KunmyonChoi: Same as above comment.

lib/ContextualPopup/ContextualPopup.js
anishlg: When the model is true then the if condition will be true and scrim.showAtZIndex() will execute. But when model turns to false the if condition will return false and the scrim.hideAtZIndex() won't execute. the model and scrimWhenModal again verified again inside. ideally the model shouldn't check in this if condition here.

lib/ContextualPopup/ContextualPopup.js
anishlg: When the model is true it will return moon.scrimTransparent.make() and create the scrim and when the model turns to false it will is returning  moon.scrim.make(), but ideally to hide the proper scrim it required to return the same moon.scrimTransparent.make().

KunmyonChoi:
@anishlg , your code is working fine.
But, it looks like the showHideScrim is written with some assumption like... it can be called only from showingChanged. I think we need to break the assumption. 
This is the case that we want to show hide scrim whenever related variables are changed. 
I directly modified code. Please review my change and give any feedback. Thanks.

anishlg:
Hi Kunmyon, your code change is perfectly working. I have checked the code change with the changes made to the sample in enyo-strawman, both working fine.

KunmyonChoi:
@anishlg , would you like to apply the same change into PRs for master?

graynorton:
@KunmyonChoi, @anishlg, we generally shouldn't be making any PRs to master for right now. Everything from 2.6.0-dev will get collapsed back into master together (probably within the next couple of months), so master is effectively frozen for now.

KunmyonChoi:
@anishlg , then would you like to close the PR for master?

lib/ContextualPopup/ContextualPopup.js
```diff
@@ -127,6 +127,12 @@ module.exports = kind(
 		*/
 		spotlightModal: false,
 
+		/* @type {Boolean}
+		* @default false
+		* @public
+		*/
+		modal: false,
```
aarontam: Is this needed, if the default value is also `false` in `enyo/Popup`? Also, we should consider cleaning up the documentation here if we choose to keep it (adding a description, and not placing any pragmas on the opening line of the comment block, which should begin with /** instead of /*).

lib/ContextualPopup/ContextualPopup.js
```diff
@@ -532,18 +545,32 @@ module.exports = kind(
 	/**
 	* @private
 	*/
+	updateScrim: function (old, value, source) {
+		// We sync modal and spotlightModal here because binding doesn't guarantee sequence
+		// when it is used with observers.
+		if (source == 'modal') this.set('spotlightModal', this.modal);
```
aarontam: Seems like this will trigger `updateScrim` multiple times as we are setting the value of another of its observed properties - would like to avoid that if possible.

lib/ContextualPopup/ContextualPopup.js
KunmyonChoi: I don't think this is needed. So, let's remove these lines.

KunmyonChoi:
@aarontam , created clean PR here. https://github.com/enyojs/moonstone/pull/2704

PR #2166, opened 2015-05-21
Enyo-1313-InputHeader: Placeholder Does Not Display Focus State

srinivasv123:
Issue: when we hover on the input field of Header, placeholder and input
text is not getting the focus.

Cause: lack of css classes.

Fix: added required CSS classes to achieve the same.

DCO-1.1-Signed-Off-By: Srinivas V srinivas.v@lge.com

### DIFF
--- a/lib/InputHeader/InputHeader.less
+++ b/lib/InputHeader/InputHeader.less
@@ -49,3 +49,9 @@
 .moon-input-header-input-decorator.spotlight.moon-focused > .moon-input::-moz-input-placeholder {
 	color: @moon-input-header-placeholder-color;
 }
+.moon-input-header-input-decorator.spotlight > .moon-input::-webkit-input-placeholder {
+	color: @moon-spotlight-text-color;
+}
+.moon-input-header-input-decorator.spotlight.moon-focused > .moon-input::-webkit-input-placeholder {
+	color: @moon-input-header-placeholder-color;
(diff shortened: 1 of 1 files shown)


aarontam:
It's not so much a lack of CSS classes as it is something might changed with the precedence or the ordering of the classes. I think I'd want to know what exactly changed, as these rules were previously being applied correctly: https://github.com/enyojs/moonstone/blob/2.6.0-dev/lib/InputHeader/InputHeader.less#L37.

aarontam:
I also noticed that the alignment of the placeholder text is higher than it should be, so I think we'll want to investigate more and come up with a complete solution to address all regressions related to the placeholder text.

aarontam:
Ah ha! When we merged this fix (https://github.com/enyojs/moonstone/commit/6a53ff295757b636e098826e4116df936431d24e), we missed the fact that we were mixing vendor-specific pseudo classes in the selectors; any unrecognized selectors will cause the browser to stop parsing that rule. I'll create another PR that correctly integrates the changes from this commit without breaking our existing rules.

lishammel:
@aarontam and @srinivasv123  - can this PR be closed out?  Looks like it was fixed with another PR? Thanks.

aarontam:
Hi @lishammel, yes this can be closed, thanks!

PR #1148, opened 2014-02-25
GF-61401: Request scroll into view from onSpotlightFocusED instead of on...

kevinpschaaf:
...SpotlightFocus, to prevent spot-reentrance.

DCO-1.1-Signed-Off-By: Kevin Schaaf <kevin.schaaf@lge.com>

PR #2167, opened 2015-05-21
ENYO-1054: Accessibility: add text to speech accessibility support to…

jaewonjang:
… moonstone widgets, part 1

Apply accessibilty on moonstone ExpandablePicker and VideoPlayer.

ExpandablePicker:  When spotlight focused on Expandablepicker, the dom
which is focused has aria-label including only the content of
ExpandablePicker. In this case, the currentValue should be added in
aria-label as well, so observing the content of currentValue and add it
using accessibilityHint.

VideoPlayer: set accessibilityLabel to each icon button according to UX
document.

https://jira2.lgsvl.com/browse/ENYO-1054

### DIFF
--- a/lib/ExpandablePicker/ExpandablePickerAccessibilitySupport.js
+++ b/lib/ExpandablePicker/ExpandablePickerAccessibilitySupport.js
@@ -13,15 +13,15 @@ module.exports = {
 	initAccessibility: kind.inherit(function (sup) {
 		return function (props) {
 			sup.apply(this, arguments);
-			this.$.currentValue.observe('content', this.valueChanged, this);
-			this.valueChanged();
+			this.$.currentValue.observe('content', this.changeLabel, this);
+			this.changeLabel();
 		};
 	}),
 
 	/**
 	* @private
 	*/
-	valueChanged: function() {
+	changeLabel: function() {
 		this.set('accessibilityHint', this.$.currentValue.getContent());
 	}
 };
--- a/lib/ExpandablePicker/ExpandablePickerAccessibilitySupport.js
+++ b/lib/ExpandablePicker/ExpandablePickerAccessibilitySupport.js
@@ -1,0 +1,27 @@
+var
+	kind = require('enyo/kind');
+
+/**
+* @name ExpandablePickerAccessibilityMixin
+* @mixin
+*/
+module.exports = {
+
+	/**
+	* @private
+	*/
+	initAccessibility: kind.inherit(function (sup) {
+		return function (props) {
+			sup.apply(this, arguments);
+			this.$.currentValue.observe('content', this.valueChanged, this);
+			this.valueChanged();
+		};
+	}),
+
+	/**
+	* @private
+	*/
+	valueChanged: function() {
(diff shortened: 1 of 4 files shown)


jaewonjang:
This PR has dependency on #1148 and https://github.com/enyojs/spotlight/pull/176

seunghoh:
1st reviewed. Looks good to me

lib/ExpandablePicker/ExpandablePickerAccessibilitySupport.js
```diff
@@ -0,0 +1,27 @@
...
+			sup.apply(this, arguments);
+			this.$.currentValue.observe('content', this.valueChanged, this);
+			this.valueChanged();
+		};
+	}),
+
+	/**
+	* @private
+	*/
+	valueChanged: function() {
```
ryanjduffy: We should avoid adding methods in the form xxxChanged that aren't property change handlers for properties on this object. If this, or a subkind, added a value property (or really, just called `set('value', 0)`), this method would be automatically called unintentionally.

lib/VideoPlayer/VideoPlayerAccessibilitySupport.js
```diff
@@ -0,0 +1,58 @@
...
+		};
+	}),
+
+	/**
+	* @private
+	*/
+	updatePlayPauseButtons: kind.inherit(function (sup) {
+		return function (props) {
+			sup.apply(this, arguments);
+			if (this._isPlaying) {
```
ryanjduffy: This could be a bit more efficient by choosing the label in the if block or via ternary op and assigning it after.

```
var label = this._isPlaying ? $L('Pause') : $L('Play');
this.$.fsPlayPause.set('accessibilityLabel', label);
this.$.ilPlayPause.set('accessibilityLabel', label);
```

lib/VideoPlayer/VideoPlayerAccessibilitySupport.js
```diff
@@ -0,0 +1,58 @@
...
+	}),
+
+	/**
+	* @private
+	*/
+	moreButtonTapped: kind.inherit(function (sup) {
+		return function (props) {
+			sup.apply(this, arguments);
+			var index = this.$.controlsContainer.getIndex();
+			if (index === 0) {
```
ryanjduffy: Same here

```javascript
var label = index === 0 ? $L('More') : $L('Less');
this.$.moreButton.set('accessibilityLabel', label);
```

lib/VideoPlayer/VideoPlayerAccessibilitySupport.js
```diff
@@ -0,0 +1,58 @@
...
+*/
+module.exports = {
+
+	/**
+	* @private
+	*/
+	initAccessibility: kind.inherit(function (sup) {
+		return function (props) {
+			sup.apply(this, arguments);
+			this.$.jumpBack.set('accessibilityLabel', $L('jumpBack'));
```
ryanjduffy: The convention in Moonstone is to use the English version of the text within `$L()` rather than a key. So 'Fast Forward' instead of 'fastForward'. We should confirm what that text should be as well.

PR #2168, opened 2015-05-21
Fixes ENYO-1677

ryanjduffy:
## Issue
In the new module system, there's no obvious need to have singletons as their purpose was to make a globally available instance. Instead, the kind can be instanced and exposed from the module. Now, the singletons are being exported into the global namespace unnecessarily potentially allowing usage without importing the required module (which is fragile).

# Fix
Remove singleton usage in moonstone/History


ryanjduffy:
Not needed

PR #2170, opened 2015-05-22
Wp 774 TimePicker support Ethiopia 

DavidUM:
Issue
-------
Make timePicker support Ethiopia time

Fix
----
Current timePicker does not consider that meridiem has some entries more than two -AM & PM-
Fix related logic

### DIFF
--- a/source/DatePicker.js
+++ b/source/DatePicker.js
@@ -232,7 +232,7 @@
 		pickerChanged: function (sender, inEvent) {
 			if(this.syncingPickers) return true;
 
-			if (typeof ilib !== 'undefined') {			
+			if (typeof ilib !== 'undefined') {
 				if (sender == this.$.day) this.value.setDays(inEvent.value);
 				else if (sender == this.$.month) this.value.setMonths(inEvent.value);
 				else this.value.setYears(inEvent.value);
--- a/source/DatePicker.js
+++ b/source/DatePicker.js
@@ -229,47 +229,27 @@
 		/**
 		* @private
 		*/
-		pickerChanged: function (inSender, inEvent) {
+		pickerChanged: function (sender, inEvent) {
 			if(this.syncingPickers) return true;
 
-			var day = this.$.day.getValue(),
-				month = this.$.month.getValue(),
-				year = this.$.year.getValue(),
-				value = this.value ? this.value : null,
-				maxDays, localeValue;
-			var valueHours = value ? value.getHours() : 0;
-			var valueMinutes = value ? value.getMinutes() : 0;
-			var valueSeconds = value ? value.getSeconds() : 0;
-			var valueMilliseconds = value ? value.getMilliseconds() : 0;
-
-			if (typeof ilib !== 'undefined') {
-				maxDays = this.monthLength(year, month);
-				localeValue = ilib.Date.newInstance({
-					day: (day <= maxDays) ? day : maxDays,
-					month: month,
-					year: year,
-					hour: valueHours,
-					minute: valueMinutes,
-					second: valueSeconds,
-					millisecond: valueMilliseconds
-				});
-				this.setValue(localeValue);
+			if (typeof ilib !== 'undefined') {			
+				if (sender == this.$.day) this.value.setDays(inEvent.value);
+				else if (sender == this.$.month) this.value.setMonths(inEvent.value);
+				else this.value.setYears(inEvent.value);
 			} else {
-				maxDays = this.monthLength(year, month);
-				this.setValue(new Date(year, month-1, (day <= maxDays) ? day : maxDays,
-					valueHours,
-					valueMinutes,
-					valueSeconds,
-					valueMilliseconds));
+				if (sender == this.$.day) this.value.setDate(inEvent.value);
+				else if (sender == this.$.month) this.value.setMonth(inEvent.value - 1);
+				else this.value.setYear(inEvent.value);
 			}
 
+			this.valueChanged();
 			return true;
 		},
 
 		/**
 		* @private
 		*/
-		setChildPickers: function (inOld) {
+		setChildPickers: function () {
 			if (this.value) {				
 				var values = this.calcPickerValues();
 
--- a/source/DatePicker.js
+++ b/source/DatePicker.js
@@ -106,16 +106,6 @@
 		},
 
 		/**
-		* @private
-		*/
-		initILib: function () {
-			this.inherited(arguments);
-			if (typeof ilib !== 'undefined' && this.value) {
-				this.localeValue = ilib.Date.newInstance({unixtime: this.value.getTime(), timezone: "local"});
-			}
-		},
-
-		/**
 		 * When [iLib]{@glossary ilib} is supported, calculates the minimum year in the
 		 * current calendar. Otherwise, returns the value of the published property
 		 * [minYear]{@link moon.DatePicker#minYear}.
@@ -246,7 +236,7 @@
 				month = this.$.month.getValue(),
 				year = this.$.year.getValue(),
 				value = this.value ? this.value : null,
-				maxDays;
+				maxDays, localeValue;
 			var valueHours = value ? value.getHours() : 0;
 			var valueMinutes = value ? value.getMinutes() : 0;
 			var valueSeconds = value ? value.getSeconds() : 0;
@@ -254,7 +244,7 @@
 
 			if (typeof ilib !== 'undefined') {
 				maxDays = this.monthLength(year, month);
-				this.localeValue = ilib.Date.newInstance({
+				localeValue = ilib.Date.newInstance({
 					day: (day <= maxDays) ? day : maxDays,
 					month: month,
 					year: year,
@@ -263,7 +253,7 @@
 					second: valueSeconds,
 					millisecond: valueMilliseconds
 				});
-				this.setValue(this.localeValue);
+				this.setValue(localeValue);
 			} else {
 				maxDays = this.monthLength(year, month);
 				this.setValue(new Date(year, month-1, (day <= maxDays) ? day : maxDays,
@@ -280,11 +270,7 @@
 		* @private
 		*/
 		setChildPickers: function (inOld) {
-			if (this.value && typeof ilib !== 'undefined') {
-				this.localeValue = ilib.Date.newInstance({unixtime: this.value.getTime(), timezone: "local"});
-			}
-
-			if (this.localeValue || this.value) {
+			if (this.value) {				
 				var values = this.calcPickerValues();
 
 				this.$.year.set('value', values.fullYear);
@@ -305,16 +291,18 @@
 				maxDays: 31
 			};
 			if (typeof ilib !== 'undefined') {
-				if (this.localeValue) {
-					values.fullYear = this.localeValue.getYears();
-					values.month = this.localeValue.getMonths();
-					values.date = this.localeValue.getDays();
+				// if ilib exists, this.value is instance of ilib.Date
+				if (this.value) {
+					values.fullYear = this.value.getYears();
+					values.month = this.value.getMonths();
+					values.date = this.value.getDays();
 				}
 				if (values.fullYear) {
 					values.maxMonths = this._tf.cal.getNumMonths(values.fullYear);
 					values.maxDays = this.monthLength(values.fullYear, values.month);
 				}
 			} else {
+				// if ilib doesn't exist, this.value is instance of JS Date
 				if (this.value) {
 					values.fullYear = this.value.getFullYear();
 					values.month = this.value.getMonth()+1;
--- a/source/DatePicker.js
+++ b/source/DatePicker.js
@@ -245,11 +245,12 @@
 			var day = this.$.day.getValue(),
 				month = this.$.month.getValue(),
 				year = this.$.year.getValue(),
+				value = this.value ? this.value : null,
 				maxDays;
-			var valueHours = this.value ? this.value.getHours() : 0;
-			var valueMinutes = this.value ? this.value.getMinutes() : 0;
-			var valueSeconds = this.value ? this.value.getSeconds() : 0;
-			var valueMilliseconds = this.value ? this.value.getMilliseconds() : 0;
+			var valueHours = value ? value.getHours() : 0;
+			var valueMinutes = value ? value.getMinutes() : 0;
+			var valueSeconds = value ? value.getSeconds() : 0;
+			var valueMilliseconds = value ? value.getMilliseconds() : 0;
 
 			if (typeof ilib !== 'undefined') {
 				maxDays = this.monthLength(year, month);
@@ -262,7 +263,7 @@
 					second: valueSeconds,
 					millisecond: valueMilliseconds
 				});
-				this.setValue(this.localeValue.getJSDate());
+				this.setValue(this.localeValue);
 			} else {
 				maxDays = this.monthLength(year, month);
 				this.setValue(new Date(year, month-1, (day <= maxDays) ? day : maxDays,
--- a/source/TimePicker.js
+++ b/source/TimePicker.js
@@ -492,7 +492,7 @@
 					if (this.meridiemEnable === true) {
 						this.createComponent(
 							{classes: 'moon-date-picker-wrap', components:[
-								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', meridiems: this.meridiems || [{name: 'AM', start: '00:00', end: '11:59'}, {name: 'PM', start: '12:00', end: '23:59'}], max: this.meridiems.length - 1 || 1, onChange: 'meridiemPickerChanged'},
+								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', meridiems: this.meridiems || [{name: 'AM', start: '00:00', end: '11:59'}, {name: 'PM', start: '12:00', end: '23:59'}], max: this.meridiems? this.meridiems.length - 1 : 1, onChange: 'meridiemPickerChanged'},
 								{name: 'meridiemLabel', content: this.meridiemText, classes: 'moon-date-picker-label moon-divider-text'}
 							]}
 						);
--- a/source/TimePicker.js
+++ b/source/TimePicker.js
@@ -424,16 +424,17 @@
 		*
 		* @param  {Number} hour - hour between 0 to 23
 		* @param  {Number} minute - minute between 0 to 59
-		* @return  {Number} index - index of this.meridiems
+		* @return {Number} 		- index of this.meridiems
 		*
 		* @private
 		*/
 		getMeridiemIndex: function (hour, minute) {
-			var meridiems = this.meridiems;
+			var meridiems = this.meridiems,
+				start, end, time;
 			for (var i = 0; i < meridiems.length; i++) {
-				var start = parseInt(meridiems[i]['start'].substring(0,2) + meridiems[i]['start'].substring(3,5), 10);
-				var end = parseInt(meridiems[i]['end'].substring(0,2) + meridiems[i]['end'].substring(3,5), 10);
-				var time = hour * 100 + minute;
+				start = parseInt(meridiems[i]['start'].substring(0,2) + meridiems[i]['start'].substring(3,5), 10);
+				end = parseInt(meridiems[i]['end'].substring(0,2) + meridiems[i]['end'].substring(3,5), 10);
+				time = hour * 100 + minute;
 					
 				if ( start <= time && time <= end) { 
 					return i; 
@@ -574,7 +575,7 @@
 			var hour = this.value.getHours() + (event.value - event.old);
 
 			if (this.value) {
-				this.updateValue('hour', hour);
+				this.updateValue(hour);
 			}
 
 			return true;
@@ -603,26 +604,31 @@
 			if(this.syncingPickers) return true;
 
 			var meridiems = this.meridiems,
-				hour = this.$.hour.getValue(),
-				oldIndex = event.old,
-				newIndex = this.$.meridiem.getValue(),
-				offset, start, end, newHour;
+				oldMeridiem = meridiems[event.old],
+				newMeridiem = meridiems[this.$.meridiem.get('value')],
+				oldHour = this.$.hour.get('value'),
+				startHour = parseInt(newMeridiem['start'], 10),
+				offset = oldHour - parseInt(oldMeridiem['start'], 10);
 
-			offset = hour - parseInt(meridiems[oldIndex]['start'], 10);
-			start = parseInt(meridiems[newIndex]['start'], 10);
-			end = parseInt(meridiems[newIndex]['end'], 10);
-			newHour = offset + start;
-
-			if (newHour != hour) {
-				if (newHour > end) {
-					newHour = end;
-				}
-				this.updateValue('hour', this.value.getHours() + (newHour - hour));
+			if (meridiems.length == 2) {
+				this.updateValue(startHour + offset);
 			} else {
-				// there is only single case. 
-				// Ethiopia calendar has different meridiem status between 06:00 and 06:01
-				start = parseInt(meridiems[newIndex]['start'].split(':')[1], 10);
-				this.updateValue('minute', this.value.getMinutes() + start);
+				var oldMinute =  this.$.minute.get('value'),
+					endHour = parseInt(newMeridiem['end'], 10),
+					startMinute = parseInt(newMeridiem['start'].split(':')[1], 10),
+					endMinute = parseInt(newMeridiem['end'].split(':')[1], 10),
+					newHour = startHour + offset, 
+					newMinute;
+				
+				if (startHour * 100 + startMinute > newHour * 100 + oldMinute) {
+					newHour = startHour;
+					newMinute = startMinute;
+				} else if (endHour * 100 + endMinute < newHour * 100 + oldMinute) {
+					newHour = endHour;
+					newMinute = endMinute;
+				}
+				offset = newHour - oldHour;
+				this.updateValue(this.value.getHours() + offset, newMinute);
 			}
 
 			return true;
@@ -639,14 +645,11 @@
 		/**
 		* @private
 		*/
-		updateValue: function (pickerName, newValue) {
+		updateValue: function (newHour, newMinute) {
 			var valueTime = this.value.getTime();
 
-			if (pickerName == 'hour') {
-				this.value.setHours(newValue);	
-			} else {
-				this.value.setMinutes(newValue);
-			}
+			if (newHour !== undefined) { this.value.setHours(newHour); } 
+			if (newMinute !== undefined) { this.value.setMinutes(newMinute); }
 
 			// in the rare case that the value didn't change because it was snapped back to the
 			// same value due to DST rules, push it back another hour.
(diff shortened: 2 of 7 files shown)


source/TimePicker.js
```diff
@@ -419,9 +442,11 @@
 					fmtParams.locale = this.locale;
 				}
 				var merFormatter = new ilib.DateFmt(fmtParams);
-				var am = ilib.Date.newInstance({hour:10});
-				var pm = ilib.Date.newInstance({hour:14});
-				this.meridiems = [merFormatter.format(am), merFormatter.format(pm)];
+				if (merFormatter.meridiems == 'chinese') {
+					merFormatter.meridiems = 'default';
+				}
+
```
DavidUM: Hi @graynorton, this part is locale-specific code which I referred in Jira comment.  
Chinese and Ethiopia are only two locale which has some meridiem states more than AM&PM.
However, in last year, we decided to not display whole meridiem unless there is strong request.

This time, Ethiopia asked to show every meridiem and Chinese did not.
So I need to use locale-specific code.

source/TimePicker.js
```diff
@@ -470,10 +492,11 @@
 					if (this.meridiemEnable === true) {
 						this.createComponent(
 							{classes: 'moon-date-picker-wrap', components:[
-								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', value: valueHours > 12 ? 1 : 0, meridiems: this.meridiems || ['am','pm'], onChange: 'meridiemPickerChanged'},
+								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', meridiems: this.meridiems || [{name: 'AM', start: '00:00', end: '11:59'}, {name: 'PM', start: '12:00', end: '23:59'}], max: this.meridiems.length - 1 || 1, onChange: 'meridiemPickerChanged'},
```
aarontam: Hi @DavidUM, there is no `meridiems` property on this kind by default, so among other things, this will not work if `ilib` is not present (as the `ilib` initialization sets the `meridiems` property). We either need more guard code, or should set the initial value of `meridiems` to the defaults you added here.

source/TimePicker.js
```diff
@@ -628,11 +670,11 @@
 		*/
 		setChildPickers: function (inOld) {
 			if (this.value) {
-				var hour = this.value.getHours();
-				this.$.hour.setValue(hour);
-				this.$.minute.setValue(this.value.getMinutes());
-				if (this.meridiemEnable === true) {
-					this.$.meridiem.setValue(hour > 11 ? 1 : 0);
+				var values = this.calcPickerValues();
+				this.$.hour.setValue(values.hour);
```
aarontam: Let's use the generic setter for this and the following lines i.e. `this.$.hour.set('value', values.hour);`

source/TimePicker.js
```diff
@@ -585,14 +608,29 @@
...
-				// reset the hour to < 12 and then add 12 if it's pm
-				hour = hour%12 + value*12;
-				this.updateHours(hour);
+			var meridiems = this.meridiems,
+				hour = this.$.hour.getValue(),
+				oldIndex = event.old,
+				newIndex = this.$.meridiem.getValue(),
+				offset, start, end, newHour;
+
+			/*jshint -W030 */
```
aarontam: Why is this needed?

source/TimePicker.js
```diff
@@ -53,14 +53,37 @@
...
+		*
+		* @param  {Number} hour - hour between 0 to 23
+		* @param  {Number} minute - minute between 0 to 59
+		* @return  {Number} index - index of this.meridiems
+		* @private
+		*/
+		getMeridiemIndex: function (hour, minute) {
+			var meridiems = this.meridiems;
+			for (var i = 0; i < meridiems.length; i++) {
+				var startArr = meridiems[i]['start'].substring(0,2) + meridiems[i]['start'].substring(3,5),
```
aarontam: Is there a reason why we're splitting on ":" in some places and using `substring` with specific indices in other locations, like this?

source/TimePicker.js
```diff
@@ -470,10 +492,11 @@
 					if (this.meridiemEnable === true) {
 						this.createComponent(
 							{classes: 'moon-date-picker-wrap', components:[
-								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', value: valueHours > 12 ? 1 : 0, meridiems: this.meridiems || ['am','pm'], onChange: 'meridiemPickerChanged'},
+								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', meridiems: this.meridiems || [{name: 'AM', start: '00:00', end: '11:59'}, {name: 'PM', start: '12:00', end: '23:59'}], max: this.meridiems.length - 1 || 1, onChange: 'meridiemPickerChanged'},
 								{name: 'meridiemLabel', content: this.meridiemText, classes: 'moon-date-picker-label moon-divider-text'}
 							]}
 						);
+						this.$.meridiem.value = this.$.meridiem.getMeridiemIndex(values.hour, values.minute);
```
aarontam: Is it possible to set the value when creating the `MeridiemPicker`, rather than after the fact?

source/TimePicker.js
```diff
@@ -585,14 +608,29 @@
...
+			/*jshint +W030 */
+
+			if (newHour != hour) {
+				if (newHour > end) {
+					newHour = end;
+				}
+				this.updateValue('hour', this.value.getHours() + (newHour - hour));
+			} else {
+				// there is only single case. 
+				// Ethiopia calendar has different meridiem status between 06:00 and 06:01
```
aarontam: I'm noticing the meridiem changing between 05:59 and 06:00, instead of 06:00 and 06:01, as stated.

source/TimePicker.js
DavidUM: Oh yes. I should change variable name.

And I realize that old code are remained. 
During some git processing, I made mistake

source/TimePicker.js
aarontam: Never mind, was in the wrong meridiem (can't read this language!).

source/TimePicker.js
DavidUM: We should use `getMeridiemIndex()` to find meridiem value.
However, this method is API of moon.MeridiemPicker.
So I placed this line after creation.

And to ignore notice, I assigned its value instead of using setter.

source/TimePicker.js
DavidUM: Right. Neither to me.
I just compared its shape. haha.

aarontam: Ok here's a question for you :smile: When using one of the other meridiems (as I was), is it correct that the meridiem changes between 05:59 and 06:00?

source/TimePicker.js
aarontam: Ah right, I guess I was more wondering about the ergonomics and encapsulation here. It seems like we might be able to pass something like an `hour` and a `minute` parameter to `MeridiemPicker` and have it compute its own value internally, rather than set the value of `MeridiemPicker` from another kind, based on this other kind calling a method of `MeridiemPicker` (similarly, on L675). It seems like `getMeridiemIndex` would be more of an internal method for `MeridiemPicker`.

source/TimePicker.js
DavidUM: Yes! I leave meridiem info for Ethiopia for you.

{ name: 'ጠዋት',
    start: '00:00',
    end: '05:59' },
  { name: 'ቀትር',
    start: '06:00',
    end: '06:00' },
  { name: 'ከሰዓት',
    start: '06:01',
    end: '11:59' },
  { name: 'ከምሽቱ',
    start: '12:00',
    end: '17:59' },
  { name: 'ከሌሊቱ',
    start: '18:00',
    end: '23:59' } ]

That's so interesting :)

source/TimePicker.js
```diff
@@ -628,35 +668,64 @@
...
+
+			return values;
+		},
+
+		/**
+		* @private
+		*/
+		valueChanged: function (old) {
+			if (typeof ilib !== 'undefined') {
+				this.localeValue = ilib.Date.newInstance({unixtime: this.value.getTime(), timezone: 'local'});
```
aarontam: I think there is a case where we would change the value to `null` (as in to reset the `TimePicker`). Currently this would throw an error.

aarontam:
I found some odd behavior when changing the meridiem value via the picker. Using the "down" button of the meridiem picker, the meridiem value animates back to the same value, while the hour decrements, which seems strange, if not incorrect. Can you verify?

source/TimePicker.js
```diff
@@ -470,10 +492,11 @@
 					if (this.meridiemEnable === true) {
 						this.createComponent(
 							{classes: 'moon-date-picker-wrap', components:[
-								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', value: valueHours > 12 ? 1 : 0, meridiems: this.meridiems || ['am','pm'], onChange: 'meridiemPickerChanged'},
+								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', meridiems: this.meridiems || [{name: 'AM', start: '00:00', end: '11:59'}, {name: 'PM', start: '12:00', end: '23:59'}], max: this.meridiems? this.meridiems.length - 1 : 1, onChange: 'meridiemPickerChanged'},
 								{name: 'meridiemLabel', content: this.meridiemText, classes: 'moon-date-picker-label moon-divider-text'}
 							]}
 						);
+						this.$.meridiem.value = this.$.meridiem.getMeridiemIndex(values.hour, values.minute);
```
aarontam: Just mirroring my previous comment here as it got list in the previous diff:

Ah right, I guess I was more wondering about the ergonomics and encapsulation here. It seems like we might be able to pass something like an hour and a minute parameter to MeridiemPicker and have it compute its own value internally, rather than set the value of MeridiemPicker from another kind, based on this other kind calling a method of MeridiemPicker (similarly, on L675). It seems like getMeridiemIndex would be more of an internal method for MeridiemPicker. At the very least, it seems like we could encapsulate the setting of the value inside the MeridiemPicker, and expose a public method to call, that takes an hour and minute parameter.

source/TimePicker.js
aarontam: Thanks! This is super helpful.

source/TimePicker.js
```diff
@@ -585,14 +608,27 @@
...
+			offset = hour - parseInt(meridiems[oldIndex]['start'], 10);
+			start = parseInt(meridiems[newIndex]['start'], 10);
+			end = parseInt(meridiems[newIndex]['end'], 10);
+			newHour = offset + start;
+
+			if (newHour != hour) {
+				if (newHour > end) {
+					newHour = end;
+				}
+				this.updateValue('hour', this.value.getHours() + (newHour - hour));
```
aarontam: It seems like this triggers a value update for `TimePicker`, which eventually calls `setChildPickers` and sets the meridiem value back to the previous value.

source/TimePicker.js
```diff
@@ -585,14 +608,27 @@
...
+				oldIndex = event.old,
+				newIndex = this.$.meridiem.getValue(),
+				offset, start, end, newHour;
 
-			if (this.value) {
-				// value is 0 for am, 1 for pm
-				// reset the hour to < 12 and then add 12 if it's pm
-				hour = hour%12 + value*12;
-				this.updateHours(hour);
+			offset = hour - parseInt(meridiems[oldIndex]['start'], 10);
```
aarontam: Just wanted to confirm that the desired behavior is to preserve the offset from the start of the previous meridiem, when changing meridiems. Is this what occurs in other interfaces for this locale?

cc @ehoogerbeets

source/TimePicker.js
DavidUM: Yes you are right.
localeValue has a meaning when this.value is exist.
We do not concern about localeValue when this.value is null

source/TimePicker.js
DavidUM: When we have just two locale like AM/PM, hour and minute are not changed when we change meridiem betw'n AM/PM.
For example, AM 11:45 will become PM 11:45.
I though this behavior is general in other locale too.

Yes, we'd better get confrim from @ehoogerbeets

ehoogerbeets: For the Ethiopian time system, you cannot switch the meridiem arbitrarily without updating the time. The system goes like this: 

````
12:00-5:59 morning
6:00       noon
6:01-11:59 afternoon
12:00-5:59 evening
6:00-11:59 night
````

So if the UI is showing 7:00 in the afternoon, and you change the meridiem to evening, I would expect that the time gets updated to 12:00, which is the earliest time in the evening.

Comment (graynorton) [inline source/TimePicker.js]: @ehoogerbeets, are there any other time systems that behave this way, or is it only Ethiopian?

Is there any kind of generic indicator from iLib that the current locale requires this type of behavior (changing time when the meridiem changes)?

Basically I'm looking for the most general solution we can apply to the Moonstone control.

File: source/TimePicker.js [hunk 12]
```diff
@@ -609,17 +645,18 @@
 		/**
 		* @private
 		*/
-		updateHours: function (hour) {
+		updateValue: function (newHour, newMinute) {
 			var valueTime = this.value.getTime();
 
-			this.value.setHours(hour);
+			if (newHour !== undefined) { this.value.setHours(newHour); }
```
aarontam: I think we can change this to be `if (newHour != null) this.value.setHours(newHour);`. Similarly, we can do this for the next line as well.

source/TimePicker.js
```diff
@@ -628,35 +665,64 @@
...
 			if (this.value) {
-				var hour = this.value.getHours();
-				this.$.hour.setValue(hour);
-				this.$.minute.setValue(this.value.getMinutes());
-				if (this.meridiemEnable === true) {
-					this.$.meridiem.setValue(hour > 11 ? 1 : 0);
+				var values = this.calcPickerValues();
+				this.$.hour.set('value', values.hour);
+				this.$.minute.set('value', values.minute);
+				if (this.meridiemEnable === true) {
```
aarontam: I think this can be a simple truthy check i.e. `if (this.meridiemEnable) {`

DavidUM: Oh, I agree. Thanks.

source/TimePicker.js
```diff
@@ -470,10 +487,11 @@
 					if (this.meridiemEnable === true) {
 						this.createComponent(
 							{classes: 'moon-date-picker-wrap', components:[
-								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', value: valueHours > 12 ? 1 : 0, meridiems: this.meridiems || ['am','pm'], onChange: 'meridiemPickerChanged'},
+								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', meridiems: this.meridiems || [{name: 'AM', start: '00:00', end: '11:59'}, {name: 'PM', start: '12:00', end: '23:59'}], max: this.meridiems? this.meridiems.length - 1 : 1, onChange: 'meridiemPickerChanged'},
 								{name: 'meridiemLabel', content: this.meridiemText, classes: 'moon-date-picker-label moon-divider-text'}
 							]}
 						);
+						this.$.meridiem.value = this.getMeridiemIndex(values.hour, values.minute);
```
aarontam: I was thinking more like an `updateValue` method we expose on `MeridiemPicker`, that would accept an `hours` and a `minutes` parameter, and would encapsulate the logic found in `getMeridiemIndex`. I realize it's difficult to properly separate everything without doing a wholesale change, but I would like to reduce the amount of meridiem knowledge the `TimePicker` has, and keep that within `MeridiemPicker`. What do you think about this?

source/TimePicker.js
```diff
@@ -470,10 +487,11 @@
 					if (this.meridiemEnable === true) {
 						this.createComponent(
 							{classes: 'moon-date-picker-wrap', components:[
-								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', value: valueHours > 12 ? 1 : 0, meridiems: this.meridiems || ['am','pm'], onChange: 'meridiemPickerChanged'},
+								{kind:'moon.MeridiemPicker', name:'meridiem', classes:'moon-date-picker-field', meridiems: this.meridiems || [{name: 'AM', start: '00:00', end: '11:59'}, {name: 'PM', start: '12:00', end: '23:59'}], max: this.meridiems? this.meridiems.length - 1 : 1, onChange: 'meridiemPickerChanged'},
```
aarontam: So if `this.meridiems` is not specified, we set the `max` to 1 by default. Do you know how this interacts with the statement on L615: https://github.com/enyojs/moonstone/pull/2170/files#diff-0b78987cba1669c429231298cc6c83acR615

source/TimePicker.js
```diff
@@ -419,9 +414,31 @@
 					fmtParams.locale = this.locale;
 				}
 				var merFormatter = new ilib.DateFmt(fmtParams);
-				var am = ilib.Date.newInstance({hour:10});
-				var pm = ilib.Date.newInstance({hour:14});
-				this.meridiems = [merFormatter.format(am), merFormatter.format(pm)];
+				
+				this.meridiems = merFormatter.getMeridiemsRange(fmtParams);
```
aarontam: Do we need to set this as a property of `TimePicker`, or we can keep this set within just `MeridiemPicker`, so that we only have a single system of record. We are referencing `this.meridiems` on L606 (https://github.com/enyojs/moonstone/pull/2170/files#diff-0b78987cba1669c429231298cc6c83acR606), but we can simply retrieve the value from the event originator.

aarontam:
Hi @DavidUM, thanks for the updates! I made a few comments and also noticed that repeatedly changing the meridiem (via pressing the down button of the meridiem picker), causes the minute to increment, so I believe there is something off with the offset still. Let me know if you are unable to reproduce - for me it was pressing until I got to the ከሰዓት meridiem.

source/TimePicker.js
DavidUM: {{this.meridiems}} has default value.
Unless developer sets {{this.meridiems}} to undefined, it will has value.

source/TimePicker.js
DavidUM: Thank you. Let me update it from 2.6.0-dev.
Cause I'd like to make at least changes as I can to prevent regression issue.

source/TimePicker.js
ehoogerbeets: Yes, the Chinese meridiems are like the Ethiopian ones, should we choose to support them. There are other ones like that as well, but ilib doesn't support them yet.

Ilib has a call to retrieve all the meridiems and their start/end times: DateFmt.DateFmt.getMeridiemsRange() which returns an array of {name:string, start:string, end:string} objects which give the details of what the meridiems are called, when they start in "hours:minutes" format, and when they end, also in "hours:minutes". If you only key off of that, then you can support any meridiems style that ilib supports.

source/TimePicker.js
aarontam: Hi @DavidUM, thanks for the updates! I think we can remove this assignment, right?

source/TimePicker.js
```diff
@@ -53,14 +53,73 @@
...
+
+			var merFormatter = new ilib.DateFmt(fmtParams);	
+			this.meridiems = merFormatter.getMeridiemsRange(fmtParams);
+		},
+
+		/**
+		* @private
+		*/
+		create: function() {
+			this.initILib();
```
aarontam: I think for the sake of consistency in 2.5 (we've made `enyo-ilib` a dependency of Moonstone in 2.6, so this is not applicable for 2.6), we may want to guard this based on the existence of `ilib`, as we are doing here: https://github.com/enyojs/moonstone/blob/2.5-upkeep/source/DateTimePickerBase.js#L186. What do you think?

source/TimePicker.js
```diff
@@ -53,14 +53,73 @@
...
+		/**
+		* Set meridiem value based on given hour and minute
+		*
+		* @param  {Number} hour - hour between 0 to 23
+		* @param  {Number} minute - minute between 0 to 59
+		*
+		* @public
+		*/
+		setValueByTime: function (hour, minute) {
+			this.value = this.getMeridiemIndex(hour, minute);
```
aarontam: Should this call the setter instead, i.e. `this.set('value', this.getMeridiemIndex(hour, minute)`? Also, I'm wondering if we can fold `getMeridiemIndex` into this function as I don't think it needs to be a separate, public function - it is logic internal to `MeridiemPicker` that only it should know about.

source/TimePicker.js
```diff
@@ -628,35 +690,64 @@
...
-				var hour = this.value.getHours();
-				this.$.hour.setValue(hour);
-				this.$.minute.setValue(this.value.getMinutes());
-				if (this.meridiemEnable === true) {
-					this.$.meridiem.setValue(hour > 11 ? 1 : 0);
+				var values = this.calcPickerValues();
+				this.$.hour.set('value', values.hour);
+				this.$.minute.set('value', values.minute);
+				if (this.meridiemEnable) {					
+					this.$.meridiem.set('value', this.$.meridiem.getMeridiemIndex(values.hour, values.minute));
```
aarontam: Should this use the `setValueByTime` method instead?

source/TimePicker.js
DavidUM: I agree.
From 2.6.0-dev, we assume that we have ilib basically, but in 2.5-upkeep we'd better check it.
I'll add guard code.

source/TimePicker.js
DavidUM: We can combine `getMeridiemIndex` into this method.
But value assignment is designed because I do not want to notify it.

aarontam: Hm, so when does `valueChanged` get called?

DavidUM: Let me take a look again. Thanks @aarontam

source/TimePicker.js
```diff
@@ -53,14 +53,67 @@
...
+				start = parseInt(meridiems[i]['start'].substring(0,2) + meridiems[i]['start'].substring(3,5), 10);
+				end = parseInt(meridiems[i]['end'].substring(0,2) + meridiems[i]['end'].substring(3,5), 10);
+				time = hour * 100 + minute;
+
+				if ( start <= time && time <= end) {
+					if (this.value != null) {
+						this.set('value', i);
+					} else {
+						//at initial time, we don't want to trigger meridiemPickerChanged method
+						this.value = i;
```
aarontam: Instead of doing this, what do you think about using a call to silence/unsilence, similar to what is done in `moon.DatePicker`? https://github.com/enyojs/moonstone/blob/2.5-upkeep/source/DatePicker.js#L186.

DavidUM: I had a same idea with you. We can re-factor those controls.

And I also consider using silence in this part but I think using assignment is more cheaper.
If you mind to use assignment, let me fix it with using silence.
Thank you @aarontam

aarontam: I think `silence` is only marginally more expensive (it also consists of an assignment, and adds an increment operation), but, for example, to handle the case where value might be set to `null` later and we want to run `setValueByTime` to fire an `onChange` event, I think we should model the behavior in `DatePicker` so that the behavior is more robust.

DavidUM: I agree and will commit soon. Thank you Aaron.

source/TimePicker.js
```diff
@@ -585,14 +602,32 @@
...
-			if (this.value) {
-				// value is 0 for am, 1 for pm
-				// reset the hour to < 12 and then add 12 if it's pm
-				hour = hour%12 + value*12;
-				this.updateHours(hour);
+			if (meridiems.length == 2) {
+				this.updateValue(startHour + offset);
+			} else {
+				var oldMinute =  this.$.minute.get('value'),
+					endHour = parseInt(newMeridiem['end'], 10),
```
aarontam: Let's use dot notation (i.e. `newMeridiem.end`) where possible. This applies to the following two lines, as well.

source/TimePicker.js
```diff
@@ -53,14 +53,60 @@
...
+		* @param  {Number} hour - hour between 0 to 23
+		* @param  {Number} minute - minute between 0 to 59
+		*
+		* @public
+		*/
+		setValueByTime: function (hour, minute) {
+			var meridiems = this.meridiems,
+				start, end, time;
+			for (var i = 0; i < meridiems.length; i++) {
+				start = parseInt(meridiems[i]['start'].substring(0,2) + meridiems[i]['start'].substring(3,5), 10);
```
aarontam: Let's use dot notation for object properties here and on the following line.

source/TimePicker.js
```diff
@@ -75,13 +121,13 @@
 		* @private
 		*/
 		setupItem: function (inSender, inEvent) {
-			var index = inEvent.index % this.range;
-			this.$.item.setContent(this.meridiems[index]);
+			var index = inEvent.index % this.range || 0;
+			this.$.item.setContent(this.meridiems[index]['name']);
```
aarontam: Dot notation here as well.

aarontam:
Now that we're setting the `value` to be an `ilib.Date` object, that means `ilib` is a hard dependency of these controls; if we go down this path, we should remove all of the `ilib` existence checks and related conditionals.

source/DatePicker.js
```diff
@@ -262,7 +253,7 @@
 					second: valueSeconds,
 					millisecond: valueMilliseconds
 				});
-				this.setValue(this.localeValue.getJSDate());
+				this.setValue(localeValue);
```
aarontam: Is it still necessary to do this, or can we just update the existing `ilib.Date` object?

source/TimePicker.js
```diff
@@ -585,14 +602,32 @@
 		meridiemPickerChanged: function (sender, event) {
 			if(this.syncingPickers) return true;
 
-			var hour = this.$.hour.get('value'),
-				value = event.value;
+			var meridiems = event.originator.get('meridiems'),
+				oldMeridiem = meridiems[event.old],
+				newMeridiem = meridiems[this.$.meridiem.get('value')],
+				oldHour = this.$.hour.get('value'),
+				startHour = parseInt(newMeridiem['start'], 10),
```
aarontam: Dot notation here as well.

ryanjduffy:
ilib is a hard dependency for moonstone (for now at least)

aarontam:
Yep! This is for 2.5 :smile:

source/TimePicker.js
```diff
@@ -585,14 +602,32 @@
 		meridiemPickerChanged: function (sender, event) {
 			if(this.syncingPickers) return true;
 
-			var hour = this.$.hour.get('value'),
-				value = event.value;
+			var meridiems = event.originator.get('meridiems'),
+				oldMeridiem = meridiems[event.old],
```
aarontam: Rather than updating the properties of the `onChange` event fired by `IntegerPicker`, I wonder if we can just encapsulate this logic inside `MeridiemPicker` whenever its value changes (we would have access to the previous value), as we should know the most recent hour and minute (via `setValueByTime`), and come up with a clean way for `TimePicker` to retrieve the hour and minute in this change handler. It seems like the logic here is very meridiem-specific.

source/DatePicker.js
ehoogerbeets: ilib was not a hard dependency for moonstone before. In fact Gray was pretty adamant that it should not be. Is there some reason it has to become a hard dependency now?

source/TimePicker.js
DavidUM: I agree with your thought. Let me update it.

PR #2171, opened 2015-05-22
ENYO-1483-added a new property called 'state'

rajyavardhanp:
There is a requirement to add a new line to the tool tip. So it is
achieved using a new property 'state'

### DIFF
--- a/lib/Tooltip/Tooltip.js
+++ b/lib/Tooltip/Tooltip.js
@@ -156,7 +156,16 @@ module.exports = kind(
 		* @default null
 		* @public
 		*/
-		contentUpperCase: null
+		contentUpperCase: null,
+
+		/**
+		* The property is introduced to support multiline tooltip
+		* when state is declared, it will create a multiline popup.  
+		* @type {string}
+		* @default undefined
+		* @public
+		*/
+		state: undefined
 	},
 
 	/**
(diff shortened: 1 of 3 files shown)


dmikey:
Is there a reason we can't or would not call this property `multiline` ?

rajyavardhanp:
If the UI is considered, 'multiline' can be a suitable name. but 'multiline'  sounds like a boolean property and the name  doesn't represent the exact  functionality of the property. So I opt for 'state' over 'multilene' as per the requirement.
Thankyou.

lib/Tooltip/Tooltip.js
```diff
@@ -156,7 +156,16 @@ module.exports = kind(
...
+		contentUpperCase: null,
+
+		/**
+		* The property is introduced to support multiline tooltip
+		* when state is declared, it will create a multiline popup.  
+		* @type {string}
+		* @default undefined
+		* @public
+		*/
+		state: undefined
```
otisnate: I think ```null``` is better than ```state```

otisnate:
I understand ```state``` is for 2nd line of tooltip. But ```state``` seems like too normal word

lib/Tooltip/Tooltip.js
dmikey: Ok, thanks for explaining this, I think some of the confusion from the word state, is that it is really 2nd line content. And the word `state` just doesn't lend itself to being verbose that this is content. Seems like if we need to sent this content with a property, we should call it something like `subText`or something to indicate that it is both content AND moves to the next line.

Curious why this couldn't be accomplished using HTML spans inside the tooltip content?

rajyavardhanp: @otisnate Yes. It should be `null` . I will change it. 
@dmikey  It can be achieved, but tool-tip design is based on one line. So it takes more effort (should know the internals too). The position of the  tip of the tooltip and the height of the tooltip (block div) are highly dependent on each other in the current design. So that's the first hurdle i see for the app developer to use HTML spans. 
'subText'  sounds a suitable name. you are right about `state` doesn't give that feel of content.  But based on the app requirement, when I gave the name `state`, I was not expecting of any random string. The values of the property are some predefined limited no.of strings. 
However, we are looking for more use cases for this scenario, once its done I will change the name to a more appropriate one.
Thankyou.

ryanjduffy:
Closing in favor of #2184

PR #2172, opened 2015-05-22
Enyo 1285 derekanderson

dmikey:
issue. Marquee Text is ellipsis but doesn't roll when content has same length with div width

fix. when rendered, request the marquee to calculate it's distance.  Remove the ellipse formatting if the distance to move is 0.


DavidUM:
Interesting solution! :+1:

PR #2175, opened 2015-05-26
ENYO-1543-Customization on moon.Dialog

rajyavardhanp:
'bodyComponents' property is added, which allows the user can replace
the bodyText component in dialog with the components of one's choice.
Enyo-DCO-1.1-Signed-off-by: RajyavardhanP <rajyavardhan.p@lge.com>

### DIFF
--- a/lib/Dialog/Dialog.js
+++ b/lib/Dialog/Dialog.js
@@ -94,8 +94,8 @@ module.exports = kind(
 		uppercase: true,
 
 		/**
-		* When assigned with components, will replace the bodyText component with the controls
-		* given in the array. defaulted value should be revisited.
+		* When assigned with components before creation, will replace the {@link moon.Dialog#message}
+		* component with the controls given in the array.
 		*
 		* @type {Array}
 		* @default null
--- a/lib/Dialog/Dialog.js
+++ b/lib/Dialog/Dialog.js
@@ -94,6 +94,16 @@ module.exports = kind(
 		uppercase: true,
 
 		/**
+		* When assigned with components, will replace the bodyText component with the controls
+		* given in the array. defaulted value should be revisited.
+		*
+		* @type {Array}
+		* @default null
+		* @public
+		*/
+		bodyComponents: null,
+
+		/**
 		* @deprecated Replaced by [uppercase]{@link moon.Dialog#uppercase}.
 		*
 		* Formerly defaulted to `true`, now defaults to `null` and will only have
@@ -152,6 +162,12 @@ module.exports = kind(
 	* @private
 	*/
 	create: function () {
+		if(this.bodyComponents) {
+			this.tools = this.tools.slice(0,this.tools.length-2);
+			for(var comp in this.bodyComponents) {
+				this.tools.push(this.bodyComponents[comp]);
+			}
+		}
 		Popup.prototype.create.apply(this, arguments);
 
 		// FIXME: Backwards-compatibility for deprecated property - can be removed when
@@ -161,7 +177,7 @@ module.exports = kind(
 
 		this.titleChanged();
 		this.subTitleChanged();
-		this.messageChanged();
+		this.bodyComponents? '' : this.messageChanged();
 	},
 
 	/**
--- a/lib/Dialog/Dialog.js
+++ b/lib/Dialog/Dialog.js
@@ -154,7 +154,9 @@ module.exports = kind(
 			]}
 		]},
 		{kind: Divider, classes: 'moon-dialog-divider'},
-		{name: 'message', kind: BodyText, classes: 'moon-dialog-content'},
+		{name: 'bodyComps', components: [
+			{name: 'message', kind: BodyText, classes: 'moon-dialog-content'}
+		]},
 		{name: 'spotlightDummy', kind: Control, spotlight: false}
 	],
 
@@ -162,13 +164,13 @@ module.exports = kind(
 	* @private
 	*/
 	create: function () {
-		if(this.bodyComponents) {
-			this.tools = this.tools.slice(0,this.tools.length-2);
-			for(var comp in this.bodyComponents) {
-				this.tools.push(this.bodyComponents[comp]);
-			}
-		}
 		Popup.prototype.create.apply(this, arguments);
+		if(this.bodyComponents) {
+			this.$.message.destroy();
+			var owner = this.hasOwnProperty('bodyComponents') ? this.getInstanceOwner() : this;
+			this.$.bodyComps.createComponents(this.bodyComponents,{owner:owner});
+		}
+
 
 		// FIXME: Backwards-compatibility for deprecated property - can be removed when
 		// the titleUpperCase property is fully deprecated and removed. The legacy
@@ -177,7 +179,7 @@ module.exports = kind(
 
 		this.titleChanged();
 		this.subTitleChanged();
-		this.bodyComponents? '' : this.messageChanged();
+		this.messageChanged();
 	},
 
 	/**
@@ -217,6 +219,7 @@ module.exports = kind(
 	* @private
 	*/
 	messageChanged: function () {
-		this.$.message.setContent(this.message);
+		if(this.bodyComponents)
+			this.$.message.setContent(this.message);
 	}
 });
--- a/lib/Dialog/Dialog.js
+++ b/lib/Dialog/Dialog.js
@@ -219,7 +219,7 @@ module.exports = kind(
 	* @private
 	*/
 	messageChanged: function () {
-		if(this.bodyComponents)
+		if(!this.bodyComponents)
 			this.$.message.setContent(this.message);
 	}
 });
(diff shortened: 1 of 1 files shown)


KunmyonChoi:
Thanks for the PR.
After testing, I found that the controls given by bodyComponents doesn't handle event.
I just added below code.

KunmyonChoi:
```
			bodyComponents: [
				{kind: Button, content: 'Go get a coffee', ontap: 'hideDialog'},
				{kind: Button, content: 'No you should get a coffee', ontap: 'hideDialog'}
			]
```

lib/Dialog/Dialog.js
```diff
@@ -152,6 +162,12 @@ module.exports = kind(
 	* @private
 	*/
 	create: function () {
+		if(this.bodyComponents) {
+			this.tools = this.tools.slice(0,this.tools.length-2);
```
KunmyonChoi: You don't need to remove spotlightDummy from tools here.

lib/Dialog/Dialog.js
```diff
@@ -161,7 +177,7 @@ module.exports = kind(
 
 		this.titleChanged();
 		this.subTitleChanged();
-		this.messageChanged();
+		this.bodyComponents? '' : this.messageChanged();
```
KunmyonChoi: I think we can move condition to inside of messageChanged.

KunmyonChoi:
@rajyavardhanp , I suggest to reference moonstone panel.js headerComponents implementation.
https://github.com/enyojs/moonstone/blob/master/source/Panel.js#L326

lib/Dialog/Dialog.js
```diff
@@ -144,7 +154,9 @@ module.exports = kind(
 			]}
 		]},
 		{kind: Divider, classes: 'moon-dialog-divider'},
-		{name: 'message', kind: BodyText, classes: 'moon-dialog-content'},
+		{name: 'bodyComps', components: [
```
KunmyonChoi: Hi @rajyavardhanp ,
I think this will work without wrapping message with another div. We can just create bodyComponents on message not bodyComps.

rajyavardhanp: hi @KunmyonChoi . Yes but the problem would be `BodyText ` kind. It doesn't add any components but it will add classes and extra properties which will be applied on DOM. So I adopted this approach.

lib/Dialog/Dialog.js
```diff
@@ -153,6 +165,12 @@ module.exports = kind(
 	*/
 	create: function () {
 		Popup.prototype.create.apply(this, arguments);
+		if(this.bodyComponents) {
+			this.$.message.destroy();
```
aarontam: Hi @rajyavardhanp, rather than create something we could need to destroy here, can we extend `initComponents` and conditionally create our components there instead?

lib/Dialog/Dialog.js
```diff
@@ -201,6 +219,7 @@ module.exports = kind(
 	* @private
 	*/
 	messageChanged: function () {
-		this.$.message.setContent(this.message);
+		if(!this.bodyComponents)
```
ryanjduffy: Please adjust the `if` statement to have a space before the opening parenthesis and, in this case, either enclosing the body with `{}` or moving the body up to the same line as the `if` statement.

lib/Dialog/Dialog.js
```diff
@@ -153,6 +165,12 @@ module.exports = kind(
 	*/
 	create: function () {
 		Popup.prototype.create.apply(this, arguments);
+		if(this.bodyComponents) {
+			this.$.message.destroy();
+			var owner = this.hasOwnProperty('bodyComponents') ? this.getInstanceOwner() : this;
+			this.$.bodyComps.createComponents(this.bodyComponents,{owner:owner});
```
ryanjduffy: Please add a space before the opening `{` and after the colon.

```
this.$.bodyComps.createComponents(this.bodyComponents, {owner: owner});
```

lib/Dialog/Dialog.js
KunmyonChoi: Hi @aarontam , you mean like this code? https://github.com/enyojs/moonstone/blob/2.6.0-dev/lib/Button/Button.js#L166

lib/Dialog/Dialog.js
KunmyonChoi: Okay that make sense.

lib/Dialog/Dialog.js
aarontam: Yep something like what we're doing in `moon.Button`.

KunmyonChoi: Hi @rajyavardhanp , would you like to consider @aarontam 's suggestion?

rajyavardhanp: The basic idea is not to create unnecessary components. But creating the components part is not handled by the dialog component, it's being taken care by enyo Popup and we are just providing `tools` not `components`. My previous commit in the this PR has a different approach. 
@aarontam  Can you please check it once. !? I can proceed with extending initComponents also but it will change the design of Dialog and seems a bit overhead for the component.
Thankyou.

aarontam: Hm, if we removed the definition of the "message" control from our component declaration, would something like this work for the implementation of `initComponents`?

```javascript
initComponents: function () {
			this.inherited(arguments);
			if (this.bodyComponents) {
				var owner = this.hasOwnProperty('bodyComponents') ? this.getInstanceOwner() : this;
				this.$.bodyComps.createComponents(this.bodyComponents,{owner:owner});
			} else {
				this.$.bodyComps.createComponent({
					name: 'message',
					kind: BodyText,
					classes: 'moon-dialog-content'
				});
			}
		}
```

rajyavardhanp: hi @aarontam Yes, it works. I committed the new code as you suggested. Please review it.
Thanks

ryanjduffy:
I just made a similar change to Tooltip to support multi-line content. In that change (#2184), we're using the standard components property and adding them to the label. Here, we're checking for components and either rendering the label (message in this case) or the components. I think we need to be consistent (which might mean both PRs need to change)

@aarontam - thoughts on this one?

aarontam:
I hadn't taken a full step back when looking at this PR, I was more commenting on the specific use of `destroy` earlier. I think having multiple child components is more of an `Enyo` approach, so that would be my gut feeling - thanks for the perspective! What aspects of https://github.com/enyojs/moonstone/pull/2184 did you envision requiring a change?

ryanjduffy:
I guess I have both an API and implementation question. For the API, I don't think we should introduce another `components`-like property when the existing property isn't used. For the implementation, should we conditionally create either the text content div or the components (like this PR does) or always create the tex content div and assign it as the controlParent so the components are added to it (like the other PR).

aarontam:
Agreed on not introducing another `components`-like property. Going back and forth in my head about the cost of always having an additional DOM node in the hierarchy vs. the potentially less elegant solution of conditionally creating the controls (which could make other code less elegant as well, if we require a check for what "state" we're in vs. having a single state of the controls that handles both cases). For the case of `Tooltip`, it seems that a fair number of these can be created in a given app and it might make sense to only create the additional DOM hierarchy if necessary. Doesn't seem as salient for `Dialog` in terms of number of instances, but it wouldn't hurt either. Thoughts (specifically around creating more DOM vs. less general solution)?

ryanjduffy:
In these cases (both Dialog and Tooltip), the extra node is present in the exception case and not the normal case. That leads me to think the extra complexity to omit it and guard against its reference isn't warranted.

KunmyonChoi:
Hi @Djspaceg , do you think @ryanjduffy 's approach is the same as what you mentioned while you are in Korea? Automatically create child component instead when the content is not set as a string but a kind definition?

Djspaceg:
It does sound the same. Ideally we'd implement it at a deep enough level where many controls could take advantage of this switch-a-roo capability. We could drop a ton of busy-work tedious code, once it's implemented too.

aarontam:
Forgot to respond on GH. @ryanjduffy I had been thinking about the other PR incorrectly - the concept in that PR seems more desirable then. @Djspaceg that's interesting about making this more of a built-in framework feature - where did you envision this taking place, and do you have a sense of other places where we can take advantage of such a feature? If we want to do something like that, it seems like we should implement this first and have both Dialog and Tooltip be our first customers. :)

Djspaceg:
It's doable at just the moonstone level, but i'd really like to see it at the UiComponent level. Basically, anything that could have text content could also support multiple controls, just like regular HTML tags. There could be an api to only allow one of the two, I guess... We could immedietly use it on moon.Button for sure, where we already do this in a super in-elegant way; but also on moon.Spinner (simple implementation), and maybe even InputHeader.

graynorton:
Hmm...so what's the difference between what you're proposing, Blake, and what we already have in the relationship between 'content' and the 'components' block?

Djspaceg:
@graynorton, after tinkering with some samples because of your comment, I realized that what I said above was literally the current behavior... I was thinking about a way to have different automated logic for handling content vs components, but I need more time to plan out my idea.

Usually the case we see is where we want to accept either content or components into the client area, but we only want to wrap content in a marquee-frame, and leave components alone.

ryanjduffy:
Closing in favor of #2267 and enyojs/enyo#1220

PR #2177, opened 2015-05-26
ENYO-1174: Updated button press-state (bulge/expand)

Djspaceg:


Djspaceg:
Opening under a more appropriate ticket number.

PR #2178, opened 2015-05-26
ENYO-1679: Updated button press-state (bulge/expand)

Djspaceg:


lib/Button/Button.less
```diff
@@ -16,6 +27,14 @@
 	padding: 0 @moon-button-h-padding;
 	margin: 0 @moon-spotlight-outset;
 	color: @moon-button-text-color;
+	-webkit-transform: translateZ(0);
+	-webkit-transition: transform 0.2s ease-out;
+	transform: translateZ(0);
+	transition: transform 0.2s ease-out;
+	-webkit-animation-duration: 0.2s;
```
aarontam: Hi @Djspaceg, should we also add the non-prefixed versions of the various animation properties?

Djspaceg: We can't (we can, but it's not supported yet). keyframe animation isn't done yet. Everyone still uses prefixes. (This surprised me so I specifically looked into it before submitting this PR.)

aarontam: Gotcha! I had noticed we added the non-prefixed versions of the animation properties in other places in the framework, but if we're still on prefixes, no worries.

lib/Button/Button.less
```diff
@@ -16,6 +27,14 @@
 	padding: 0 @moon-button-h-padding;
 	margin: 0 @moon-spotlight-outset;
 	color: @moon-button-text-color;
+	-webkit-transform: translateZ(0);
+	-webkit-transition: transform 0.2s ease-out;
+	transform: translateZ(0);
```
ryanjduffy: Instead of explicitly adding the `translateZ(0)` here, you can use the `.moon-composite` class I added during the lite work. It has that as well as `will-change` so we can consolidate the layer promotion into a single place.

lib/Button/Button.less
```diff
@@ -1,4 +1,15 @@
 /* Button */
+
+@-webkit-keyframes moonButtonBulge {
+	0%, 100% { transform: translateZ(0); }
```
ryanjduffy: I didn't test on the TV tonight but on my Mac, the translateZ wasn't necessary here to maintain a compositing layer during and after the animation.

lib/Button/Button.less
Djspaceg: If we did, it must have been to support future browsers. Chrome/blink/webkit doesn't yet. Firefox may, now that I think about it... We should check.

lib/Button/Button.less
Djspaceg: What should go into the 0 and 100 percent keyframes? Just empty?

lib/Button/Button.less
Djspaceg: Good Idea. I'll change that.

lib/Button/Button.less
ryanjduffy: Not sure if you could leave it empty. Maybe `scale(1)` for clarity?

Djspaceg: It appears that that line can just be omitted entirely.

PR #2180, opened 2015-05-27
Enyo-1722 Set default value false to do not affect on moon.Slider's s…

DavidUM:
…ub-kind

Issue
-------
Current default value of {{enableJumpIncrement}} is true. So some controls which use moon.Slider will have jumpIncrement button even though they do not want it.

Cause
--------
Default value is true.

Fix
-----
Set its defalut value to false

### DIFF
--- a/lib/Slider/Slider.js
+++ b/lib/Slider/Slider.js
@@ -137,10 +137,10 @@ module.exports = kind(
 		* When `true`, jumpIncrement button will be displayed in both side of slider.
 		*
 		* @type {Boolean}
-		* @default true
+		* @default false
 		* @public
 		*/
-		enableJumpIncrement: true,
+		enableJumpIncrement: false,
 
 		/**
 		* Sliders will increase or decrease as much as this percentage value in either direction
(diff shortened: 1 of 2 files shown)


aarontam:
Hi @DavidUM, is this something we actually want to maintain as being enabled by default, so that developers do not have to explicitly set this property to `true` just to meet the updated design requirements? If this is not the case, or I'm misinterpreting the new requirements, then we should also remove this property setting from `moon.VideoTransportSlider`: https://github.com/enyojs/moonstone/blob/2.6.0-dev/lib/VideoTransportSlider/VideoTransportSlider.js#L122.

DavidUM:
Hi @aarontam 
This requirment came from SmartShare.
I think other sub kind from other apps should not fix their code due to this new property.
How about your thought?

If you agree with my thought, I'll update videoTransportSlider too.

aarontam:
Hi @DavidUM, based on the ticket, I didn't realize the requirement was for a specific app. Thanks for the clarification - if that's the case, let's go ahead and update `VideoTransportSlider` then. Thanks!

PR #2181, opened 2015-05-27
Enyo-1709 Make headerContainer as instance of defaultKind which is mo…

DavidUM:
…on.Item.

Issue
------
From 2.6.0-pre.5, headerContainer of ExpandableListItem is unspottable.

Cause
--------
Its kind is defaultKind. but when we convert it to modular structure, we denoted it as moon.Control

Fix
----
Remove kind assignment for headerContainer

### DIFF
--- a/lib/ExpandableListItem/ExpandableListItem.js
+++ b/lib/ExpandableListItem/ExpandableListItem.js
@@ -173,7 +173,7 @@ module.exports = kind(
 	components: [
 		// headerContainer required to avoid bad scrollWidth returned in RTL for certain text
 		// widths (webkit bug)
-		{name: 'headerContainer', kind: Control, classes: 'moon-expandable-list-item-header moon-expandable-picker-header moon-expandable-list-header', onSpotlightFocus: 'headerFocus', ontap: 'expandContract', components: [
+		{name: 'headerContainer', classes: 'moon-expandable-list-item-header moon-expandable-picker-header moon-expandable-list-header', onSpotlightFocus: 'headerFocus', ontap: 'expandContract', components: [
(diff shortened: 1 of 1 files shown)


lib/ExpandableListItem/ExpandableListItem.js
```diff
@@ -173,7 +173,7 @@ module.exports = kind(
 	components: [
 		// headerContainer required to avoid bad scrollWidth returned in RTL for certain text
 		// widths (webkit bug)
-		{name: 'headerContainer', kind: Control, classes: 'moon-expandable-list-item-header moon-expandable-picker-header moon-expandable-list-header', onSpotlightFocus: 'headerFocus', ontap: 'expandContract', components: [
+		{name: 'headerContainer', classes: 'moon-expandable-list-item-header moon-expandable-picker-header moon-expandable-list-header', onSpotlightFocus: 'headerFocus', ontap: 'expandContract', components: [
```
aarontam: Alternatively, I wonder if we can just explicitly declare `kind: Item` instead as it seems the intention of the `defaultKind` is to be the default kind of the child controls added to the client control. For example, if `ExpandablePicker` had been engineered a little differently and didn't completely overwrite the components of `ExpandableListItem` (maybe some future work to unify all of these variants), with its setting of the `defaultKind` to `CheckboxItem`, that would cause problems for the header here.

ryanjduffy: I agree with @aarontam here. Let's explicitly declare the kind.

aarontam:
Can we close this in favor of https://github.com/enyojs/moonstone/pull/2292, so we don't need to squash the commits in this PR? :smile:

PR #2182, opened 2015-05-27
ENYO-1543: Scroller doesn't display last line when using p tag on con…

KunmyonChoi:
…tent (Updated)

Issue:
- When I set <p> tag inside of scroller, Inner div (Scroll content) margin-Top push outer div (scroller client) down.
- This issue can be happens when user set margin top inside of moon.Scroller.
- When I set explicit width or height on inner control to explode, then native scroll bar is visible.

Fix:
- Set overflow: auto on container div
- Hide native scroller

### DIFF
--- a/css/Scroller.less
+++ b/css/Scroller.less
@@ -10,7 +10,12 @@
 	width: auto;
 	height: auto;
 	position: relative;
-	overflow: visible;
+	overflow: auto;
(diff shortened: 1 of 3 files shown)


ryanjduffy:
Superceded by #2185 and enyojs/enyo#1181

PR #2183, opened 2015-05-28
Fixes ENYO-1483

ryanjduffy:
## Issue
Tooltip only supports a single line of text and multiline support is desired.

## Fix
Rework CSS and positioning logic to support a `components` block of multiple children


ryanjduffy:
Wrong target ...

PR #2186, opened 2015-05-29
ENYO-1738 moon.ImageMultiRes is missing in Enyo 2.6

yeram:
Add moon.ImageMultiRes kind.


yeram:
Actually, less file is empty. I'm not sure we need this file or not.

yeram:
enyo.Image is covering this feature. We don't need this.

PR #2188, opened 2015-05-30
Enyo 1740 Refactoring css class for CheckboxItem

DavidUM:
Issue
-------
check mark layout of CheckboxItem needs to fix

Cause
--------
CheckboxItem uses some css class of moon.Item. But we do not require it from .js file

Fix
----
Refactor CheckboxItem.less to meet some feature of moon.Item


ryanjduffy:
Closing in favor of #2269

PR #2190, opened 2015-05-30
ENYO-1313: Separate out vendor-specific pseudo-selector rules.

aarontam:
### Issue
When we merged the changes from https://github.com/enyojs/moonstone/pull/1945, we missed the fact that there were mixed vendor-specific selectors for `InputHeader`'s styling, which will cause the browser (at least WebKit) to stop parsing the rule once an unrecognized selector is hit. This resulted in several previously-applied styles disappearing from `InputHeader`.

### Fix
This may have gone the way of over-generalization as this is primarily useful in the cases where we have vendor-specific selectors that need the same rules applied while wanting to keep things DRY, but I've added a mixin function that accepts a rule and an array of selectors, and in turn applies the rules to each of these selectors. This assumes that you are calling the mixin function (`explode-rule`) inside a selector (which I believe is necessary in terms of how LESS parsing works), and it will append the `&` to signify the selector is to be combined directly with the parent selector. Also, we may want to consider adding the full gamut of vendor prefixes, but for the sake of replicating the previous functionality, I declined to do this for the time being.

### DIFF
--- a/lib/Header/Header.less
+++ b/lib/Header/Header.less
@@ -125,9 +125,12 @@
 			border: 0px;
 			width: 100%;
 			box-sizing: border-box;
-			background-color: transparent;
 
-			> .moon-input {
+			&.moon-focused {
+				color: @moon-input-header-text-color;
+			}
+
+			.moon-input {
 				.moon-header-text;
 				margin: 0px;
 				padding-left: 1px;
@@ -135,42 +138,35 @@
 				display: inline-block;
 				box-sizing: border-box;
 				line-height: 1em;
-				color: @moon-input-header-text-color;
+				color: inherit;
 				width: 100%;
 				text-overflow: ellipsis;
+
+				.input-placeholder({
+					color: @moon-input-header-placeholder-color;
+					margin-top: 12px;
+					line-height: 1.25em;
+				});
 			}
 
 			&.spotlight {
-				.moon-input {
+				&:not(.moon-focused) .moon-input {
 					color: inherit;
+					background-color: @moon-spotlight-background-color;
 
 					.input-placeholder({
 						color: inherit;
 					});
 				}
 
-				> .moon-input {
-					background-color: @moon-spotlight-background-color;
-				}
-
-				&.moon-focused > .moon-input {
+				&.moon-focused .moon-input {
 					background: none;
-
-					.input-placeholder({
-						color: @moon-input-header-placeholder-color;
-					});
 				}
 			}
 		}
 
-		.moon-input.moon-header-title {
-			position: static;
-
-			.input-placeholder({
-				color: @moon-input-header-placeholder-color;
-				margin-top:12px;
-				line-height: 1.25em;
-			});
+		.moon-header-title-below {
+			margin-top: 0;
 		}
 	}
 }
--- a/lib/Header/Header.less
+++ b/lib/Header/Header.less
@@ -116,6 +116,63 @@
 	.moon-header-client-text {
 		line-height: @moon-button-small-height;
 	}
+
+	//InputHeader classes
+	&.moon-input-header {
+		.moon-input-header-input-decorator {
+			margin: -1px 0px 0px;
+			padding: 0px;
+			border: 0px;
+			width: 100%;
+			box-sizing: border-box;
+			background-color: transparent;
+
+			> .moon-input {
+				.moon-header-text;
+				margin: 0px;
+				padding-left: 1px;
+				padding-right: 1px;
+				display: inline-block;
+				box-sizing: border-box;
+				line-height: 1em;
+				color: @moon-input-header-text-color;
+				width: 100%;
+				text-overflow: ellipsis;
+			}
+
+			&.spotlight {
+				.moon-input {
+					color: inherit;
+
+					.input-placeholder({
+						color: inherit;
+					});
+				}
+
+				> .moon-input {
+					background-color: @moon-spotlight-background-color;
+				}
+
+				&.moon-focused > .moon-input {
+					background: none;
+
+					.input-placeholder({
+						color: @moon-input-header-placeholder-color;
+					});
+				}
+			}
+		}
+
+		.moon-input.moon-header-title {
+			position: static;
+
+			.input-placeholder({
+				color: @moon-input-header-placeholder-color;
+				margin-top:12px;
+				line-height: 1.25em;
+			});
+		}
+	}
 }
 
 // neutral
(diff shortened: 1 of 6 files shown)


ryanjduffy:
It's definitely a slick way to solve the problem. I can see the over-generalization angle. In run-time code, that's a more compelling argument for me because debugging what is then usually overly complex code is painful. In this case, the code is evaluated build-time so it is (probably) easier to track down issues. That said, we probably seldom look at the compiled CSS in the course of testing things so we would likely never come across any problems until they were bugs.

Another approach, wholly inspired by yours, would be to create a mixin for each of these vendor-specific selectors or rules which we could call as needed. It's a bit more verbose but more clear as a result. Again, because it's build-time code, the verbosity doesn't translate into run-time cost.

```less
.placeholder(@rule) {
    &::webkit-input-placeholder,
    &::moz-placeholder {
        @rule();
    }
}

.moon-input {
    .placeholder({
        line-height: 2;
    })
}
```

Thoughts?

aarontam:
Thanks for the feedabck! I realize it's just an example, but I think a generic mixin name such as `.placeholder` will have a high likelihood of conflicting with a class utilizing the same name (`explode-rule` could probably be named better too). Naming aside, I don't think we can place those vendor-specific selectors as part of the same selector group, so it would have to be something like this:

```
.moon-mixin-placeholder(@rule) {
    &::webkit-input-placeholder {
        @rule();
    }
    &::moz-placeholder {
        @rule();
    }
}

.moon-input {
    .placeholder({
        line-height: 2;
    })
}
```

What do you think of this? I can keep the `explode-rule` mixin in my back pocket. :wink:

ryanjduffy:
Agreed about the `.placeholder` name. I was just illustrating. I misunderstood the nature of the bug so you're right to put those as separate selector blocks. Silly that we have to but there you are :)

If you take this route, might consider a non-moonstone-specific naming convention. I think these could ultimately move to enyo core but we should wait to do that until we convert those CSS files to LESS. Maybe `mixin-input-placeholder` or `mixin-prefix-placeholder` or something along those lines...

aarontam:
I like it overall, mainly for the increased readability and semantics - I think we may be able to use something similar to the `explode-rule` mixin for other cases, after doing some thinking, but we can optimize that later. I also prefixed the other mixins with `mixin-` to prevent any potential collisions in the future.

lib/Header/Header.less
```diff
@@ -125,24 +125,57 @@
...
 	}
-	&.moon-medium-header .moon-header-title, {
-		line-height: @moon-non-latin-medium-header-line-height;
+
+	/* Special override for Japanese font (LG Display_JP) which, as of 07-Feb-2014
+	has a taller baseline. This corrects both Header and InputHeader. 2014-02-24
+	Please remove the line-height rule below if this font file is changed. */
+	&.enyo-locale-ja .moon-header-title,
+	&.enyo-locale-ja .moon-input-header .moon-input-header-input-decorator > .moon-input {
+		line-height: 1.25em;
```
Djspaceg: So the font rules have changed. Is this custom line-height still necessary, since now standard latin has a taller line-height?

aarontam: Not sure - I thought you wrote that comment. :smile: No objections to removing it based on what you said.

lib/Header/Header.less
```diff
@@ -162,77 +195,58 @@
...
-	&.enyo-locale-te, // Telugu
-	&.enyo-locale-kn, // Kannada
-	&.enyo-locale-ml, // Malayalam
-	&.enyo-locale-mr, // Marathi
-	&.enyo-locale-bn, // Bengali
-	&.enyo-locale-pa {// Panjabi
-		.moon-header .moon-header-title {
-			height: 159px;
+//InputHeader classes
+.moon-input-header {
```
Djspaceg: Lets tuck this whole block up into L#119 so it's logically a subset of `.moon-header`. It will then also fit with the rest of the code, and let the non-latin and RTL code be last in the file, like other controls.

aarontam: Good call (yougotitdude)

lib/Header/Header.less
Djspaceg: HAH! That's why I added the date, so I could tell when the code was hella old.

lib/Header/Header.less
```diff
@@ -162,77 +195,58 @@
...
+			line-height: 1em;
+			color: @moon-input-header-text-color;
+			width: 100%;
+			text-overflow: ellipsis;
+		}
+
+		&.spotlight {
+			& .moon-input {
+				.input-placeholder({
+					color: @moon-spotlight-text-color;
```
Djspaceg: Can we simplify these rules and specify the color at the lowest level like `.spotlight`, then use `inherit` if the rules don`t get picked up when they should? Seems like a really complicated set of cases that could be restructured to make it easier to follow the cases.

aarontam: Hmm, so the `input-placeholder` mixin creates the pseudo-classes and they need to be attached to `moon-input`, which is why it ended up being structured like this. I can take away some of the unnecessary use of "&" here, as I mechanically made the conversion.

aarontam:
I think this branch is ripe for some squashing!

lib/Header/Header.less
```diff
@@ -116,6 +116,62 @@
...
+				display: inline-block;
+				box-sizing: border-box;
+				line-height: 1em;
+				color: @moon-input-header-text-color;
+				width: 100%;
+				text-overflow: ellipsis;
+			}
+
+			&.spotlight {
+				.moon-input {
```
Djspaceg: I was meaning more along these lines:
```
&.spotlight {
    .moon-input {
        color: inherit;

        .input-placeholder({
            color: inherit;
        });
    }
    > .moon-input {
        background-color: inherit;
    }
    ...
}

Comment (aarontam) [inline lib/Header/Header.less]: Ah! Yes we can definitely do that.

Comment (Djspaceg) [inline lib/Header/Header.less]: You'll have to make sure that the pseudo-selectors inherit the _right_ color, but otherwise it should be dandy.

**Comment by otisnate** [comment]
If you are going to update ```VideoPlayer.less``` in ```moonstone```, I think you need to modify ```VideoPlayer.less``` in ```moonstone-extra```, too

**Comment by Djspaceg** [comment]
@otisnate It has these fixes already.

**Comment by ryanjduffy** [comment]
Should 4329461 be included in #1228 instead of this PR?

**Comment by aarontam** [comment]
Makes sense - updated!

File: lib/VideoPlayer/VideoPlayer.less [hunk 5]
```diff
@@ -45,32 +45,32 @@
 
 /* Fullscreen CSS */
 :-webkit-full-screen.moon-video-player {
-	.full-screen-video-player;
+	.full-screen-video-player();
```
otisnate: What is difference between ```.full-screen-video-player``` and ```.full-screen-video-player()``` ?

aarontam: This was actually not needed and superseded by https://github.com/enyojs/moonstone/pull/2306. It only appears this way because we split up the commits and moved one of the commits to the other PR.

Djspaceg: Functionally they're the same (as far as LESS cares). The trailing `()` are useful to visually differentiate a basic mixin from a mixin-function with no arguments.

PR #2191, opened 2015-06-01
ENYO-1687 : Apply accessibility on moonstone CheckboxItem

kbs12e:
1) Apply accessibilty on moonstone CheckboxItem.
2) Apply accessiblity on moonstone Icon because CheckboxItem's dom has aria-label which includes font hash-code content.
3) Add accessibility on moonstone Checkbox because CheckboxItem consists of Checkbox and content which have aria-label and tabIndex, so we should remove aria-label and tabIndex of children because CheckboxItem has already aria-label and tabIndex.

## Issue 
1) When CheckboxItem is focused, Only checkboxItem's content is read out. 
2) When font-based Icon is focused, Font hash-code content is read out. 
3) When CheckboxItem's child is focused in touch devices, this child component's aria-label is readout.

## Cause
1) The CheckboxItem consists of checkbox and text item. When spotlight is focused on CheckboxItem, the dom which is focused has aria-label including only the content of this text item. In this case, CheckboxItem should work like checkbox.
2) To support font-based Icon, Icon sets the font hash-code to dom content value. In this case, If accessiblity feature is true, added aria-label is made using Icon's content value and aria-label is read out.
3) Enyo Control adds tabIndex and aria-label to all components which have content, so Checkbox and content have aria-label and tabIndex.

## Fix
1) Add checkbox role and aria-checked attribute.
2) Remove font hash-code value of aria-label.
3) Remove aria-label and tabIndex from child component of CheckboxItem using accessibilityDisabled API 

### DIFF
--- a/lib/Checkbox/Checkbox.js
+++ b/lib/Checkbox/Checkbox.js
@@ -7,11 +7,13 @@ require('moonstone');
 
 var
 	kind = require('enyo/kind'),
+	options = require('enyo/options'),
 	ri = require('enyo/resolution'),
 	Checkbox = require('enyo/Checkbox');
 
 var
-	Icon = require('../Icon');
+	Icon = require('../Icon'),
+	CheckboxAccessibilitySupport = require('./CheckboxAccessibilitySupport');
 
 /**
 * {@link moon.Checkbox} is a box that, when tapped, shows or hides a checkmark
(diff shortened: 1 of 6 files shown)


kbs12e:
This PR has dependency on https://github.com/enyojs/spotlight/pull/176

seunghoh:
Child components of CheckboxItem (string and checkbox) don't have to have aria-label and tabindex since CheckboxItem has. Add accessibilityDisabled API to disabling after https://github.com/enyojs/enyo/pull/1188 merged

kbs12e:
There are a lot of changes according to dependency is changed, so many commit log is not neccessary. we will make PR again with latest version.

PR #2198, opened 2015-06-04
ENYO-1688 :[Accessibility] Apply accessibility on ToggleButton

kbs12e:
Apply accessibility on moonstone ToggleButton
However, ToggleButton's child component has tabIndex and aria-label, so we should remove this properties because ToggleButton already has button role and related aria attributes.
Thus, apply accessibility on moonstone Button because ToggleButton extends Button.

## Issue 
Should read out toggle status when the spotlight is focused to ToggleButton

## Cause
ToggleButton requires press-and-release cycle to change their value.
To support readout toggle-status, we should add aria-pressed attribute to ToggleButton. 

## Fix
Add aria-pressed attribute to ToggleButton
Remove tablndex and aria-label from Button using accessibilityDisabled API.

### DIFF
--- a/lib/Button/Button.js
+++ b/lib/Button/Button.js
@@ -8,13 +8,15 @@ require('moonstone');
 var
 	kind = require('enyo/kind'),
 	util = require('enyo/utils'),
+	options = require('enyo/options'),
 	Control = require('enyo/Control'),
 	Button = require('enyo/Button');
 
 var
 	Marquee = require('../Marquee'),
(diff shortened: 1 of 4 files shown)


kbs12e:
This PR has dependency on https://github.com/enyojs/spotlight/pull/176

kbs12e:
There are a lot of changes according to dependency is changed, so many commit log is not neccessary. we will make PR again with latest version.

PR #2199, opened 2015-06-04
ENYO1288 : Transparent button in image header and full image pattern(rank: 27)

suhlee:
## Issue

There is a requirement for transparent button in Dreadlocks.

##### For the requirement, I added 'translucent' property. Valid values are `'opaque'`, `'translucent'`, and `'transparent'`. Default value is 'opaque'

ENYO-DCO-1.1-Signed-Off-By: Suhyung Lee suhyung2.lee@lge.com

### DIFF
--- a/lib/Button/Button.js
+++ b/lib/Button/Button.js
@@ -100,7 +100,16 @@ module.exports = kind(
 		* @default null
 		* @public
 		*/
-		contentUpperCase: null
+		contentUpperCase: null,
+
+		/**
+		* The translucent of the button; valid alues are `'opaque'`, `'translucent'`, and `'transparent'`.
+		*
+		* @type {String}
+		* @default opaque
+		* @public
+		*/
+		translucent: 'opaque'
 	},
 
 	/**
@@ -167,6 +176,7 @@ module.exports = kind(
 		if (this.contentUpperCase !== null) this.uppercase = this.contentUpperCase;
 
 		this.contentChanged();
+		this.translucentChanged();
 		Button.prototype.initComponents.apply(this, arguments);
 	},
 
@@ -268,5 +278,13 @@ module.exports = kind(
 		if (!this.showing && this.hasClass('pressed')) {
 			this.undepress();
 		}
-	}
+	},
+
+	/**
+	* @private
+	*/
+	translucentChanged: function(inOldvalue){
+		this.removeClass('translucent-' + inOldvalue);
+		this.addClass('translucent-' + this.translucent);
+	}	
 });
--- a/lib/Button/Button.js
+++ b/lib/Button/Button.js
@@ -285,8 +285,8 @@ module.exports = kind(
 	* @private
 	*/
 	backgroundOpacityChanged: function (old) {
-		var opacity = this.get('backgroundOpacity');
-		this.removeClass(old);
+		var opacity = this.backgroundOpacity;
+		if (old) this.removeClass(old);
 		if (opacity == 'translucent' || opacity == 'transparent') {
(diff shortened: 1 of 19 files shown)


lib/Button/Button.js
```diff
@@ -268,5 +279,16 @@ module.exports = kind(
 		if (!this.showing && this.hasClass('pressed')) {
 			this.undepress();
 		}
+	},
+
+	/**
+	* @private
+	*/
+	backgroundOpacityChanged: function (old) {
+		var opacity = this.get('backgroundOpacity');
```
aarontam: Instead of doing this, can we use the second parameter passed to the change handler i.e. `backgroundOpacityChanged: function (was, is) {`

lib/IconButton/IconButton.js
```diff
@@ -195,5 +210,16 @@ module.exports = kind(
 		if (inEvent.originator === this) {
 			this.bubble('onRequestScrollIntoView');
 		}
+	},
+
+	/**
+	* @private
+	*/
+	backgroundOpacityChanged: function (old) {
```
aarontam: Thinking out loud and wondering if this can be abstracted, as the same logic will be shared by both `Button` and `IconButton`, to start.

aarontam:
I think we'll also want to update the `noBackground` property in `NewPagingControl`: https://github.com/enyojs/moonstone/blob/2.6.0-dev/lib/NewPagingControl/NewPagingControl.js#L67.

lib/Button/Button.js
ryanjduffy: Unfortunately, that will fail on the call during `create()` because it's called without args. At a minimum, we should change it to `this.backgroundOpacity` to avoid the unnecessary `get()` call. I'm all for using `get()` from outside a kind to protect the interface but from within I'd avoid it for the perf benefit.

On the following line, i would add an existence check to again skip an unnecessary function call.

```
if (old) this.removeClass(old);
```

lib/IconButton/IconButton.js
```diff
@@ -145,10 +158,12 @@ module.exports = kind(
 	},
 
 	/**
+	* Legacy support code, this shall be deprecated
+	*
 	* @private
```
ryanjduffy: add `@deprecated Replaced by backgroundOpacity`

lib/IconButton/IconButton.js
```diff
@@ -92,7 +92,17 @@ module.exports = kind(
 		* @default false
 		* @public
```
ryanjduffy: add `@deprecated Replaced by backgroundOpacity`

lib/IconButton/IconButton.js
aarontam: Too messy/complicated vs. the potential gain here.

suhlee: Aaron, could you please tell me  what is the potential gain?? 
I see your point about complicated and needed to be abstracted(less mixin)
The reason why I am asking is this backgroundOpacity may be applied to other controls like videoplayer control(play , pause button....)
Thanks

aarontam: Hi @suhlee, the main gain would be to avoid repetition and to generalize this behavior, so that we can both easily reuse the functionality, and make updates in a single location. After talking with @ryanjduffy, we decided against it for now, as the repeated functionality is fairly small and limited, and there would be a lot of complications abstracting the styling. For the controls in `VideoPlayer`, those are `IconButton` controls, so the functionality should already be present.

suhlee: I got it now! :)

PR #2200, opened 2015-06-04
Slightly different approach to translucency

Djspaceg:
I refactored what you had to match up with what I was thinking about.

This code does the following:
* Implements just translucent and transparent, opaque is the default so it doesn't have any special rules.
* Implements the background opacity of the pressed states of buttons, as well as allows spotlight to still look correct.
* Addresses the disabled state questions.
* Updates all existing controls to use the new technique.

### DIFF
--- a/lib/Button/Button.less
+++ b/lib/Button/Button.less
@@ -31,24 +31,20 @@
 		min-width: @moon-button-large-min-width;
 	}
 
-	//to handle transparent
-	.translucent(@translucent) {
-		color: @moon-translucent-font-color;
-		background-color: fade(@moon-button-background-color, @translucent);
+	// //to handle transparent
+	// .translucent(@translucent) {
+	// 	color: @moon-translucent-font-color;
+	// 	background-color: fade(@moon-button-background-color, @translucent);
+	// }
+
+	&.translucent:not(.disabled) {
+		background-color: fade(@moon-button-background-color, @moon-button-translucent-opacity);
 	}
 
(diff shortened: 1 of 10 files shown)


lib/IconButton/IconButton.js
```diff
@@ -155,10 +158,12 @@ module.exports = kind(
 	},
 
 	/**
+	* Legacy support code, this shall be deprecated
+	*
 	* @private
 	*/
 	noBackgroundChanged: function () {
-		this.addRemoveClass('no-background', this.noBackground);
+		this.set('backgroundOpacity', this.noBackground ? 'transparent' : 'opacity');
```
ryanjduffy: `'opaque'` instead of `'opacity'`?

Djspaceg: OOPS! Good catch. Thanks

lib/Button/Button.less
```diff
@@ -56,9 +52,18 @@
...
+		}
+		&.transparent:not(.disabled) {
+			background-color: transparent;
+		}
 	}
 
-	&.spotlight {
+	&.spotlight,
+	&.spotlight.translucent:not(.disabled),
+	&.spotlight.transparent:not(.disabled) {
```
suhlee: 65 +      &.spotlight.translucent:not(.disabled),
66 +      &.spotlight.transparent:not(.disabled) 
we need those selectors? those selectors appear again below.

PR #2201, opened 2015-06-04
Fixes ENYO-898

ryanjduffy:
## Issue
Animation logic was spread between Drawers and Drawer with reliance on
enyo/Drawer's implementation.

## Fix
Incorporate @KunmyonChoi's changes to use CSS transforms and refactor things a bit. Due to the need to coordinate the animation of a drawer and the client area of Drawers, we've tightly coupled the two controls and therefore made Drawer a private kind.


KunmyonChoi:
Hi @ryanjduffy , 
When I see apps code, they are hardly handling  onDrawerAnimationEnd event to catch timing to do something. So, I compared that part from previous behavior by adding below code on DrawerSample.

```
	handlers: {
		onDrawerAnimationEnd: 'end'
	},	
	end: function (sender, event) {
		console.log('animationEnd', sender.id, event.originator.id, event.originator.open);
	},
```

The result of comparison says, new drawer doesn't fire onDrawerAnimationEnd event on: 
- initial render: In previous code, each controlDrawer fires event on initial time 
- partial drawer open: In previous code, partialDrawer_controlDrawer fires event on open
- partial drawer close: In previous code, partialDrawer_controlDrawer fires event on close

And I noticed that the sample doesn't work without minor change like below code.
- this.$.partialDrawer.setOpen(true); => this.$.partialDrawer.set('open', true);

I checked these to see backward compatibility.

KunmyonChoi:
Unfortunately, I couldn't have a chance to measure FPS of this code on TV today. But, when we compare animation performance between animator and pure CSS. Pure CSS implementation shows better FPS on TV. And that's why Yeram made this PR.

KunmyonChoi:
Hi @ryanjduffy 
Do you have any reason to merge drawer code into drawers folder?
While I'm trying to apply new drawer, I noticed that I need to change my require statement.
Merging into drawers itself looks good, but it looks like requires unexpected app change.

KunmyonChoi:
Hi @ryanjduffy 
In case of audioPlayback control, the playback itself is a drawer and doesn't allow users to set controlDrawerComponents. So, I thought about little bit different approach on renderOnShow support on drawer directly. Please let me know what you think on this. 

https://github.com/enyojs/moonstone/pull/2206

ryanjduffy:
Re: moving Drawer, I think it's the right thing to do in this case as Drawer isn't designed to be used from within any other context than from within Drawers. As you discovered, the kind is still accessible as `require('moonstone/Drawers').Drawer` so unique use cases like yours is still possible. If we start to see more Drawer subkinds (either in the apps or the framework) then I'd be inclined to make it 'public' again.

lib/Drawers/Drawers.js
```diff
@@ -235,39 +242,69 @@ module.exports = kind(
 	* @private
 	*/
 	setupHandles: function () {
-		var controls, index,
-			handles = [];
+		var handle,
+			def = {kind: Item, owner: this},
+			tap = this.bindSafely(this.handleTapped);
+
+		this.drawers.forEach(function (drawer) {
```
aarontam: Similar question to before - think we should switch to a standard `for` loop here?

ryanjduffy:
@aarontam fixed the for loop and rebased on 2.6.0-dev

PR #2202, opened 2015-06-05
ExpandableListItem: Disabled ListItem is Enabled and wrong Caret Icon offset

anishlg:
### Issue:

The ListItem is enabled and unexpectedly expands and collapses even though it's disabled property is set as true.

In the ListItem header caret icon is placed down to the item header (wrong offset)

### Fix:

The ListItem header was a Control kind so it has changed to Item kind to achieve the disable feature. And the top-padding is reduced to arrange the caret icon offset.

### DIFF
--- a/lib/ExpandableListItem/ExpandableListItem.js
+++ b/lib/ExpandableListItem/ExpandableListItem.js
@@ -173,7 +173,7 @@ module.exports = kind(
 	components: [
 		// headerContainer required to avoid bad scrollWidth returned in RTL for certain text
 		// widths (webkit bug)
-		{name: 'headerContainer', kind: Control, classes: 'moon-expandable-list-item-header moon-expandable-picker-header moon-expandable-list-header', onSpotlightFocus: 'headerFocus', ontap: 'expandContract', components: [
+		{name: 'headerContainer', kind: Item, classes: 'moon-expandable-list-item-header moon-expandable-picker-header moon-expandable-list-header', onSpotlightFocus: 'headerFocus', ontap: 'expandContract', components: [
 			{name: 'header', kind: MarqueeText}
 		]},
 		{name: 'drawer', kind: ExpandableListDrawer, resizeContainer:false, classes: 'moon-expandable-list-item-client', components: [
--- a/lib/ExpandableListItem/ExpandableListItem.less
+++ b/lib/ExpandableListItem/ExpandableListItem.less
@@ -1,11 +1,12 @@
 /* ExpandableListItem Header*/
 .moon-expandable-list-item-header {
+  padding-right: @moon-spotlight-outset + 30px;
   margin-bottom: 0px;
   box-sizing: border-box;
   max-width: 100%;
 }
 .moon-expandable-list-header.moon-expandable-picker-header:after {
-  top: @moon-spotlight-outset + 3;
+  top: @moon-spotlight-outset;
 }
 
 /* Client Items */


lib/ExpandableListItem/ExpandableListItem.js
```diff
@@ -173,7 +173,7 @@ module.exports = kind(
 	components: [
 		// headerContainer required to avoid bad scrollWidth returned in RTL for certain text
 		// widths (webkit bug)
-		{name: 'headerContainer', kind: Control, classes: 'moon-expandable-list-item-header moon-expandable-picker-header moon-expandable-list-header', onSpotlightFocus: 'headerFocus', ontap: 'expandContract', components: [
+		{name: 'headerContainer', kind: Item, classes: 'moon-expandable-list-item-header moon-expandable-picker-header moon-expandable-list-header', onSpotlightFocus: 'headerFocus', ontap: 'expandContract', components: [
```
DavidUM: Good approach!

DavidUM:
Looks good to me.
But expandableListItem does not have {{moon-expandable-picker-header}}
So it also has same issue with https://jira2.lgsvl.com/browse/ENYO-1740

We can make another jira ticket to fix it.

lib/ExpandableListItem/ExpandableListItem.less
```diff
@@ -1,11 +1,12 @@
 /* ExpandableListItem Header*/
 .moon-expandable-list-item-header {
+  padding-right: @moon-spotlight-outset + 30px;
```
KunmyonChoi: In 2.5-upkeep, this padding is comes from .moon-expandable-picker-header and this css selector is compiled after .moon-item. But, in 2.6.0-dev, the .moon-item selector comes after .moon-expandable-picker-header and overrides padding as padding: 12px;.
Hi @aarontam, @Djspaceg , how do you think on this? I think this kind of issue can be found when css sequence is reversed by the require sequence.

lib/ExpandableListItem/ExpandableListItem.less
```diff
@@ -1,11 +1,12 @@
...
   margin-bottom: 0px;
   box-sizing: border-box;
   max-width: 100%;
 }
 .moon-expandable-list-header.moon-expandable-picker-header:after {
-  top: @moon-spotlight-outset + 3;
```
KunmyonChoi: I don't think we need this 3 here.

lib/ExpandableListItem/ExpandableListItem.less
graynorton: I haven't looked in detail at this specific case, but in general, we need to move away from CSS rules that are order-dependent. It was actually never a good practice, even before modules, as it would have been far too easy to accidentally break functionality by changing the order of CSS includes. The fix, in the general case, is to write more specific selectors for more specific controls (e.g., moonstone/Item is a more "general" control than moonstone/ExpandablePicker and its header), so that styling provided by the more general control can never override styling provided by the more specific control.

Djspaceg: @graynorton Is correct; it certainly needs to be a more specific rule for this type of override. On a more general note, I've been looking at this control's CSS quite a bit lately (for line height stuff) and I'm thoroughly convinced ExpandablePicker's CSS needs to be completely reworked. Its design and construction doesn't take advantage of LESS structuring and easily gets clobbered in cases like this, and has disappointingly long class names.

lib/ExpandableListItem/ExpandableListItem.less
DavidUM: I thought that but if you compare it with 2.5-upkeep, you may agree this.

lib/ExpandableListItem/ExpandableListItem.less
DavidUM: @Djspaceg Yes, some css related with expandablePicker need to rework.
Some controls like ExpandableListItem, Accordion have css dependency issue.

lishammel:
@anishlg  or @DavidUM  - it looks like the Jira associated with this PR has been closed with another fix. Can we close this out? Thanks.

lishammel:
closing out as the jira associated with this PR has been closed with another fix.

PR #2203, opened 2015-06-05
ENYO-1183: Music Player & Sample

KunmyonChoi:
- Re-implement moon.AudioPlayback and moon.PlaylistSupport
- AudioPlayback play model
- PlaylistSupport feed model from collection and navigate through collection when play prev/next
- Providing toggleRepeat and toggleShuffle


KunmyonChoi:
Sample for music player is here. https://github.com/enyojs/enyo-strawman/pull/29

lib/PlaylistSupport/PlaylistSupport.js
```diff
@@ -0,0 +1,370 @@
...
+	*/
+	removeAudio: function (model) {
+		var c = this.get('collection');
+		c && c.remove(model);
+	},
+
+	/**
+	* @private
+	*/
+	_removeAudio: function (sender, event) {
```
Djspaceg: On each of these functions where we refer to `event`, `event` should be changed to `ev` instead, because "event" is a reserlved global variable in Internet Explorer (and when we want it to work later, we won't have to change it.)

lib/AudioPlayback/AudioPlayback.js
```diff
@@ -41,326 +35,351 @@ var
...
 /**
-* @namespace moon
-* @class moon.AudioListItem
-* @extends enyo.Control
+* `moon.AudioPlayback` is meant to be used with {@link moon.Drawers}.
+* This extends a {@link moon.Drawer} by adding an audio playback control
+* and playlist for the imported audio.
+*
+* ```
+* {kind:'moon.Drawers', drawers:[
```
Djspaceg: Docs still refer to string kind instead of constructor.

lib/AudioPlayback/AudioPlayback.js
```diff
@@ -41,326 +35,351 @@ var
...
+* 	{
+* 		kind: 'moon.AudioPlayback',
```
Djspaceg: Here too, docs still refer to string kind instead of constructor.

lib/AudioPlayback/AudioPlayback.js
```diff
@@ -41,326 +35,351 @@ var
...
 
 	/**
 	* @private
 	*/
-	controlDrawerComponents:[],
+	enterTapArea: function (sender, event) {
+		if (event.originator.name == 'tapArea') {
+			this.startPreview();
+			if (!this.$.slider.disabled && this.hideButtonsOnSlider) {
+				this.$.playbackControls.setShowing(false);
```
Djspaceg: We should use the newer `this.$.playbackControls.set('showing', false);` in these instead. There's several of these, I won't point out each of them.

lib/AudioPlayback/AudioPlayback.js
```diff
@@ -1,26 +1,20 @@
...
-
 var
 	kind = require('enyo/kind'),
+	utils = require('enyo/utils'),
 	EnyoAudio = require('enyo/Audio'),
-	Img = require('enyo/Image'),
-	Signals = require('enyo/Signals');
+	Signals = require('enyo/Signals'),
+	EventEmitter = require('enyo/EventEmitter'),
+	Model = require('enyo/Model');
```
Djspaceg: `Model` is created but not needed.

lib/AudioPlayback/AudioPlayback.js
```diff
@@ -369,99 +388,144 @@ module.exports = kind(
...
+		var totalTime = this.getDuration(),
+			currentTime = this.getCurrentTime(),
+			bufferTime = this.getBufferedTime(),
+			playheadPos, bufferheadPos;
+
+		totalTime = isNaN(totalTime) ? 0 : totalTime;
+		currentTime = isNaN(currentTime) ? 0 : currentTime;
+
+		playheadPos = (totalTime == 0) ? 0 : (currentTime * 100) / totalTime;
+		bufferheadPos = (bufferTime == 0) ? 0 : (bufferTime * 100) / totalTime;
```
Djspaceg: These two would be better as falsey statements:
```
bufferheadPos = (!bufferTime) ? 0 : (bufferTime * 100) / totalTime;
```

lib/AudioPlayback/AudioPlayback.js
```diff
@@ -369,99 +388,144 @@ module.exports = kind(
...
+	getBufferedTime: function() {
+		var bufferData = this.$.media.getBuffered(),
+			numberOfBuffers = bufferData.length,
+			highestBufferPoint = 0,
+			duration = this.getDuration() || 0,
+			endPoint = 0,
+			i
+		;
+
+		if (duration === 0 || isNaN(duration)) {
```
Djspaceg: This could be falsey too:
```
if (!duration || isNaN(duration)) {
```

lib/AudioPlayback/AudioPlayback.js
```diff
@@ -471,11 +535,10 @@ module.exports = kind(
...
 	play: function () {
-		this.$.audio.play();
-		if (this.playheadJob === null) {
-			this.playheadJob = setInterval(this.bindSafely('updatePlayhead'), 500);
-		}
-		this.$.btnPlay.applyStyle('background-image', 'url(assets/icon-pause-btn.png)');
+		if (!this.$.media.get('src')) return;
+		this.$.media.play();
+		this.startPlayheadJob();
+		this.$.btnPlay.setIcon('pause');
```
Djspaceg:
```
this.$.btnPlay.set('icon', 'pause');
```

PR #2206, opened 2015-06-08
ENYO-1183: Support renderOnShow on drawer.

KunmyonChoi:
- When drawer change it's showing status, it is checking it's node and measure controlDrawer height before open.


ryanjduffy:
Good catch! I've added a slightly different approach (b396f3a) to my original PR. If `moonstone/Drawer` were truly usable outside of `moonstone/Drawers`, we'd need a solution that was self-contained. Since that isn't the case, we can simplify a bit and call `show()` in Drawers before `open` or `controlsOpen` is changed so the positioning logic is unchanged.

PR #2210, opened 2015-06-09
ENYO-1382 & ENYO-1383: Changes font size for ExpandableListItem and LabeledTextItem accordin…

JayCanuck:
…g the new guidelines

ENYO-1382: Expandable List Item font size needs to be 27px
ENYO-1383: Labeled Text Item: Label font size is smaller than text below

### DIFF
--- a/lib/ExpandablePicker/ExpandablePicker.less
+++ b/lib/ExpandablePicker/ExpandablePicker.less
@@ -38,7 +38,6 @@
 
 /* Current Value */
 .moon-expandable-picker-current-value {
-	.moon-body-text;
 	.moon-item-body-text;
 	color: @moon-expandable-picker-text-color;
 	margin: 0px;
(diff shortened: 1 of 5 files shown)


lib/ExpandablePicker/ExpandablePicker.less
```diff
@@ -38,7 +38,7 @@
 
 /* Current Value */
 .moon-expandable-picker-current-value {
-	.moon-body-text;
+	.moon-item-body-text;
```
aarontam: With this change, for 1080p, the `line-height` will change from `33px` to `36px` in a non-latin locale, and will change from `39px` to `39.5999984741211px` in other locales. I wanted to verify that this was intentional, thanks!

JayCanuck: Updated to include line-height

aarontam: Thanks! Will we need to add a rule like this for `moon-item-body-text` instead? https://github.com/enyojs/moonstone/blob/2.6.0-dev/css/moonstone-text.less#L179

The line-height is now `39px` in both non-latin and other locales.

Djspaceg:
I think this set of changes is too complicated/elaborate for addressing these issues. I'm gonna take a look at re-evaluating both of these controls and do another PR. @JayCanuck are you ok with closing this PR?

PR #2211, opened 2015-06-09
ENYO-1733: Changed visibility of border and height in settings app to…

tejaswineesohoni:
… control CSS instead of overriding.

### DIFF
--- a/lib/LabeledTextItem/LabeledTextItem.less
+++ b/lib/LabeledTextItem/LabeledTextItem.less
@@ -1,18 +1,14 @@
 /* Labeled Text Item */
 .moon-labeledtextitem {
 	min-width: 336px;
-	height: 192px;
 	overflow: hidden;
-	margin: 0px;
 }
 
 /* Label */
 .moon-labeledtextitem .label {
 	.moon-sub-header-text;
 	color: @moon-label-text-color;
-	border-top: 3px solid @moon-label-border-color;
-	margin: 0px 0px 3px 0px;
-	padding: 6px 0px 0px 0px;
+	margin-bottom: 6px;
 }
 .enyo-locale-non-latin .moon-labeledtextitem .label {
 	.enyo-locale-non-latin .moon-sub-header-text;
@@ -30,8 +26,6 @@
 	color: @moon-labeled-item-text-color;
 	line-height: @moon-body-line-height;
 	text-transform: none;
-	margin: 0px;
-	padding: 0px;
 }
 .enyo-locale-non-latin .moon-labeledtextitem .text {
 	.enyo-locale-non-latin .moon-body-text;


lib/LabeledTextItem/LabeledTextItem.less
```diff
@@ -10,7 +10,7 @@
 .moon-labeledtextitem .label {
 	.moon-sub-header-text;
 	color: @moon-label-text-color;
-	border-top: 3px solid @moon-label-border-color;
+	border-top: 0px solid @moon-label-border-color;
```
aarontam: Hi @tejaswineesohoni, the change looks and behaves just fine overall - I think we can probably remove the entire declaration from the rule by taking out this line completely. Thanks!

lib/LabeledTextItem/LabeledTextItem.less
```diff
@@ -10,7 +10,6 @@
 .moon-labeledtextitem .label {
 	.moon-sub-header-text;
 	color: @moon-label-text-color;
-	border-top: 3px solid @moon-label-border-color;
 	margin: 0px 0px 3px 0px;
 	padding: 6px 0px 0px 0px;
```
Djspaceg: We can safely drop `margin`, and `padding`, adding only `margin-bottom: 6px;` in their place.
* `.label`'s padding just pushes the title farther from the top, so it can be dropped.
* Increasing `.label`'s `margin-bottom` makes its text about the same distance from the top as it is from the value text, which more closely matches ExpandablePicker's style.

lib/LabeledTextItem/LabeledTextItem.less
```diff
@@ -1,7 +1,7 @@
 /* Labeled Text Item */
 .moon-labeledtextitem {
 	min-width: 336px;
-	height: 192px;
+	height: auto;
```
Djspaceg: Lets remove the lines for `height` and `margin: 0px;`.
* `height: auto` isn't needed (like @ryanjduffy said in Jira) since height is already inherently "auto". One less rule to parse, yay!
* margin has no effect, as it's naturally got no margin. Don't need a rule to reiterate it.

lib/LabeledTextItem/LabeledTextItem.less
```diff
@@ -1,18 +1,14 @@
...
 }
 
 /* Label */
 .moon-labeledtextitem .label {
 	.moon-sub-header-text;
 	color: @moon-label-text-color;
-	border-top: 3px solid @moon-label-border-color;
-	margin: 0px 0px 3px 0px;
-	padding: 6px 0px 0px 0px;
+	margin: 0px 0px 6px 0px;
```
Djspaceg: This line can just specify: `margin-bottom: 6px`. We don't need to explicitly assign 0px to top, right, or left. Thanks!

PR #2212, opened 2015-06-09
ENYO-1496: AppInfo-checking components in moon.History

JayCanuck:
Checks the appinfo.json for the disableBackHistoryAPI flag.

### DIFF
--- a/lib/History/History.js
+++ b/lib/History/History.js
@@ -4,6 +4,8 @@ var
 	kind = require('enyo/kind'),
 	ready = require('enyo/ready'),
 	dispatcher = require('enyo/dispatcher'),
+	platform = require('enyo/platform'),
+	Ajax = require('enyo/Ajax'),
 	Signals = require('enyo/Signals');
 
 /**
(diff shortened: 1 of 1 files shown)


lib/History/History.js
```diff
@@ -100,6 +74,14 @@ var History = module.exports = kind.singleton({
 		this._popStateHandler = this.bindSafely(this.popStateHandler);
 
 		this.enableBackHistoryAPIChanged();
+
+		this._appInfoReq = new Ajax({url:"appinfo.json"});
```
ryanjduffy: Let's guard this with a platform check so it only runs on webOS.

lib/History/History.js
```diff
@@ -100,6 +74,14 @@ var History = module.exports = kind.singleton({
...
+		this._appInfoReq.response(function(inSender, inResponse) {
+			if(inResponse !== undefined) {
```
ryanjduffy: Minor code style items - add a space between `function` and `if` and the opening paren. No need to prefix arguments with `in`.

lib/History/History.js
```diff
@@ -100,6 +74,14 @@ var History = module.exports = kind.singleton({
...
+				this.set("enableBackHistoryAPI", !inResponse.disableBackHistoryAPI);
+			}
+		}, this);
+		this._appInfoReq.go({});
```
ryanjduffy: No need for the empty object `{}` here. Beyond the unnecessary memory allocation, it also forces enyo/Ajax to do more work.

PR #2222, opened 2015-06-12
DRD-5039-artdahm Enyo2Sampler: layout of Basic Inputs differs from ba…

dragongears:
…se and browser version


Djspaceg:
Do we need to wrap each set of input controls in a container?

dragongears:
Oops. This should be a pull on 2.6.0.

PR #2227, opened 2015-06-16
ENYO-1687 : apply accessibility on moonstone checkboxItem

kbs12e:
Apply accessibility on moonstone CheckboxItem. 

## Issue

When CheckboxItem is focused, Only child checkbox is readout.

## Cause

The CheckboxItem consists of checkbox and text item. When spotlight is focused on CheckboxItem, only child checkbox component is readout because CheckboxItem does not have checkbox role and aria-checked attribute.

Checkbox and Text Item are child component of CheckboxItem, so they do not need tabIndex and accessibility-related code and If CheckboxItem has accessibilityDisabled:true, their accessibilityDisabled value also should be true To implement this work, we added accessibilityDisabled attribute to each component source code.

In CheckboxItem, **accessibilityDisabled:true** is added to checkbox and text control regardless of accessibilityDisabled of CheckboxItem.
In Checkbox, **accessibilityDisabled:true** is added to checkbox Icon regardless of accessibilityDisabled of Checkbox.
On the other hand, In Icon, **accessibilityDisabled:this.accessibilityDisabled** is added to child component, which is created dynamically when Icon is small. Because small Icon should have tabIndex when Icon used alone and accessibilityDisabled value is false
 
## Fix
Add checkbox role and aria-checked attribute to CheckboxItem.
Set accessibilityDisabled of child components to true.

### DIFF
--- a/lib/Checkbox/Checkbox.js
+++ b/lib/Checkbox/Checkbox.js
@@ -101,7 +101,7 @@ module.exports = kind(
 	* @private
 	*/
 	components: [
-		{name: 'checkboxIcon', kind: Icon, icon: 'check'}
+		{name: 'checkboxIcon', kind: Icon, accessibilityDisabled: true, icon: 'check'}
 	],
 
 	/**
--- a/lib/CheckboxItem/CheckboxItemAccessibilitySupport.js
+++ b/lib/CheckboxItem/CheckboxItemAccessibilitySupport.js
@@ -1,0 +1,38 @@
+/**
+* Provides a mixin to add accessibility support to
+* {@link module:moonstone/CheckboxItem~CheckboxItem}
+*
+* @module moonstone/CheckboxItem/CheckboxItemAccessibilitySupport
+* @private
+*/
+
+var
+	kind = require('enyo/kind');
+
+/**
+* @name CheckboxItemAccessibilitySupport
+* @mixin
+*/
+module.exports = {
+
+	/**
+	* @private
+	*/
+	observers: [
+		{method: 'updateAccessibilityAttributes', path: ['checked']}
+	],
+
+	/**
+	* @private
+	*/
+	updateAccessibilityAttributes: kind.inherit(function (sup) {
+		return function () {
+			var enabled = !this.accessibilityDisabled;
+			sup.apply(this, arguments);
+			this.setAttribute('role', enabled ? 'checkbox' : null);
+			this.setAttribute('tabindex', enabled ? 0 : null);
+			this.setAttribute('aria-checked', enabled ? String(this.checked) : null);
+			this.setAttribute('aria-labelledby', enabled ? this.$.client.getId() : null);
+		};
+	})
+};
--- a/lib/Icon/Icon.js
+++ b/lib/Icon/Icon.js
@@ -317,7 +317,7 @@ module.exports = kind(
 	*/
 	smallChanged: function () {
 		if (this.small) {
-			var ta = this.createComponent({name: 'tapArea', classes: 'small-icon-tap-area', allowHtml: this.allowHtml, isChrome: true});
+			var ta = this.createComponent({name: 'tapArea', classes: 'small-icon-tap-area', accessibilityDisabled: this.accessibilityDisabled, allowHtml: this.allowHtml, isChrome: true});
 
 			if (this.generated) {
 				ta.render();
--- a/lib/CheckboxItem/CheckboxItem.js
+++ b/lib/CheckboxItem/CheckboxItem.js
@@ -7,6 +7,7 @@ require('moonstone');
 
 var
 	kind = require('enyo/kind'),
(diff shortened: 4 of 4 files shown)


kbs12e:
This PR has dependency on https://github.com/enyojs/enyo/pull/1207

lib/Icon/Icon.js
```diff
@@ -59,7 +60,8 @@ var icons = {
 	gear              : '&#983083;',  // \0F002B
 	plug              : '&#983084;',  // \0F002C
 	lock              : '&#983085;'   // \0F002D
-};
+},
+	IconAccessibilitySupport = require('./IconAccessibilitySupport');
```
seunghoh: Move this line before the var icons for better readability

lib/Icon/IconAccessibilitySupport.js
```diff
@@ -0,0 +1,30 @@
...
+* @name IconAccessibilityMixin
+* @mixin
+*/
+module.exports = {
+
+	/**
+	* @private
+	*/
+	observers: [
+		{method: 'updateAccessibilityAttributes', path: ['icon']}
```
ryanjduffy: You need to observe `small` rather than `icon` as the former drives the creation of the tapArea control

lib/Icon/IconAccessibilitySupport.js
```diff
@@ -0,0 +1,30 @@
...
+	/**
+	* @private
+	*/
+	updateAccessibilityAttributes: kind.inherit(function (sup) {
+		return function (props) {
+			sup.apply(this, arguments);
+			// this.$.tapArea component is dynamically created using createComponents in smallChanged(), and call iconChanged(), so
+			// should check whether or not it exists.
+			if (this.$.tapArea) {
+				this.$.tapArea.set('accessibilityDisabled', this.accessibilityDisabled);
```
ryanjduffy: Do you mean to propagate accessibilityDisabled or to always disable accessibility for the tapArea?

lib/Checkbox/Checkbox.js
```diff
@@ -103,7 +103,7 @@ module.exports = kind(
 	* @private
 	*/
 	components: [
-		{name: 'checkboxIcon', kind: Icon, icon: 'check'}
+		{name: 'checkboxIcon', kind: Icon, accessibilityDisabled:true, icon: 'check'}
```
seunghoh: add space after colon -> accessibilityDisabled: true

lib/Icon/Icon.js
```diff
@@ -319,7 +319,7 @@ module.exports = kind(
 	*/
 	smallChanged: function () {
 		if (this.small) {
-			var ta = this.createComponent({name: 'tapArea', classes: 'small-icon-tap-area', allowHtml: this.allowHtml, isChrome: true});
+			var ta = this.createComponent({name: 'tapArea', classes: 'small-icon-tap-area', accessibilityDisabled:this.accessibilityDisabled, allowHtml: this.allowHtml, isChrome: true});
```
seunghoh: add space after colon -> accessibilityDisabled: this.accessibilityDisabled

lib/CheckboxItem/CheckboxItemAccessibilitySupport.js
```diff
@@ -0,0 +1,30 @@
...
+	* @private
+	*/
+	updateAccessibilityAttributes: kind.inherit(function (sup) {
+		return function (props) {
+			var enabled = !this.accessibilityDisabled;
+			sup.apply(this, arguments);
+			this.setAttribute('role', enabled ? 'checkbox' : null);
+			this.setAttribute('tabindex', enabled ? 0 : null);
+			this.setAttribute('aria-checked', enabled ? String(this.checked) : null);
+			this.setAttribute('aria-labelledby', enabled ? this.getId() : null);
```
ryanjduffy: Should this use the client's id rather than the control's id? Otherwise, it's declaring that it's labeling itself but it doesn't have any direct content.

```
this.setAttribute('aria-labelledby', enabled ? this.$.client.getId() : null);
```

kbs12e: We use aria-labelledby to read child components traversely. If spotlight focused component has some components as child, we add aria-labelledby to this spotlight focused component. In CheckboxItem, your opinion also is right but we thought that aria-labelledby should add to spotlight focused components in moonstone.

kbs12e: We have more checked again about above aria-labelledby. 
ChromeVox does not read chekboxItem content when aria-labelledby is added to control's Id.
ChromeVox reads it well when aria-labelledby is only added to client's id. 
However, WebOS TV and Nexus5's screen reader read well in both case, so we will change aria-labelledby's value to client's id for consistency.

PR #2230, opened 2015-06-16
Enyo 1798 tejaswineesohoni

tejaswineesohoni:
ENYO-1798: Added marquee to LabeledTextItem

### DIFF
--- a/lib/LabeledTextItem/LabeledTextItem.js
+++ b/lib/LabeledTextItem/LabeledTextItem.js
@@ -17,7 +17,7 @@ var
 
 /**
 * {@link moon.LabeledTextItem}, which extends {@link moon.Item}, is a
-* [control]{@link enyo.Control} that combines text content with a text label.
+* [control]{@link enyo.Control} that combines text content with a text label and marquees it.
 *
 * @namespace moon
 * @class moon.LabeledTextItem
--- a/lib/LabeledTextItem/LabeledTextItem.js
+++ b/lib/LabeledTextItem/LabeledTextItem.js
@@ -9,11 +9,9 @@ var
 	kind = require('enyo/kind');
 
 var
-	Item = require('../Item');
-
-var 
+	Item = require('../Item'),
 	Marquee = require('../Marquee'),
-	marqueeText = Marquee.Text;
+	MarqueeText = Marquee.Text;
 
 /**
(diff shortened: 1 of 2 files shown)


lib/LabeledTextItem/LabeledTextItem.js
```diff
@@ -44,7 +46,7 @@ module.exports = kind(
 	* @private
 	*/
 	components:[
-		{name: 'label', classes: 'label'},
+		{name: 'label', classes: 'label', kind: marqueeText},
```
viodragon2: Should be MarqueeText

lib/LabeledTextItem/LabeledTextItem.less
```diff
@@ -10,7 +10,6 @@
 .moon-labeledtextitem .label {
 	.moon-sub-header-text;
 	color: @moon-label-text-color;
-	border-top: 3px solid @moon-label-border-color;
 	margin: 0px 0px 3px 0px;
 	padding: 6px 0px 0px 0px;
```
viodragon2: The purpose of the style change is to match the style of ExpandablePicker. Shouldn't we take margin and padding off, too?

aarontam: I think only the `border-top` and `height` were mentioned in the ticket, based on discussion with Gruettner. @Djspaceg, do you have any thoughts on this?

Djspaceg: I believe we can safely drop the `border-top`, `margin`, and `padding`, adding only `margin-bottom: 6px;` in their place. On `.moon-labeledtextitem` above, we can also remove the lines for `height` and `margin`.

* `.label`'s padding just pushes the title farther from the top, so it can be dropped.
* Increasing `.label`'s `margin-bottom` makes its text about the same distance from the top as it is from the value text, which more closely matches ExpandablePicker's style.
* `height: auto` isn't needed (like @ryanjduffy said in Jira) since height is already inherently "auto". One less rule to parse, yay.
* margin on `.moon-labeledtextitem` has no effect, as it's naturally got no margin. Don't need a rule to reiterate it.

aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2348.

PR #2233, opened 2015-06-16
ENYO-1285 : Marquee Not Ellipse 

dmikey:
Issue.

When a marquee is within a parent, that has a 0,0,0,0 Bounds, the the marquee also reports a 0,0,0,0 bounds.

Fix.

As the drawer to re-render after being opened, so that we will ask the children of the drawer to fire a reflow.

### DIFF
--- a/source/ExpandablePicker.js
+++ b/source/ExpandablePicker.js
@@ -370,6 +370,9 @@
 			this.inherited(arguments);
 			this.$.currentValue.setShowing(!this.open);
 			this.setActive(this.getOpen());
+			if(this.open) {
+				this.$.drawer.render();	
+			}
 		},
 
 		/**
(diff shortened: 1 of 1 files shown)


source/ExpandablePicker.js
```diff
@@ -370,6 +370,9 @@
 			this.inherited(arguments);
 			this.$.currentValue.setShowing(!this.open);
 			this.setActive(this.getOpen());
+			if(this.open) {
+				this.$.drawer.render();
```
graynorton: @dmikey, did you try explicitly reflowing, instead of re-rendering? If we can get away with just a reflow, that would be preferable for performance reasons.

dmikey: I could not use reflow on the parent, since it doesn't notify the children. I did create this PR to add a reflow call down through children https://github.com/enyojs/enyo/pull/1212

dmikey:
I didn't like this solution, closing PR in favor of targeted solution

PR #2238, opened 2015-06-17
ENYO-1688 : apply accessibility on moonstone ToggleButton

kbs12e:
Apply accessibility on moonstone ToggleButton
However, ToggleButton's child component has tabIndex, so we should remove it because ToggleButton already has button role and related aria attributes.
Thus, add **accessibilityDisabled:true** to moonstone Button because ToggleButton extends Button.

## Issue

Should read out toggle status when the spotlight is focused to ToggleButton

## Cause

ToggleButton requires press-and-release cycle to change their value.
To support readout toggle-status, we should add aria-pressed attribute to ToggleButton.

## Fix

Add aria-pressed attribute to ToggleButton
Remove tablndex from Button using accessibilityDisabled API.
In addition, add aria-labelledby attribute because screen reader does not reads when child component has tabindex.

### DIFF
--- a/lib/Button/Button.js
+++ b/lib/Button/Button.js
@@ -9,12 +9,14 @@ var
 	kind = require('enyo/kind'),
 	util = require('enyo/utils'),
 	Control = require('enyo/Control'),
-	Button = require('enyo/Button');
+	Button = require('enyo/Button'),
+	options = require('enyo/options');
 
 var
 	Marquee = require('../Marquee'),
 	MarqueeSupport = Marquee.Support,
-	MarqueeText = Marquee.Text;
+	MarqueeText = Marquee.Text,
+	ButtonAccessibilitySupport = require('./ButtonAccessibilitySupport');
 
 /**
 * {@link module:moonstone/Button~Button} is an {@link module:enyo/Button~Button} with Moonstone styling applied.
@@ -46,7 +48,7 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	mixins: [MarqueeSupport],
+	mixins: options.accessibility ? [ButtonAccessibilitySupport, MarqueeSupport] : [MarqueeSupport], 
 
 	/**
 	* @private
@@ -163,7 +165,7 @@ module.exports = kind(
 	*/
 	initComponents: function () {
 		if (!(this.components && this.components.length > 0)) {
-			this.createComponent({name: 'client', kind: MarqueeText, classes: 'button-client', isChrome: true});
+			this.createComponent({name: 'client', kind: MarqueeText, classes: 'button-client', isChrome: true, accessibilityDisabled: true});
 			this.createComponent({name: 'tapArea', kind: Control, classes: 'button-tap-area', isChrome: true});
 		}
 		if (this.small) this.smallChanged();
--- a/lib/ToggleButton/ToggleButtonAccessibilitySupport.js
+++ b/lib/ToggleButton/ToggleButtonAccessibilitySupport.js
@@ -1,0 +1,27 @@
+var
+	kind = require('enyo/kind');
+
+/**
+* @name ToggleButtonAccessibilityMixin
+* @mixin
+*/
+module.exports = {
+
+	/**
+	* @private
+	*/
+	observers: [
+		{method: 'updateAccessibilityAttributes', path: 'value'}
+	],
+
+	/**
+	* @private
+	*/
+	updateAccessibilityAttributes: kind.inherit(function (sup) {
+		return function (was, is, prop) {
+			var enabled = !this.accessibilityDisabled;
+			sup.apply(this, arguments);
+			this.setAttribute('aria-pressed', enabled ? String(this.value) : null);
+		};
+	})
+};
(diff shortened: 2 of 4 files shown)


lib/Button/Button.js
```diff
@@ -165,7 +165,7 @@ module.exports = kind(
 	*/
 	initComponents: function () {
 		if (!(this.components && this.components.length > 0)) {
-			this.createComponent({name: 'client', kind: MarqueeText, isChrome: true});
+			this.createComponent({name: 'client', kind: MarqueeText, isChrome: true, accessibilityDisabled:true});
```
seunghoh: add one space after colon

lib/ToggleButton/ToggleButtonAccessibilitySupport.js
```diff
@@ -0,0 +1,28 @@
...
+
+	/**
+	* @private
+	*/
+	updateAccessibilityAttributes: kind.inherit(function (sup) {
+		return function (was, is, prop) {
+			var enabled = !this.accessibilityDisabled;
+			sup.apply(this, arguments);
+			this.setAttribute('aria-pressed', enabled ? String(this.value) : null);
+			this.setAttribute('aria-labelledby', enabled ? this.getId() : null);
```
ryanjduffy: There's a similar, but more complicated, question to CheckboxItem. In particular, now that you've added `aria-hidden` support, the text within the inner MarqueeText won't be read (at least by ChromeVox) when you set `aria-labelledby` to this.id. However, you can't blindly set it to `this.$.client.getId()` as with CheckboxItem because client may not exist if the ToggleButton was populated with other components.

To support the appropriate extensibility, I'd further suggest that you add a property, perhaps `accessibilityLabeledBy` to this mixin that allows consumers or extenders of ToggleButton specify the name of the control that contains the label. It could default to 'client' so the common scenario of a content-only ToggleButton would work as is.

kbs12e: We will more consider whether or not we provide accessibilityLabeledBy mixin as public API. However, In ToggleButton case, we can set aria-labelledby to this.$.client.getId() because ToggleButton simply extends Button. We checked that screen reader(ChromeVox and TV) reads text within the inner MarqueeText well when we set aria-labelledby to this.$.client.getId() though this.$.client is aria-hidden:true

seunghoh: @ryanjduffy One thing I am concerning is moonstone should consider all platform (including PC ChromVox)? Enyo core definitely should but moonstone? webOS TV, Android talkback and PC Chromvox support ARIA all differently. Specially, chromVox works much less and different from iOS, android and webOS TV Blink. Well...the best is we can support all platform but it makes our code complicated in some cases. Good example is aria-hidden with aria-labelledby. My opinion is we will consider webOS and chromVOX as much as possible, but if moonstone code goes too much complicated then I'd focus on TV case. Anyway, we can do both for togglebutton with bongsub's suggestion.

For making `accessibilityLabelledBy` by wrapping `aria-labelledby`, I feel like not to do right now, we need take more time to check many cases. `aria-labelledby' works mutually exclusive with `aria-label` and if user uses `accessibilityLabel` and `accessibilityLabelledBy` together then we need to make policy all cases. For short term, rather to open API, i'd add `aria-labelledby` inside of component code whenever it needs

CC @graynorton

ryanjduffy: To the support question, I think Enyo will ultimately need to support a broader set of accessibility tooling than moonstone necessarily would. So when support for features like `aria-hidden` and -aria-labelledby` land in the enyo/AccessibilitySupport, we should be conscious of how those features work on key platforms. We don't yet, afaik, have an supported platforms list for accessibility so which platforms would still be up for debate. In the near term, I'm okay with focusing on what webOS supports provided we are clear when we make concessions to that end so we can fix things up later.

You make a valid point regarding the overlap of label and labelledby. That's something we'll need to address. Providing a default for labelledby is a poor design choice since it would complicate determining which value to take.

That said, the issue with ToggleButton still exists. The following code is supported by Moonstone but will not work:

```
{kind: ToggleButton, components: [
  {content: 'a simple control but could be anything'}
]}
```

Before changing to use `this.$.client.getId()`, that example would have worked because the inner control isn't `aria-hidden`. Now, there's a runtime error because `this.$.client` doesn't exist. That was the point I was trying to make originally.

The purpose of adding the labelledby property was to allow the developer a way to inform the mixin which control should be the label when they've customized the components as my example above.

A simpler, and likely just as functional, solution may be to choose either `this.$.client` when it exists and otherwise use the ToggleButton's id. That still puts the reasonably responsibility on the developer to correctly tag any internal controls for accessibility without overcomplicating the logic here.

```
	updateAccessibilityAttributes: kind.inherit(function (sup) {
		return function () {
			var enabled = !this.accessibilityDisabled,
				id = this.$.client ? this.$.client.id : this.id;
			sup.apply(this, arguments);
			this.setAttribute('aria-labelledby', enabled ? id : null);
		};
	})
```

```
// the developer would then need to choose which internal contents would be in the a11y tree.
{kind: ToggleButton, components: [
  {accessibilityDisabled: true, content: 'Shouldn\'t be in the accessibility tree'},
  {content: 'Relevant Content for Accessibility'}
]}
```

seunghoh: Thanks for your advice Ryan, I missed that point.

PR #2239, opened 2015-06-17
ENYO-1635: ExpandableListItem: Focus Does Not Display on Item's Header

akrasnyi:


lishammel:
Should this PR be closed?  Sounds like we're using a different solution?

aarontam:
Hi @lishammel, yep I think we want to go with the solution from https://github.com/enyojs/moonstone/pull/2181.

PR #2241, opened 2015-06-18
DRD-11853: After I select an option of the expandable picker, I can't…

jaewonjang:
… hear any sound.

Blink skip focasable child when parent is focused, so to be able to read
child innertext I set the 'aria-labelledby' on parent node
headerWrapper.

https://jira2.lgsvl.com/browse/DRD-11853


jaewonjang:
This PR has dependency on https://github.com/enyojs/moonstone/pull/2227 because expandable picker is using checkbox item as children components.

jaewonjang:
replace this PR with https://github.com/enyojs/moonstone/pull/2260 because of changing issue number

PR #2245, opened 2015-06-20
ENYO-1761: Toggle Item might be shifted down on TV

kranga-lge:
### Issue:
Toggle Item shifted down on TV but not in browser
### Fix:
For some mysterious reason, 'position: relative' on tv caused icon to be shifted down on tv but not browser. 'position: absolute' worked on both.

Enyo-DCO-1.1-Signed-off-by: Krishna Rangarajan <krishna.rangarajan@lge.com>


ryanjduffy:
Closing in favor of #2315

PR #2246, opened 2015-06-20
ENYO-1330: Update PageUp/Down key handling based on new ux

sbcho:
## Issue
After the scrolling completes following by PageUp/Down key, the top item in the scroll frame does not get re-spotted.

## Fix
Update PageUp/Down key handing based on description of new behavior (in ENYO-1330)

Enyo-DCO-1.1-Signed-off-by: Sungbae Cho sb.cho@lge.com

### DIFF
--- a/css/moonstone-mixins.less
+++ b/css/moonstone-mixins.less
@@ -42,3 +42,12 @@
 	width: 100% !important;
 	height: 100% !important;
 }
+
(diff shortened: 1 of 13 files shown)


sbcho:
Closed since page up/down key doesn't use farthest function anymore

PR #2247, opened 2015-06-22
ENYO-1762: apply backgroundOpacity property to closebutton controls

anishlg:
close buttons in Popup, Dialog and ListActions controls had a css property background-color: transparent, this property is replaced with backgroundOpacity: transparent as a different approch to implement the translucency properties 

### DIFF
--- a/lib/Popup/Popup.less
+++ b/lib/Popup/Popup.less
@@ -31,14 +31,6 @@
 	right: @moon-spotlight-outset;
 	top: @moon-spotlight-outset;
 	margin: 0;
-	background-repeat: no-repeat;
-	background-position: 0 0;
-	color: @moon-popup-close-text-color;
-
-	&.pressed,
-	&:active {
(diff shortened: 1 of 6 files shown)


lib/Popup/Popup.less
```diff
@@ -31,7 +31,6 @@
 	right: @moon-spotlight-outset;
 	top: @moon-spotlight-outset;
 	margin: 0;
-	background-color: transparent;
```
Djspaceg: Line #40 below should also be removed.
```
&:active {
 		background-color: transparent;
```

anishlg:
removed the background-color: transparent property from moon-popup-close active class.

ryanjduffy:
This fix looks good. Let's also remove `@moon-popup-close-text-color` and `@moon-popup-close-spotlight-color` from the dark and light variables LESS files since they appear to be unnecessary now. You'll also need to resolve the merge conflict. Thanks!

PR #2253, opened 2015-06-23
ENYO-1654 : moon.ProgressBar needs a tooltip to show current progress.

suhlee:
## Issue
moon.ProgressBar needs a tooltip(popup) to show current progress.

## Fix
I extracted popup components and made PopupSupport.  Also, I organized slide and progressbar to apply this change.
WIth this change, app developer can use popup support in progresbar.
If user sets "popup:true", then popup will be shown.


ryanjduffy:
You have the right idea to pull the popup-related code out of Slider but I don't think a mixin is the right home. As it exists now, it's too specialized to be reused in another context so instead let's move it into ProgressBar and ensure that Slider has the necessary opportunities to override methods or configure members to customize the popup.

suhlee:
@ryanjduffy , I have changed my PR following your comment. I removed mixin and put popup-related code into progressBar. Please review it, Thanks,

PR #2256, opened 2015-06-23
ENYO-1870: Greatly simplified the logical structure and amount of rules for button

Djspaceg:
* ContextualPopupButton now has the same active state appearance as grouped buttons.
* Removed redundant/excessive code cases (&.active, &.spotlight.pressed, etc)
* Setup all state rules to be either a child of :not(disabled) or [disabled], which will include more cases and be easier to understand what's going on.
* Simplified logic cases for combinations of translucency, spotlight, active, pressed, focused, etc.

### DIFF
--- a/lib/Button/Button.less
+++ b/lib/Button/Button.less
@@ -170,31 +170,8 @@
 		color: inherit;
 	}
 
-	&.spotlight {
-		background-color: @moon-spotlight-color;
-
-		* {
-			color: @moon-spotlight-text-color;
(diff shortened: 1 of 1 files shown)


DavidUM:
Awesome!
How do you come up with this improvement?

Djspaceg:
Thanks! The existing button code just looked too bulky, so I reorganized it, which helped reveal what was redundant or unnecessary.

PR #2259, opened 2015-06-25
ENYO-1852: Time/DatePicker closed status not properly updated with locale change

anishlg:
### Issue:
When the DateTime picker is opened and then changed the locale, the picker collapsed and look like it is closed. But the arrow still points upwards (^) and the property remains as opened.

### Cause:
onLocaleChanged the picker is destroying and recreating the client control  but the open property is not resetting to false.

### Fix:
In refresh() function DateTimePickerBase, the open property resettled to false.


DavidUM:
Looks Good to me. Thank you @anishlg

PR #2260, opened 2015-06-25
ENYO-1858: Accessibility: After I select an option of the expandable …

jaewonjang:
…picker, I can't hear any sound.

- Blink skip focasable child when parent is focused, so to be able to
read
child innertext I set the 'aria-labelledby' on parent node
headerWrapper.
- It doesn't need aria attributes on children components because they
can get the focus, so add accessibilityDisabled true on each child
component of ExpandablePicker

https://jira2.lgsvl.com/browse/ENYO-1858

### DIFF
--- a/lib/ExpandablePicker/ExpandablePickerAccessibilitySupport.js
+++ b/lib/ExpandablePicker/ExpandablePickerAccessibilitySupport.js
@@ -12,9 +12,10 @@ module.exports = {
 	*/
 	updateAccessibilityAttributes: kind.inherit(function (sup) {
 		return function (was, is, prop) {
-			var enabled = !this.accessibilityDisabled;
+			var enabled = !this.accessibilityDisabled,
+				ids = this.$.header.getId() + ' ' + this.$.currentValue.getId();
 			sup.apply(this, arguments);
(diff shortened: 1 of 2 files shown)


lib/ExpandablePicker/ExpandablePickerAccessibilitySupport.js
```diff
@@ -10,18 +10,14 @@ module.exports = {
...
 	*/
-	initAccessibility: kind.inherit(function (sup) {
-		return function (props) {
+	updateAccessibilityAttributes: kind.inherit(function (sup) {
+		return function (was, is, prop) {
+			var enabled = !this.accessibilityDisabled;
 			sup.apply(this, arguments);
-			this.$.currentValue.observe('content', this.changeLabel, this);
-			this.changeLabel();
+			this.$.headerWrapper.setAttribute('aria-labelledby', enabled ? this.$.headerWrapper.getId() : null);
```
ryanjduffy: Guessing you'll have the same labelledby issue here since $.header contains the text but will be `aria-hidden` and therefore not read when its ancestor is set as the labelled by id.

PR #2262, opened 2015-06-25
ENYO-1631: Integrate latest NewDataList, etc. changes driven by SmartShare and EPG

graynorton:


lib/ScrollControls/ScrollControls.js
```diff
@@ -105,6 +103,9 @@ var ScrollControls = module.exports = kind(
 		}
 		if (this.scroller) {
 			this.scroller.on('scrollabilityChanged', this._updateScrollability);
+			this.subControls.forEach(this.bindSafely(function (nom) {
```
ryanjduffy: Is `subControls` intended to be overridden by something? If not, I don't think we're gaining anything with the array and then iterating over it. `forEach` isn't particularly performant as well so, if we need the array, we should probably swap it out for a `for` loop instead.

graynorton: No, you can flatten this out (remove the array and just explicitly set the scroller property on each subcontrol). I don't remember why I did this, but my guess is that at some earlier point I may have been performing another operation on each of the subcontrols and didn't like the brittleness of having to update multiple places in the code if anything changed.

PR #2264, opened 2015-06-26
ENYO-1886: All GridListImageItems should be measured the same way.

Djspaceg:


graynorton:
Closing this, as we instead pulled the fix into the superkind in the layout library (https://github.com/enyojs/layout/pull/134)

PR #2273, opened 2015-07-01
ENYO-1880: Add ability to change indices without transition animation.

aarontam:
### Issue
Based on the changes from https://github.com/enyojs/enyo/pull/1216, we need to update spottability when we utilize `animateTo`, in addition to when the `indexChanged` handler is run.

### Fix
We have factored out spottability update code into its own method, and call that from the appropriate methods.

### DIFF
--- a/lib/LightPanels/LightPanels.js
+++ b/lib/LightPanels/LightPanels.js
@@ -59,8 +59,8 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	indexChanged: function () {
-		this.updateSpottability.apply(this, arguments);
+	indexChanged: function (was, is) {
+		this.updateSpottability(was, is);
 		LightPanels.prototype.indexChanged.apply(this, arguments);
 	},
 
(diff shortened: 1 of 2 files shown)


lib/LightPanels/LightPanels.js
```diff
@@ -51,10 +51,26 @@ module.exports = kind(
...
+	animateTo: function (index) {
+		this.updateSpottability(this.index, index);
+		LightPanels.prototype.animateTo.apply(this, arguments);
+	},
+
+	/**
+	* @private
+	*/
+	indexChanged: function () {
+		this.updateSpottability.apply(this, arguments);
```
ryanjduffy: While effectively identical, I think I'd prefer specifying the arguments on `indexChanged` and explicitly passing them to `updateSpottability` rather than apply. I think it's a bit clearer.

aarontam: Haha, I actually went back and forth on this one; made an update.

PR #2276, opened 2015-07-02
ENYO-1959 ToggleItem: toggle image layout fault

dragongears:


ryanjduffy:
This is fixed by #2269 which fixes the hierarchy so that moonstone/CheckboxItem inherits from moonstone/Item.

PR #2277, opened 2015-07-02
ENYO-1981: Add preTransition, general refactoring.

aarontam:
### Issue
There were some issues with being able to spot and interact with a transitioning panel, and there was a need to maintain focus on a panel, even if there were no spottable items (otherwise the last spotted control would be spotted).

### Fix
We have implemented a `preTransition` method that sets the appropriate `Spotlight` state of a given panel based  on whether or not it is active, and we have added a dummy `Spotlight` control. This change is also dependent on https://github.com/enyojs/enyo/pull/1222.


lib/LightPanel/package.json
```diff
@@ -1,6 +0,0 @@
-{
-	"main": "LightPanel.js",
-	"styles": [
-		"LightPanel.less"
```
ryanjduffy: This isn't included in the build any longer because it's not referenced by LightPanels/package.json

lib/LightPanels/LightPanel.js
```diff
@@ -0,0 +1,200 @@
...
+			this.$.header.createComponents(this.headerComponents, {owner: owner});
+		}
+	},
+
+	/**
+	* This overridable method is called before a transition.
+	*
+	* @public
+	*/
+	preTransition: function () {
```
ryanjduffy: Call this optional feedback - may want to call the super here (and for `postTransition`) even though it's a no-op right now so we don't forget later. Certainly a debateable point.

lib/LightPanels/LightPanels.js
```diff
@@ -38,63 +35,8 @@ module.exports = kind(
...
-		this._hasSpotlightFocus = true;
-	},
-
-	/**
-	* @private
-	*/
-	leave: function () {
-		this._hasSpotlightFocus = false;
-	}
+	defaultKind: LightPanel
```
ryanjduffy: More optional ... I realize it's not needed right now but should we make it a practice that all kinds have at least one class (e.g. `moon-lightpanels') for ease in customization?

lib/LightPanels/LightPanel.js
aarontam: Yeah thought about that but opted not to because of the no-op. I agree it does set us up better for future work, and is self-documenting in that sense.

lib/LightPanels/LightPanel.js
```diff
@@ -0,0 +1,200 @@
...
+
+	/**
+	* @private
+	*/
+	kind: LightPanel,
+
+	/**
+	* @private
+	*/
+	classes: 'moon-light-panel',
```
ryanjduffy: This could be a breaking change (ug) but I think the usual convention is `moon-lightpanel`. We're not consistent in that so if it feels too painful, let's leave it.

lib/LightPanel/package.json
aarontam: Great catch - thanks!

lib/LightPanels/LightPanels.js
aarontam: This is a good question. I'd normally err on the side of including it, but I'd want to nail down our naming convention first (if I use `moon-lightpanels` it won't be consistent with `enyo-light-panels` or `moon-light-panel`, but those aren't easily changeable right now without some other reworking of the app, not to mention the other Moonstone controls that we should update as well). Might need to take a step back and examine our controls's class names as a whole.

PR #2281, opened 2015-07-07
ENYO-1990 Overlay: Text Overlay Offsets to the Right

akrasnyi:


lishammel:
Closing out - Jira was closed with another fix.

PR #2285, opened 2015-07-08
ENYO-1899 : Add Text to Speech Accessibility support to SelectableItem

kbs12e:
## Issue 
Screen reader does not read SelectableItem content and status when SelectableItem is focused  

## Cause
SelectableItem looks like general radioItem, which has text and radio mark. To support accessibility, we have to add aria-related role and attribute. In SelectableItem, role is 'radio' and it needs 'aria-checked' attribute to inform whether or not SelectableItem is checked. In Addtion to, SelectableItem extends moonstone Item, which can have child components like icon or simple control, so we have to add aria-labelledby attribute for reading child content to moonstone Item.

## Fix
add radio role , aria-checked and aria-labelledby attribute to SelectableItemAccessibilitySupport mixin

### DIFF
--- a/lib/CheckboxItem/CheckboxItemAccessibilitySupport.js
+++ b/lib/CheckboxItem/CheckboxItemAccessibilitySupport.js
@@ -32,7 +32,7 @@ module.exports = {
 			this.setAttribute('role', enabled ? 'checkbox' : null);
 			this.setAttribute('tabindex', enabled ? 0 : null);
 			this.setAttribute('aria-checked', enabled ? String(this.checked) : null);
-			this.setAttribute('aria-labelledby', enabled ? this.$.client.getId() : null);
+			this.setAttribute('aria-labelledby', !this.accessibilityLabel && enabled ? this.$.client.getId() : null);
 		};
 	})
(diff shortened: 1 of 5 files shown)


lib/CheckboxItem/CheckboxItemAccessibilitySupport.js
```diff
@@ -32,7 +32,7 @@ module.exports = {
 			this.setAttribute('role', enabled ? 'checkbox' : null);
 			this.setAttribute('tabindex', enabled ? 0 : null);
 			this.setAttribute('aria-checked', enabled ? String(this.checked) : null);
-			this.setAttribute('aria-labelledby', enabled ? this.$.client.getId() : null);
+			this.setAttribute('aria-labelledby', !this.accessibilityLabel && enabled ? this.$.client.getId() : null);
```
kbs12e: CheckboxItem also has aria-labelledby attribute because It extends Item, so If user add accessibilityLabel to CheckboxItem, CheckboxItem should not have aria-labelledby attribute for reading accessibilityLabel text

PR #2288, opened 2015-07-10
ENYO-1986 [Moon.Drawers]-Control drawers's height is not updated when its expan…

srinivasv123:
…dables expands.

Issue: When components inside control drawer are expanded control
drawer's height is not getting updated

Cause: issue because of regression, in new implementation of drawer's no
event is handling the height of controlDrawer when it's inside
components are expanded.

Fix: handling core drawer's animationStep to update the control height
of drawer.

Enyo-DCO-1.1-Signed-off-by: Srinivas V srinivas,v@lge.com

### DIFF
--- a/lib/Drawers/Drawer.js
+++ b/lib/Drawers/Drawer.js
@@ -182,7 +182,8 @@ var MoonDrawer = module.exports = kind(
 	*/
 	handlers: {
 		ontransitionend: 'transitionEnded',
-		onwebkittransitionend: 'transitionEnded'
+		onwebkittransitionend: 'transitionEnded',
+		onDrawerAnimationStep: 'updateControlDrawersHeight'
 	},
 
 	/**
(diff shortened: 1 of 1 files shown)


lib/Drawers/Drawer.js
```diff
@@ -182,7 +182,8 @@ var MoonDrawer = module.exports = kind(
 	*/
 	handlers: {
 		ontransitionend: 'transitionEnded',
-		onwebkittransitionend: 'transitionEnded'
+		onwebkittransitionend: 'transitionEnded',
+		onDrawerAnimationStep: 'updateControlDrawersHeight'
```
DavidUM: Every single animation step triggers  `onDrawerAnimationStp` event. 
I do not think it is a good way for performance.

lib/Drawers/Drawer.js
```diff
@@ -228,6 +229,18 @@ var MoonDrawer = module.exports = kind(
...
 
 	/**
+	* @private
+	*/
+	updateControlDrawersHeight: function(){
+		if(this.controlDrawerComponents !== null ) {
+			var value = this.calcDrawerHeight(),
+				node = this.$.controlDrawer.hasNode(),
+				cHeight = node.getBoundingClientRect().height + value;
+			this.applyStyle('height', cHeight + 'px');
```
DavidUM: `applyStyle` is kind a heavy job for performance aspect.
Should we do those work in every animation step?

DavidUM:
Hi @srinivasv123 
Thank you for your nice work. 
However, this issue is a regression issue. In 2.5-upkeep, I can't reproduce it.
That means, there is a root cause which makes this regression.
Could you investigate more and find what is it?

I think you solution could help this issue. But I'm not sure it is good for performance.
Thank you @srinivasv123

lishammel:
Closing PR.

PR #2294, opened 2015-07-13
ENYO-1892: Add Text to Speech Accessibility support to SimplePicker

jaewonjang:
Blink will support 'aria-valuetext' for spinbutton, so we apply
spinbutton role for SimplePicker.
We don't need to manage children controls accessibilityDisabled
because if it set 'aria-hidden' true on parent, children will be
hidden as well.

https://jira2.lgsvl.com/browse/ENYO-1892

### DIFF
--- a/lib/SimplePicker/SimplePicker.js
+++ b/lib/SimplePicker/SimplePicker.js
@@ -8,13 +8,15 @@ require('moonstone');
 var
 	kind = require('enyo/kind'),
 	dom = require('enyo/dom'),
+	options = require('enyo/options'),
 	Control = require('enyo/Control');
 
 var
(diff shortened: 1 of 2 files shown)


seunghoh:
On marquee text (eg. San Francisco Airport...) of the simple picker, screen reader keep reading since simple picker has `aria-live`. Simple picker content doesn't have a focus, `aria-live` seems to be mandatory. @ryanjduffy @graynorton do you have any idea?

jaewonjang:
Blink will support 'aria-valuenow' for spinbutton, so we use 'spinbutton' role on SimplePicker to avoid reading marquee text continuosly.

PR #2299, opened 2015-07-14
Enyo-2091-davidum Re-compose this.pickers

DavidUM:
Issue
-------
When we open timePicker, it does not reflow every child pickers

Cause
--------
timeWrapper encloses hour and minute picker

Fix
----
Re compose set of pickers.

### DIFF
--- a/source/DateTimePickerBase.js
+++ b/source/DateTimePickerBase.js
@@ -197,7 +197,7 @@
 		* @private
 		*/
 		setupPickers: function (ordering) {
-			// implement in subkind, calling this.inherited() at the end
+			// implement in subkind
 		},
 
 		/**
@@ -336,7 +336,7 @@
 				delete this._tf;
 			}
 			if (this.value && typeof ilib !== 'undefined'){
-				this.localeValue = ilib.Date.newInstance({unixtime: this.value.getTime(), timezone: "local"});
+				this.localeValue = ilib.Date.newInstance({unixtime: this.value.getTime(), timezone: 'local'});
 			}
 			this.initDefaults();
 			this.render();
--- a/source/DateTimePickerBase.js
+++ b/source/DateTimePickerBase.js
@@ -198,7 +198,7 @@
 		*/
 		setupPickers: function (ordering) {
 			// implement in subkind, calling this.inherited() at the end
-			this.pickers = this.getClientControls();
+			this.pickers = this.pickers || this.getClientControls();
 		},
 
 		/**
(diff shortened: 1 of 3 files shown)


source/TimePicker.js
```diff
@@ -479,10 +479,24 @@
...
 					break;
 				}
-	
 			}
 	
 			this.inherited(arguments);
+
+			// in case of timePicker, timeWrapper enclose hour and minute pickers
+			// so, we re-organize this.pickers
+			var pickers = this.pickers;
```
aarontam: It seems a little awkward that we're allowing the super method to naïvely retrieve the client controls and assign that to `this.pickers`, only for us to reassign that to an array we build here. Also, I'd want to prevent unnecessarily looping through the controls when we have enough knowledge from the operations above. Instead, I'm wondering if we can build the array in the above switch statement (we would need to modify `wrapComponent` to return the created component, which should only be additive and wouldn't be regressive), and modify `DateTimePickerBase`'s `setupPickers` method to something like this:
```
this.pickers = this.pickers || this.getClientControls();
```
Any thoughts on this? I don't see any obvious cases where `this.pickers` would be incorrectly truthy when we run `DateTimePickerBase`'s `setupPickers` method, as this array will always be regenerated (in the case of `TimePicker`) or `null` (in the case of `DatePicker` via `reset`).

DavidUM: Hi @aarontam , Thank you for your review.
Unfortunately, I didn't consider much today due to pre.11.1 release.
But I made a simple change.
Could you check this is a same way which you commented?
Thank you.

source/TimePicker.js
```diff
@@ -455,9 +457,10 @@
 						]},
 						this
 					);
+					pickers.push(createdComponent.getClientControls()[0]);
```
aarontam: I see why you need to do this from a logistical perspective. Going back to the change where the wrapper control around the picker was added (i.e. with the `moon-date-picker-wrap`) class, `this.pickers` has been an array consisting of wrapper controls, rather than the actual hour or minute picker, so it has been incorrect for a while. I think we will need to do something similar for `DatePicker` as well, right?

DavidUM: Oh, it is my bad. I left one commit in local branch. Sorry to make you confused

source/DateTimePickerBase.js
```diff
@@ -198,7 +198,6 @@
 		*/
 		setupPickers: function (ordering) {
 			// implement in subkind, calling this.inherited() at the end
```
aarontam: I think we'll want to remove this comment if we are no longer calling `this.inherited` from the `setupPickers` methods of subkinds.

PR #2303, opened 2015-07-15
ENYO-1903: Add Text to Speech Accessibility support to Expandable Input

jaewonjang:
According to UX guide it should read
<header> + <current value> + <edit box>,
so I added 'aria-label' combining all of them.

https://jira2.lgsvl.com/browse/ENYO-1903

### DIFF
--- a/lib/ExpandableInput/ExpandableInputAccessibilitySupport.js
+++ b/lib/ExpandableInput/ExpandableInputAccessibilitySupport.js
@@ -1,0 +1,31 @@
+var
+	kind = require('enyo/kind');
+var
+	$L = require('../i18n');
+/**
+* @name ExpandableInputAccessibilityMixin
+* @mixin
+*/
+module.exports = {
+
+	/**
+	* @private
+	*/
+	observers: [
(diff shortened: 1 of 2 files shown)


lib/ExpandableInput/ExpandableInputAccessibilitySupport.js
```diff
@@ -0,0 +1,31 @@
...
+		{method: 'updateAccessibilityAttributes', path: ['value']}
+	],
+
+	/**
+	* @private
+	*/
+	updateAccessibilityAttributes: kind.inherit(function (sup) {
+		return function (was, is, prop) {
+			var enabled = !this.accessibilityDisabled,
+				text = this.$.header.getContent() + ' ' + this.currentValueText() + ' ' + $L('edit box');
```
seunghoh: Check if TTS reads OK for two spaces when currentValueText has no data.

seunghoh: Tested and reads well

PR #2311, opened 2015-07-16
Enyo-2091-davidum Re-compose this.pickers

DavidUM:
Issue
-------
When we open timePicker, it does not reflow every child pickers

Cause
--------
timeWrapper encloses hour and minute picker

Fix
----
Re compose set of pickers.

### DIFF
--- a/lib/DatePicker/DatePicker.js
+++ b/lib/DatePicker/DatePicker.js
@@ -121,7 +121,7 @@ module.exports = kind(
 	initILib: function () {
 		DateTimePickerBase.prototype.initILib.apply(this, arguments);
 		if (this.value) {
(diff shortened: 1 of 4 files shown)


lib/DatePicker/DatePicker.js
```diff
@@ -174,6 +174,7 @@ module.exports = kind(
 	* @private
 	*/
 	setupPickers: function (ordering) {
+		var pikcers = [];
```
aarontam: I think you meant `pickers` here. :smile:

aarontam:
There was an unintentionally-included commit; closing in favor of https://github.com/enyojs/moonstone/pull/2326.

PR #2316, opened 2015-07-17
Enyo-2034 Close popup using activated method

DavidUM:
Issue
-------
Activate method could not close Popup

Cause
--------
This is a regression issue of modular work.
Added line was missing during modulization.

Fix
----
Restore missing line

### DIFF
--- a/src/ContextualPopupDecorator/ContextualPopupDecorator.js
+++ b/src/ContextualPopupDecorator/ContextualPopupDecorator.js
@@ -96,6 +96,8 @@ module.exports = kind(
 				this.activator.addClass('active');
 				this.requestShowPopup();
 			}
+		} else {
+			this.requestHidePopup();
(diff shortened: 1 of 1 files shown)


webOS101:
The popup hide call cannot be inside the guard because the active will be false by the time it gets here, unless there's some test case I'm unaware of.

DavidUM:
If you mentioned `active` of contextualPopupButton, it is already updated when `activated()` is called.
Do you mean this?

webOS101:
No, I mean that the fix does not work as implemented because it can never get to the code after you moved it in the last patch.  `inEvent.originator.active` will already be false. If that step is reverted then the code appears to work.

webOS101:
I have rebased the code and tested against other ContextualPopup bugs to be sure we don't encounter any issues.

PR #2318, opened 2015-07-17
Enyo-2133 Prevent toggling selected property

DavidUM:
when we tap same button sequentialy.


lib/Slider/Slider.js
```diff
@@ -655,7 +655,8 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	spotSelect: function () {
+	spotSelect: function (sender) {
+		if (sender != this) return;
```
aarontam: I think we want to use a `===` check here.

aarontam:
I'm wondering if we could just `return true` from both the `previous` and `next` methods, instead.

DavidUM:
Hi @aarontam 
Sorry for late response. I was on vacation 

If we `return true` in previous() and next(), `onSpotlightSelect` event could not be delivered to `spotSelect()`

PR #2321, opened 2015-07-17
PLAT-1995: Reset TimePicker dropdown

sukharenko:
PLAT-1995: The time picker value's which is under the time is not changed to default after selecting 'Reset' time option

Call scrollToValue for hour and minute components


aarontam:
This appears to be resolved by https://github.com/enyojs/moonstone/pull/2326. Proposing that we close this out @sukharenko.

PR #2322, opened 2015-07-17
Fixes ENYO-2030

ryanjduffy:
## Issue
Most (all?) of the kinds extending moonstone/ExpandableListItem are incorrectly reusing CSS classes from moonstone/ExpandablePicker resulting in incorrect styling when moonstone/ExpandablePicker is not included in a build.

## Fix
Refactor moonstone/ExpandablePicker css into moonstone/ExpandableListItem which the former extends

### DIFF
--- a/lib/ExpandablePicker/ExpandablePicker.less
+++ b/lib/ExpandablePicker/ExpandablePicker.less
@@ -2,18 +2,7 @@
 .moon-expandable-picker-header {
 	margin: 0px;
 	display: inline-block;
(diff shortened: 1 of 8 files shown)


lib/ExpandablePicker/ExpandablePicker.less
```diff
@@ -2,18 +2,7 @@
 .moon-expandable-picker-header {
 	margin: 0px;
 	display: inline-block;
```
aarontam: I think we'll need to copy this declaration to ExpandableListItem.less.

PR #2324, opened 2015-07-20
ENYO-2131-inconsistent behavior of spotlight on auto collapse

rajyavardhanp:
first item in the client area, and close the list, f the focus moves out
of the drawer and auto-collapse is true.

### DIFF
--- a/lib/ExpandableListItem/ExpandableListItem.js
+++ b/lib/ExpandableListItem/ExpandableListItem.js
@@ -168,6 +168,7 @@ module.exports = kind(
 	* @private
 	*/
 	handlers: {
+		onSpotlightUp: 'spotlightUp',
 		onSpotlightDown: 'spotlightDown',
 		onDrawerAnimationEnd: 'drawerAnimationEnd'
 	},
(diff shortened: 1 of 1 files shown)


lib/ExpandableListItem/ExpandableListItem.js
```diff
@@ -289,6 +290,19 @@ module.exports = kind(
 			this.bubble('onRequestScrollIntoView');
 		}
 	},
+	
+	/**
+	* Check for the first item in the client area, and close the list
+	* if the focus moves out of the drawer and auto-collapse is true
+	*
+	* @private
+	*/
```
KunmyonChoi: l

lib/ExpandableListItem/ExpandableListItem.js
```diff
@@ -289,6 +290,19 @@ module.exports = kind(
...
+	spotlightUp: function(inSender, inEvent) {
+		var children = Spotlight.getChildren(this.$.client);
+		if (children.length && inEvent.originator == children[0] && this.getAutoCollapse() && this.getOpen()) {
```
sbcho: @rajyavardhanp Do you have a reason for checking the event comes from first spotlight item?
If there is no condition {{children.length && inEvent.originator == children[0]}}, are there specific problem?

sbcho: @rajyavardhanp I think below condition sequence looks better because it doesn't need to check first item if autoCollapse is false
{code}
if (this.getAutoCollapse() && this.getOpen() && children.length && inEvent.originator == children[0])
{code}

lib/ExpandableListItem/ExpandableListItem.js
```diff
@@ -289,6 +290,19 @@ module.exports = kind(
 			this.bubble('onRequestScrollIntoView');
```
sbcho: @rajyavardhanp I think above code isn't needed anymore, right? :)

{code}
		var direction = inEvent && inEvent.dir;

		if (this.getOpen() && this.getAutoCollapse() && direction === 'UP') {
			this.setActive(false);
		}
{code}

PR #2329, opened 2015-07-21
ENYO-1895: Add Text to Speech Accessibility support to Slider

jaewonjang:
Apply accessibility on Slider.
Add 'slider' role and 'aria-valuenow'.

https://jira2.lgsvl.com/browse/ENYO-1895

### DIFF
--- a/lib/Slider/SliderAccessibilitySupport.js
+++ b/lib/Slider/SliderAccessibilitySupport.js
@@ -1,0 +1,29 @@
+var
+	kind = require('enyo/kind');
+
+/**
+* @name SliderAccessibilityMixin
+* @mixin
+*/
+module.exports = {
+
+	/**
+	* @private
+	*/
+	observers: [
+		{method: 'updateAccessibilityAttributes', path: ['value']}
+	],
+
+	/**
+	* @private
+	*/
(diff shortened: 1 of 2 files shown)


lib/Slider/SliderAccessibilitySupport.js
```diff
@@ -0,0 +1,33 @@
...
+		return function (was, is, prop) {
+			var enabled = !this.accessibilityDisabled;
+			sup.apply(this, arguments);
+			if (this.popup) {
+				this.setAttribute('aria-label', enabled ? this.value : null);
+			} else {
+				this.setAttribute('role', enabled ? 'slider' : null);
+				this.setAttribute('aria-valuenow', enabled ? this.value : null);
+			}
+			this.setAttribute('tabindex', enabled ? 0 : null);
```
ryanjduffy: Why do we need to set tabindex here when it's also done in enyo/AccessibilitySupport?

jaewonjang:
In enyo/AccessibilitySupport we are adding tabindex in case focusable is true, but slider has no content and it is not focusable. We need tabindex making slider focusable.

jaewonjang:
I'll create another PR because of too many commit.

PR #2332, opened 2015-07-21
Fixes ENYO-2144

ryanjduffy:
## Issue

Expandable controls are duplicating component tree and features
## Fix

Refactor things for better reuse

Enyo-1.1-DCO-Signed-off-by: Ryan Duffy ryan.duffy@lge.com


lib/ExpandableInput/ExpandableInput.js
```diff
@@ -180,11 +163,12 @@ module.exports = kind(
...
 	* @private
 	*/
-	currentValueText: function () {
-		return (this.value === '') ? this.noneText : this.value;
+	currentValueShowing: function () {
+		return !this.open && this.currentValueText() !== '' && this.type != 'password';
 	},
 
 	/**
+<<<<<<< HEAD
```
aarontam: Some merge conflict remnants. :smile:

lib/Accordion/Accordion.js
```diff
@@ -89,33 +81,5 @@ module.exports = kind(
...
-		]}
-	],
-
-	/**
-	* @private
-	*/
-	bindings: [
-		{from: '.disabled', to: '.$.headerWrapper.disabled'}
-	]
+	classes: 'moon-accordion'
```
aarontam: I wonder if, somewhere down the road, we can deprecate this kind and just make this a special case of `ExpandableListItem` i.e. enabling via a property of that kind.

lib/ExpandableInput/ExpandableInput.js
```diff
@@ -294,29 +266,11 @@ module.exports = kind(
...
 			this.focusInput();
 		} else {
-			this.closeDrawerAndHighlightHeader();
+			this.expandContract();
 		}
 		return true;
-	},
-
-	/**
-	* We manually set pointer mode to `false` as this seems to be the
```
aarontam: Glad we could get rid of this :+1:

lib/LabeledTextItem/LabeledTextItem.js
```diff
@@ -86,21 +72,18 @@ module.exports = kind(
...
-	},
-	
-	/**
-	* @private
-	*/
-	textChanged: function() {
-		this.$.text.setContent(this.text);
-	}
+	components:[
+		// headerContainer required to avoid bad scrollWidth returned in RTL for certain text widths (webkit bug)
```
aarontam: Wonder if we should re-verify this behavior with Blink while we're at it.

lib/Accordion/Accordion.js
ryanjduffy: Agreed. Kept it but marked it deprecated.

lib/ExpandableInput/ExpandableInput.js
ryanjduffy: Ya, looks like this can be removed. Verified the old bug doesn't repro.

PR #2333, opened 2015-07-22
ENYO-2158: Fix guard code in `avoidScrollOnDefaultFocus`

graynorton:
Old guard code didn't account for the case where `event.index` was
zero.

### DIFF
--- a/lib/NewDataList/NewDataList.js
+++ b/lib/NewDataList/NewDataList.js
@@ -82,7 +82,7 @@ module.exports = kind({
 		// potentially not in view, and focusing would cause it to
 		// scroll it into view for no reason apparent to the user.
 		// Instead, we just focus the first visible list element.
-		if (event.focusType === 'default' && event.index) {
+		if (event.focusType === 'default' && typeof event.index === 'number') {
 			fv = this.getFullyVisibleItems()[0] || this.getVisibleItems()[0];
 			if (fv) {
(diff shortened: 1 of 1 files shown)


lib/NewDataList/NewDataList.js
```diff
@@ -82,7 +82,7 @@ module.exports = kind({
 		// potentially not in view, and focusing would cause it to
 		// scroll it into view for no reason apparent to the user.
 		// Instead, we just focus the first visible list element.
-		if (event.focusType === 'default' && event.index) {
+		if (event.focusType === 'default' && typeof event.index === 'number') {
```
ryanjduffy: In other places, we've performed this kind of check:

```
(event.index || event.index === 0)
```

Curious if there's a perf difference between that and `typeof`. I'd try with jsperf but, alas, it's still down. Any ideas?

graynorton: Doing some really simple JSPerf-esque experiments in the console (Mac Chrome and TV) seems to indicate that the performance of these checks is essentially the same.

KunmyonChoi: If original issue was aiming to fix the case of focused item is off screen (Not inside of scroller viewort), the can we compare event.index with firstFullyVisibleI and lastFullyVisibleI to filter out that item doesn't need unnecessary scroll in?

PR #2334, opened 2015-07-23
ENYO-2156-inputMode doesn't take proper classes in moon.Header

rajyavardhanp:
InputHeader classes are moved to Header, as that is the only dependency
for InputHeader.


otisnate:
If you want to remove every css in ```InputHeader.less```, I think it's better to remvoe ```InputHeader.less``` and modify ```package.json``` file

rajyavardhanp:
Hi @otisnate I changed the PR based on your comments. 
Thankyou.

aarontam:
If we're going this route, we should also remove the `moon-input-header` class from `moonstone/InputHeader` as this is already being applied by `moonstone/Header` and we don't want this unnecessarily in multiple locations.

webOS101:
I removed the unneeded class, removed the InputHeader directory.

PR #2335, opened 2015-07-23
ENYO-2128 : Fix reading text twice on marquee item

kbs12e:
## Issue 
If the string of the SelectableItem(or RadioItem) has a long text truncation, screen reader reads string twice.

## Cause
Screen reader reads spotlight focused text when audio guidance is on and component is focused with 5-way mode. However, if component internally has marquee mixin for reading long text like moonstone SelectableItem, this issue is reproduced. 
Screen reader tries to read dom content but marquee mixin creates marqueeWrapper component dynamically as soon as component is focused, so screen reader first reads content and text of marquee that is created after focusing.

## Fix
add aria-hidden:true when marquee is started and remove it when component is blured.


lib/Item/ItemAccessibilitySupport.js
```diff
@@ -18,11 +18,39 @@ module.exports = {
...
 		};
-	})
+	}),
+
+	/**
+	* @private
+	*/
+	marqueeStartHandler: function (oSender, oEvent) {
+		var enabled = !this.accessibilityDisabled;
+		if (this._marquee_distance !== 0 && !this.$.marqueeText && enabled) {
```
kbs12e: if componen's content is short, marquee is not necessary and _marquee_distance property is zero. if marqueeWrapper dom is already created, screen reader reads content once.

lib/Item/ItemAccessibilitySupport.js
```diff
@@ -18,11 +18,39 @@ module.exports = {
 	/**
 	* @private
 	*/
+	handlers: {
```
ryanjduffy: I'm concerned about adding handlers in the mixin as they'll stomp any existing handlers defined for these events. We could pursue a similar approach to Marquee and override dispatchEvent to handle these. If that's the route we take, I think we'll want to generalize that into a reusable spot (I'm thinking as another export to `enyo/dispatcher`) so both this and moonstone/Marquee can take advantage.

lib/Item/ItemAccessibilitySupport.js
ryanjduffy: Relying on private members from another mixin will make this brittle. I'm not sure what the 'less-brittle' solution is here. It feels like we need to someone better integrate accessibility with Marquee rather than relying on the Item mixin to handle it. Thoughts?

kbs12e:
This branch has a lot of changed commit, so I will re-create this branch with last commit.

PR #2336, opened 2015-07-23
ENYO-426: Marquee shows ellipsis but Does Not start rolling in moon.InputHeader Enyo-DCO-1.1-Signed-off-by: Artem Krasnyi <krasnyi.artem@lge.com>

akrasnyi:
…nputHeader

### DIFF
--- a/src/Header/Header.less
+++ b/src/Header/Header.less
@@ -237,3 +237,7 @@
 		float: right;
 	}
 }
+
+.hidden {
+    display: none !important;
+}
--- a/src/Header/Header.js
+++ b/src/Header/Header.js
@@ -277,7 +277,8 @@ module.exports = kind(
 		oninput: 'handleInput',
 		onchange: 'handleChange',
 		onRequestCreateListActions: 'handleRequestCreateComponents',
-		onListActionOpenChanged: 'handleListActionOpenChanged'
+		onListActionOpenChanged: 'handleListActionOpenChanged',
+        onSpotlightKeyDown: 'handleSpotlightKeyDown'
 	},
 
 	/**
@@ -306,7 +307,7 @@ module.exports = kind(
 			{name: 'title', kind: MarqueeText, classes: 'moon-header-text moon-header-title', canGenerate: false},
 			{name: 'inputDecorator', kind: InputDecorator, classes: 'moon-input-header-input-decorator',canGenerate: false, components: [
                 {name: 'titleInput', kind: Input, classes: 'moon-header-text moon-header-title',  onblur: 'marqueeToggle'},
-                {name: 'titleInputMarquee', kind: MarqueeText, classes: 'moon-header-text moon-header-title moon-input header-marquee', ontap: 'marqueeToggle', showing: false}
+                {name: 'titleInputMarquee', kind: MarqueeText, classes: 'moon-header-text moon-header-title moon-input header-marquee', ontap: 'marqueeToggle', onkeyup: 'marqueeToggle', showing: false}
 			]}
 		]},
 		{name: 'titleBelow', kind: MarqueeText, classes: 'moon-sub-header-text moon-header-title-below'},
@@ -801,17 +802,28 @@ module.exports = kind(
     handleChange: function (inSender, inEvent) {
         this.doInputHeaderChange({originalEvent: util.clone(inEvent, true)});
         this.$.titleInputMarquee.set('content', inEvent.originator.getValue());
-        this.marqueeToggle();
     },
 
-        /**
-         * Logic for replacing input into marquee
-         *
-         * @private
-         */
+    /**
+    * Handling Enter press, to act as tap.
+    *
+    * @private
+    */
 
-        marqueeToggle: function (){
-            if (this.$.titleInputMarquee.get('content') === '') return;
+    handleSpotlightKeyDown: function (inSender, inEvent) {
+        if (inEvent.keyCode == 13) {
+            this.marqueeToggle();
+        }
+    },
+
+    /**
+    * Logic for replacing input into marquee
+    *
+    * @private
+    */
+
+    marqueeToggle: function (){
+        if (this.$.titleInputMarquee.get('content') === '') return;
 
             var isInputShowing = this.$.titleInput.get('showing');
             this.$.titleInput.set('showing', !isInputShowing);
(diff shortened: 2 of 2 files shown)


lib/Header/Header.less
```diff
@@ -182,4 +182,12 @@
...
 		}
 	}
+}
+
+.hidden {
+  display: none!important;
+}
+
+.moon-input-header .moon-input-header-input-decorator .moon-input.header-marquee {
+  line-height: 1.24em;
```
aarontam: Where did this number come from? I noticed that when switching between the Input and the Marquee, there was a noticeable shift in the positioning of the controls.

lib/Header/Header.js
```diff
@@ -796,9 +797,31 @@ module.exports = kind(
...
+
+        /**
+         * Logic for replacing input into marquee
+         *
+         * @private
+         */
+
+        marqueToggle: function (){
+            if (this.$.titleInputMarquee.getContent() === '') return ;
+            if (this.$.titleInput.hasClass('hidden')){
```
aarontam: Can't we simplify this to the following (and remove the `inputBlur` handler`)?

```
var isInputShowing = this.$.titleInput.get('showing');
this.$.titleInput.set('showing', !isInputShowing);
this.$.titleInputMarquee.set('showing', isInputShowing);
```

lib/Header/Header.js
```diff
@@ -796,9 +797,31 @@ module.exports = kind(
...
+        this.marqueToggle();
+    },
+
+        /**
+         * Logic for replacing input into marquee
+         *
+         * @private
+         */
+
+        marqueToggle: function (){
```
aarontam: This should be `marqueeToggle`.

lib/Header/Header.js
```diff
@@ -305,7 +305,8 @@ module.exports = kind(
 		{name: 'titleWrapper', kind: Control, classes: 'moon-header-title-wrapper', components: [
 			{name: 'title', kind: MarqueeText, classes: 'moon-header-text moon-header-title', canGenerate: false},
 			{name: 'inputDecorator', kind: InputDecorator, classes: 'moon-input-header-input-decorator',canGenerate: false, components: [
-				{name: 'titleInput', kind: Input, classes: 'moon-header-text moon-header-title'}
+                {name: 'titleInput', kind: Input, classes: 'moon-header-text moon-header-title',  onblur: 'inputBlur'},
+                {name: 'titleInputMarquee', kind: MarqueeText, classes: 'moon-header-text moon-header-title moon-input header-marquee hidden', marqueeOnRender: true, ontap: "marqueToggle"}
```
aarontam: There's no need for the `hidden` class as we have a `showing` property that can be set for Controls.

lib/Header/Header.js
```diff
@@ -796,9 +797,31 @@ module.exports = kind(
 	* @fires module:moonstone/Header~Header#onInputHeaderChange
 	* @private
 	*/
-	handleChange: function (inSender, inEvent) {
-		this.doInputHeaderChange({originalEvent: util.clone(inEvent, true)});
-	},
+    handleChange: function (inSender, inEvent) {
+        this.doInputHeaderChange({originalEvent: util.clone(inEvent, true)});
+        this.$.titleInputMarquee.setContent(event.originator.getValue());
```
aarontam: Let's use the generic setter here i.e.

```
this.$.titleInputMarquee.set('content', event.originator.getValue());
```

lib/Header/Header.js
```diff
@@ -796,9 +797,31 @@ module.exports = kind(
...
+
+        marqueToggle: function (){
+            if (this.$.titleInputMarquee.getContent() === '') return ;
+            if (this.$.titleInput.hasClass('hidden')){
+                this.$.titleInput.removeClass('hidden');
+                this.$.titleInputMarquee.addClass('hidden');
+            }
+        },
+
+        inputBlur: function () {
```
aarontam: This can probably be removed, based on the above comment.

lib/Header/Header.js
aarontam: This should also reference the event parameter, which is `inEvent` in this case.

lib/Header/Header.js
aarontam: Also, do we actually need to set `marqueeOnRender:true` here? And we should surround the tap handler with single quotes, not double quotes.

lib/Header/Header.js
```diff
@@ -305,7 +305,8 @@ module.exports = kind(
 		{name: 'titleWrapper', kind: Control, classes: 'moon-header-title-wrapper', components: [
 			{name: 'title', kind: MarqueeText, classes: 'moon-header-text moon-header-title', canGenerate: false},
 			{name: 'inputDecorator', kind: InputDecorator, classes: 'moon-input-header-input-decorator',canGenerate: false, components: [
-				{name: 'titleInput', kind: Input, classes: 'moon-header-text moon-header-title'}
+                {name: 'titleInput', kind: Input, classes: 'moon-header-text moon-header-title',  onblur: 'inputBlur'},
```
aarontam: We generally use tabs for indentation, not spaces - you might need to change the settings in your editor.

lib/Header/Header.js
```diff
@@ -796,9 +797,31 @@ module.exports = kind(
...
+    },
+
+        /**
+         * Logic for replacing input into marquee
+         *
+         * @private
+         */
+
+        marqueToggle: function (){
+            if (this.$.titleInputMarquee.getContent() === '') return ;
```
aarontam: Let's use the generic getter here i.e.

```
if (this.$.titleInputMarquee.get('content') === '') return;
```

lib/Header/Header.js
```diff
@@ -796,9 +797,25 @@ module.exports = kind(
 	* @fires module:moonstone/Header~Header#onInputHeaderChange
 	* @private
 	*/
-	handleChange: function (inSender, inEvent) {
-		this.doInputHeaderChange({originalEvent: util.clone(inEvent, true)});
-	},
+    handleChange: function (inSender, inEvent) {
+        this.doInputHeaderChange({originalEvent: util.clone(inEvent, true)});
+        this.$.titleInputMarquee.set('content', inEvent.originator.getValue());
+        this.marqueToggle();
```
aarontam: This should be `marqueeToggle` :smile:

src/Header/Header.less
```diff
@@ -237,3 +237,7 @@
 		float: right;
 	}
 }
+
+.hidden {
```
clinuz: We can't have a global `.hidden` class defined here. This doesn't make sense.

lishammel:

PR #2339, opened 2015-07-24
ENYO-1906: Add Text to Speech Accessibility support to Popup and Cont…

jaewonjang:
…extual Popup

In moon.Popup and moon.ContextualPopup we need to prevent 'aria-live' attribute
because if popup has marquee text and it has focus, screen reader will read
all the content continuosly.
When popup is showing, set 'aria-hidden' false and when it disapeared set
'aria-hidden' true.
Set 'aria-live' to 'off' and apply 'aria-hidden'.

https://jira2.lgsvl.com/browse/ENYO-1906

### DIFF
--- a/lib/ContextualPopup/ContextualPopup.js
+++ b/lib/ContextualPopup/ContextualPopup.js
@@ -8,6 +8,7 @@ require('moonstone');
 var
 	kind = require('enyo/kind'),
 	dom = require('enyo/dom'),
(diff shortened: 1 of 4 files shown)


ryanjduffy:
With this PR, opening a popup will read all of its contents -- buttons, text, lists. Is that the expectation?

jaewonjang:
Yes, we expect to read all contents in popup.

PR #2340, opened 2015-07-24
ENYO-2168: Marquee doesn't recalculate distance when scroller show/hide

KunmyonChoi:
scroll thumb.

Issue:
- The moon.Scroller doesn't have a way to detect content changes.
- So, we need to bubble up onRequestSetupBonds to update scroll boundary
  and thumb state.
- The problem is even if moon.Scroller (moon.ScrollStrategy) update it's
  scroll bounds and enable scroll thumb area. All the items inside
  doesn't get any notice from scroller when scroll thumb area is
  reserved by strategy (push padding-right).

Fix:
- It need to call resize on scroller when scroll thumb state is changed,
  because the scroll content area and item width is actually changed.

### DIFF
--- a/lib/ScrollStrategy/ScrollStrategy.js
+++ b/lib/ScrollStrategy/ScrollStrategy.js
@@ -163,10 +163,6 @@ var MoonScrollStrategy = module.exports = kind(
 		{kind: Signals, onSpotlightModeChanged: 'spotlightModeChanged', isChrome: true}
 	],
 
-	observers: [
-		{method: 'resize', path: ['vEnabled', 'hEnabled']}
-	],
-
 	/**
 	* @private
 	*/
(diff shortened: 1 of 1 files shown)


lib/ScrollStrategy/ScrollStrategy.js
```diff
@@ -807,6 +811,7 @@ var MoonScrollStrategy = module.exports = kind(
 		this.$.hColumn.addRemoveClass('v-scroll-enabled', enabled);
 		this.$.pageUpControl.spotlight = enabled && this.spotlightPagingControls;
 		this.$.pageDownControl.spotlight = enabled && this.spotlightPagingControls;
+		this.set('vEnabled', enabled);
```
aarontam: Is there any reason we can't just call `resize` directly, instead of setting up observers?

Djspaceg: I thought this same thing, but after a search discovered that `ScrollControls.js:updateScrollability` uses this property.
Does seem like a heavy approach...

ryanjduffy: Notably, if both `vEnabled` and `hEnabled` change at the same time, `resize()` will be called twice. Might be worth a little extra code in `enableDisableScrollColumns` to ensure that doesn't happen.

webOS101:
I went ahead and reimplemented this without the observers and the calls to .set().  While I really liked the observer approach at first I did decide it hides the connection between cause and effect and it is not as performant as a more direct approach.  I suppose this is one of those situations that flux deals with well.

PR #2343, opened 2015-07-25
ENYO-2164: Increase delay before starting IntegerPicker scroll...

graynorton:
...to work around DOM sync issue on TV. With the previous delay,
the offsetTop value reported by the DOM API for our target node
was wrong under certain circumstances (issue could be reproduced
by repeatedly clicking to increment or decrement the picker; error
case occured roughtly once every 50-100 clicks).

I'd ultimately love to see us find a way to rework IntegerPicker
such that it didn't need to use this async trick for scrolling,
since async stuff like this is inherently brittle across different
environments; but for now, the best fix seems to be simply to
increase the delay.

After some experimentation, I settled on roughly doubling the delay
from 16 to 33 ms. I was unable to repro the error case when
clicking 500+ times at this value, and there's no perceptible
difference in responsiveness.

### DIFF
--- a/lib/IntegerPicker/IntegerPicker.js
+++ b/lib/IntegerPicker/IntegerPicker.js
@@ -580,7 +580,7 @@ module.exports = kind(
 			this.updateRepeater(index, count);
 
 			this.scrollToIndex(oldIndex, false);
-			this.startJob('valueChanged-Scroller', this.bindSafely('scrollToIndex', newIndex, true), 16);
+			this.startJob('valueChanged-Scroller', this.bindSafely('scrollToIndex', newIndex, true), 33);
 		} else {
 			// if old isn't specified, setup the repeater with only this.value and jump to it
 			this.updateRepeater(newIndex);


otisnate:
Hi @graynorton , did you test on the TV? I'm afraid that just changing timer delay could make another timing issue

kyo504:
@graynorton - I've tested your change on TimePicker sample. I couldn't see the problem but when I repeatedly clicks, a number sometimes changes without animation(scroll). I think it's a bit strange behavior. Have you seen this?

graynorton:
@otisnate, yes, this issue only happens on TV, so all of my testing was done there (on an m14+ board). As noted in my commit message, I would prefer a different sort of fix, but I believe that would require a significant rewrite of the control, which would be much too risky at this point in the program.

@kyo504, regarding the behavior you're describing (numbers sometimes changing without animation), my guess is that it has always been this way. A couple of questions:

* Does this happen only when clicking rapidly and repeatedly?

* Does it happen only when testing against this branch, or does it also happen when testing against the current 2.6.0-dev?

PR #2345, opened 2015-07-27
ENYO-2178-marquee distance is not properly updated when changeing con…

rajyavardhanp:
…tent during control is hidden

If content of the marquee is changed while control is hidden, marquee
distance is not calculated to apply proper classes.
So while changing the showing status, marquee distance is calculated.
Enyo-DCO-1.1-Signed-Off-By: Rajyavardhan P <rajyavardhan.p@lge.com>


lishammel:
After discussion with Rajyavardhan, closing out as jira associated with this has been resolved another way.

PR #2351, opened 2015-07-28
ENYO-2193: Stop marquee when tearing down as part of a cached view.

aarontam:
### Issue
When a `Marquee` control is part of a "cached" view that is torn down and re-rendered, the control does not resume marqueeing from the start i.e. as if it was first rendered.

### Fix
For the case where our `Marquee` control is being torn down as part of a cached view, we make a call to `stopMarquee` to effectively clean-up the state.

### DIFF
--- a/lib/Marquee/Marquee.js
+++ b/lib/Marquee/Marquee.js
@@ -339,6 +339,17 @@ var MarqueeSupport = {
 	}),
 
 	/**
+	* @method
+	* @private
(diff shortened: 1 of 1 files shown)


aarontam:
Didn't realize I had created this PR, but we have already merged what is effectively the same fix via https://github.com/enyojs/moonstone/pull/2380.

PR #2352, opened 2015-07-28
ENYO-1808: Converted all controls to use a standard font configuration.

Djspaceg:
* All controls have either been explicitly or inherently upgraded to use a more traditional font family/weight/style assignment method.
* Weights have been appropriately assigned to the fonts definitions file.
* All weighted-names have been replaced by normal family-only names.
* The base `.moon` and `.moon-body-text` classes have been switched to normal weight from bold so their children behave more traditionally.
* Several controls were given explicit weights to reflect the design's specifications, due to the change above.
* Moonstone Icons appearing inside of buttons needed a kerning rule reset, as the Moonstone Icons font does not support kerning. (Commented in the code)
* `.font-kerning` mixin has been updated to #1: accept alternate values, #2: not render as a class in the CSS files.


Djspaceg:
Closing in favor of #2582

PR #2354, opened 2015-07-29
ENYO-2111: ExpandablePicker: Grouped Picker Bounces on Initial Selection

iKuzko:
ENYO-2111: ExpandablePicker: Grouped Picker Bounces on Initial Selection

### DIFF
--- a/lib/ExpandableListItem/ExpandableListItem.js
+++ b/lib/ExpandableListItem/ExpandableListItem.js
@@ -165,6 +165,11 @@ module.exports = kind(
 	defaultKind: Item,
 
 	/**
+	 * @private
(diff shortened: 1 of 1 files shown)


aarontam:
Hi @iKuzko, with this change, if we collapse a picker that is near the bottom of the scroller, the scroller will not resize itself to reflect the shortened height of its contents.

iKuzko:
Hi @aarontam, i make change to commit https://github.com/enyojs/enyo/pull/1240/ to remove this bug

aarontam:
Hi @iKuzko, thanks for your update - please see my comment in the ticket for more details on the issue.

PR #2358, opened 2015-07-31
Enyo 2189 alischuk

alischuk:


aarontam:
Hi @alischuk, I think you have commits for another fix in this PR as well - might want to rebase this and/or fix the branch.

PR #2364, opened 2015-08-03
ENYO-2218: ListActions are stacked vertically, not horizontally

kyo504:
### Issue:
ListAction checks if actions are arranged vertically or horizontally by comparing total width. However it takes all the components even if their visibility is set to false so it returns incorrect result. Also it doesn't care about resolution independence

### Fix:
Calculate total width by taking only actions whose visibility is set to true and use ri.scale to get proper result based on resolution.

Enyo-DCO-1.1-Signed-Off-By: Hunkyo Jung (hunkyo.jung@lge.com)

### DIFF
--- a/lib/ListActions/ListActions.js
+++ b/lib/ListActions/ListActions.js
@@ -429,7 +429,6 @@ var ListActions = module.exports = kind(
 		}
 		this.disabledChanged();
 		this.listActionsChanged();
-		this.drawerNeedsResize = true;
 	},
 
 	/**
@@ -466,6 +465,7 @@ var ListActions = module.exports = kind(
 	listActionsChanged: function() {
 		var owner = this.hasOwnProperty('listActions') ? this.getInstanceOwner() : this;
 		this.listActions = this.listActions || [];
+		this.drawerNeedsResize = true;
 		this.renderListActionComponents(owner);
 	},
 
@@ -485,6 +485,8 @@ var ListActions = module.exports = kind(
 		var listAction, i;
 
 		this.listActionComponents = [];
+		this.$.listActions.destroyClientControls();
+
 		for (i = 0; (listAction = this.listActions[i]); i++) {
 			this.listActionComponents.push(this.createListActionComponent(listAction, owner));
 		}
@@ -555,9 +557,6 @@ var ListActions = module.exports = kind(
 	*/
 	beforeOpenDrawer: function(standardHeight, type) {
 		this.standardHeight = standardHeight;
-		if (type !== 'large') {
-			this.set('stacked', false);
-		}
 	},
 
 	//TODO: Remove the onListActionOpenChanged event. It will be deprecated in favor of the onShow/onHide events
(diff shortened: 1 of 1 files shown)


lib/ListActions/ListActions.js
```diff
@@ -648,8 +648,17 @@ var ListActions = module.exports = kind(
 	* @private
 	*/
 	shouldStack: function() {
+		var i = 0,
+		    nVisibleComps = 0;
+
+		for (i = 0; i<this.listActionComponents.length; i++) {
+			if (this.listActionComponents[i].getShowing()){
```
yeram: missing white-space before {

kyo504: Added a space between closing round bracket and open curly brace.

lib/ListActions/ListActions.js
```diff
@@ -648,8 +648,17 @@ var ListActions = module.exports = kind(
 	* @private
 	*/
 	shouldStack: function() {
+		var i = 0,
+		    nVisibleComps = 0;
+
+		for (i = 0; i<this.listActionComponents.length; i++) {
+			if (this.listActionComponents[i].getShowing()) {
+				nVisibleComps++;
```
otisnate: I think you can replace same routine as bellow
```
    optionGroup;
    for (i = 0; (optionGroup = this.listActionComponents[i]); i++) {
        if (optionGroup.showing) {
            nVisibleComps++;

```

kyo504: Thanks for your comment. I replaced conventional array iterator with assigning member in iterating condition block like you mentioned.

lib/ListActions/ListActions.js
```diff
@@ -677,6 +686,10 @@ var ListActions = module.exports = kind(
 	stackMeUp: function() {
 		var optionGroup, i;
 
+		if (this.standardHeight) {
```
aarontam: Should we abstract this out of `stackMeUp` and move this to `stackChanged`?

kyo504: @aarontam  Thanks for your comment. You're right. It's a sort of repetitive codes necessary to be moved to somewhere. Like you said, {{beforeOpenDrawer}} is the proper place so I updated PR.

lib/ListActions/ListActions.js
```diff
@@ -485,6 +485,8 @@ var ListActions = module.exports = kind(
 		var listAction, i;
 
 		this.listActionComponents = [];
+		this.$.listActions.destroyClientControls();
```
aarontam: I didn't get a chance to fix this during the rush to tag, but we could guard this by checking `this.generated`, so that we do not run this unnecessarily at create time. We can include this in a future PR.

kyo504: @aarontam - Thanks for your comment. I'll make a PR for this and it would be like below.
```javascript
if (this.generated) {
    this.$.listActions.destroyClientControls();
}
```
Do I need to create a jira ticket for this? Or do we just review and merge it?

aarontam: Hi @kyo504, let's create a JIRA ticket for this so we can more easily track it, in case we need to figure out why we made that change, in the future. We can link it to the original ENYO-2218 issue.

PR #2365, opened 2015-08-03
ENYO-2110: ExpandablePicker scroll thumb visible during collapse animation when navigating up to header.

KunmyonChoi:
Issue:
- When header (moon.Item) is focused, onRequestScrollIntoView is
  triggered by onSpotlightFocused.
- The onRequestScrollIntoView is fired while the drawer is animating
  closed, which triggers alertThumbs as the scroll viewport is smaller
  than the original content size.

Fix:
- Remove duplicated call of onRequestScrollIntoView from
  headerWrapper and moon.Item.
- Change call order of onRequestScrollIntoView and setActive.

### DIFF
--- a/source/ExpandablePicker.js
+++ b/source/ExpandablePicker.js
@@ -182,7 +182,7 @@
 		* @private
 		*/
 		components: [
-			{name: 'headerWrapper', kind: 'moon.Item', classes: 'moon-expandable-picker-header-wrapper', onSpotlightFocus: 'headerFocus', ontap: 'expandContract', components: [
+			{name: 'headerWrapper', kind: 'moon.Item', classes: 'moon-expandable-picker-header-wrapper', onSpotlightFocus: 'headerFocus', spotlightFocused: enyo.nop, ontap: 'expandContract', components: [
 				// headerContainer required to avoid bad scrollWidth returned in RTL for certain text widths (webkit bug)
 				{name: 'headerContainer', classes: 'moon-expandable-list-item-header moon-expandable-picker-header', components: [
 					{name: 'header', kind: 'moon.MarqueeText'}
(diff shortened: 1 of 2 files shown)


source/ExpandablePicker.js
```diff
@@ -182,7 +182,7 @@
 		* @private
 		*/
 		components: [
-			{name: 'headerWrapper', kind: 'moon.Item', classes: 'moon-expandable-picker-header-wrapper', onSpotlightFocus: 'headerFocus', ontap: 'expandContract', components: [
+			{name: 'headerWrapper', kind: 'moon.Item', classes: 'moon-expandable-picker-header-wrapper', onSpotlightFocus: 'headerFocus', spotlightFocused: enyo.nop, ontap: 'expandContract', components: [
```
aarontam: What is the purpose of `spotlightFocused: enyo.nop`?

aarontam: Never mind - was confusing this was an event handler. Still looks a bit odd.

source/ExpandablePicker.js
```diff
@@ -553,6 +553,25 @@
...
+		},
+
+		/**
+		* If drawer is currently open, and event was sent via keypress (i.e., it has a direction),
+		* process header focus.
+		*
+		* @fires moon.Scroller#onRequestScrollIntoView
+		* @private
+		*/
+		headerFocus: function (inSender, inEvent) {
```
aarontam: I don't like effectively duplicating the exact same logic from `moon.ExpandableListItem` here. As a potential alternative, we could do something like this: https://github.com/enyojs/moonstone/compare/ENYO-2110-kunmyonchoi-2.5-new...ENYO-2110-aarontam

aarontam:
As a general question, I am curious why we had to make changes to `moon.ExpandableListItem` if we are overriding all of these changes in `moon.ExpandablePicker`.

aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2385.

source/ExpandablePicker.js
KunmyonChoi: Hi @aarontam , sorry for the late response. Your solution looks clean and simple.

KunmyonChoi:
I modified ExpandableListItem but it was based on assumption. So, Yes. It may not be a real problem.

PR #2368, opened 2015-08-03
ENYO-2231: Slider popup in Contextual Popup has vertical blank

KunmyonChoi:
Issue:
- Slider popup has vertical noise line.

Fix:
- Make popup div round and set background color as the same as label
  color.

### DIFF
--- a/lib/ProgressBar/ProgressBar.less
+++ b/lib/ProgressBar/ProgressBar.less
@@ -26,6 +26,7 @@
 /* Popup */
 .moon-progress-bar-popup {
 	white-space: nowrap;
+	border-radius: 72px;
 	z-index: 20;
 	left: 50%;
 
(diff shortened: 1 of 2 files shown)


lib/ProgressBar/ProgressBar.less
```diff
@@ -26,6 +26,7 @@
 /* Popup */
 .moon-progress-bar-popup {
 	white-space: nowrap;
+	border-radius: 72px;
```
aarontam: Do you know how you came up with this specific number, vs. using a value similar to what we're using for Button?

KunmyonChoi: This number comes from maximum height of popup in source code.

aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2378.

PR #2369, opened 2015-08-03
ENYO-2199: Restored the custom JP header sizing

Djspaceg:
… and rearranged the rules for ideal precedence.

Also (finally) added a `.moon-large-header` class to remove many unnecessary rules that just get overridden.

### DIFF
--- a/lib/Header/Header.js
+++ b/lib/Header/Header.js
@@ -616,6 +616,7 @@ module.exports = kind(
 	* @private
 	*/
 	typeChanged: function () {
+		this.addRemoveClass('moon-large-header', this.get('type') == 'large');
 		this.addRemoveClass('moon-medium-header', this.get('type') == 'medium');
 		this.addRemoveClass('moon-small-header', this.get('type') == 'small');
 		this.contentChanged();
--- a/lib/Header/Header.less
+++ b/lib/Header/Header.less
@@ -1,7 +1,6 @@
 .moon-header {
 	.border-box;
 	color: @moon-header-text-color;
-	height: @moon-header-height-large;
 	border-top: @moon-header-border-top-width solid @moon-header-border-color;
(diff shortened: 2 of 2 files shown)


lib/Header/Header.less
```diff
@@ -179,35 +194,13 @@
...
-	&.enyo-locale-te, // Telugu
-	&.enyo-locale-kn, // Kannada
-	&.enyo-locale-ml, // Malayalam
-	&.enyo-locale-mr, // Marathi
-	&.enyo-locale-bn, // Bengali
-	&.enyo-locale-pa {// Panjabi
-		.moon-header .moon-header-title {
-			height: 159px;
-		}
-	}
```
Djspaceg: We've controlled the title's height using line-height instead, and line-height is actually taller than 159px. This is no longer necessary.

lib/Header/Header.js
```diff
@@ -616,6 +616,7 @@ module.exports = kind(
 	* @private
 	*/
 	typeChanged: function () {
+		this.addRemoveClass('moon-large-header', this.get('type') == 'large');
```
aarontam: This is probably a micro-optimization, but could we do something like this:
```
if (was) this.removeClass('moon-' + was + '-header');
if (is) this.addClass('moon-' + is + '-header');
```
Not sure if that reduces readability significantly, probably saves us a few comparisons, just a thought now that we have a class for each of the header types.

Djspaceg: What should line #339 look like with this change?

aarontam: Ah didn't realize we're calling that at create time, let's leave it as-is rather than changing that to `this.typeChanged(null, this.type);`. :+1:

Djspaceg: Yusssss

PR #2371, opened 2015-08-04
ENYO-2128 : Fix reading text twice on marquee item

kbs12e:
## Issue

If the string of the SelectableItem(or RadioItem) has a long text truncation, screen reader reads string twice.

## Cause

Screen reader reads spotlight focused text when webOS TV's audio guidance is on and component is focused with 5-way mode. However, if component internally has marquee mixin for reading long text like moonstone SelectableItem, this issue is reproduced. 
Screen reader tries to read dom content but marquee mixin creates marqueeWrapper component dynamically as soon as component is focused, so screen reader first reads content and text of marquee that is created after focusing.

## Fix

add aria-hidden:true when marquee is started and remove it when component is blured.


kbs12e:
This is not neccessary. In latest code, this issue is not reprodeced, so close this PR.

PR #2372, opened 2015-08-04
ENYO-2021: Add Text to Speech Accessibility support to ExpandableText

kbs12e:
Initial add accessibility feature to ExpandableText

## Requirement

WebOS TV should read all content of ExpandableText when more or less button is focused though content is collapsed.

## Implementation

We added ExpandableTextAccessibilitySupport to support reading a ExpandableText.
ExpandableText has client and button component. The client component has content and button is more or less button. Screen reader should not read more or less button content when this button is focused. Instead, it should read all content of client component, To implement this requirement, we add aria-labelledby attribute, which has client component id to more or less button. 


kbs12e:

PR #2379, opened 2015-08-05
ENYO-2244: Stop active marquee on teardownRender

KunmyonChoi:
## Issue:
- When index changed, lightPanels does teardownRender not destroy on
  popOnBack/popOnForward.
- When it is teardownRender, marquee doesn't stop marquee but stop flow
  and stay its position.

## Fix:
- We stop marquee on any active marquee teardownRender.


KunmyonChoi:
Create new clean PR.

PR #2383, opened 2015-08-06
ENYO-2152: Accessibility: screen reader does not read accessibilityHi…

jaewonjang:
…nt of component that has aria-labelledby

If both 'aria-label' and 'aria-labelledby' are in a dom node, screen reader will
read only 'aria-labelledby'. When developer uses only accessibilityHint he expects to
read <content> + <accessibilityHint>, but if there is 'aria-labelledby' screen reader
will read just 'aria-labelledby'.
Blink didn't support reading children node when parent is focused without
'aria-labelledby', but now it is changed on PLAT-426 and we don't need to
use 'aria-labelledby' anymore.

https://jira2.lgsvl.com/browse/ENYO-2152


jaewonjang:
This PR has dependency on https://github.com/enyojs/moonstone/pull/2332

jaewonjang:
created new PR https://github.com/enyojs/moonstone/pull/2460

PR #2386, opened 2015-08-06
ENYO-2116: Ensure that we avoid using the currentText element when computing scroll bounds when drawer is opened/closed.

aarontam:
Changes from https://github.com/enyojs/moonstone/pull/2385.


aarontam:
Closing as we have updated scrolling in 2.6 that is not susceptible to the original issue (merged for 2.5).

PR #2389, opened 2015-08-07
ENYO-1905: Add Text to Speech Accessibility support to Spinner

kbs12e:
Initial add accessibility feature to Spinner

## Requirement
Screen reader should read Spinner Loading text when Spinner is loading according to UX requirements.

## Implementation
We added SpinnerAccessibilitySupport to support reading a spinner's content.
Spinner has animating control and client content, so sreen reader should read content string when spinner is shown. However, Spinner is not spotlight component, so we should use accessibilityAlert API for reading it. If aceessibilityAlert is true, screen reader reads contents whenever it is changed.
Because Spinner' client control is marqeeeText, The scrren reader reads content whenever marquee animataion is started. To prevent this, we added aria-live: off attribute to spinner.

## Limitation
In ChromeVox, this commit doesn't work. it works only WebOS TV.


seunghoh:
Looks good to me

kbs12e:
I wll re-create PR with latest code.

PR #2390, opened 2015-08-07
ENYO-1751: Accessibility: It read value with html tag in the RichText…

jaewonjang:
… control.

In case RichText, html tag can be included on it.
If we get the value using getValue() it will return the value including
html tag and screen reader will read it.
For RichText, set the label using innerText to remove html tag.

https://jira2.lgsvl.com/browse/ENYO-1751


jaewonjang:
replaced with https://github.com/enyojs/moonstone/pull/2491

PR #2392, opened 2015-08-07
Wp 774 TimePicker support Ethiopia 

DavidUM:
Issue
-------
Make timePicker support Ethiopia time

Fix
----
Current timePicker does not consider that meridiem has some entries more than two -AM & PM-
Fix related logic

### DIFF
--- a/source/TimePicker.js
+++ b/source/TimePicker.js
@@ -39,12 +39,12 @@
 		/**
 		* @private
 		*/
-		value: null,
+		wrap: true,
 
 		/**
 		* @private
 		*/
-		wrap: true,
+		locale: null,
 
 		/**
 		* @private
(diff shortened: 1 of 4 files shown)


source/TimePicker.js
```diff
@@ -75,13 +122,13 @@
 		* @private
 		*/
 		setupItem: function (inSender, inEvent) {
-			var index = inEvent.index % this.range;
-			this.$.item.setContent(this.meridiems[index]);
+			var index = inEvent.index % this.range || 0;
```
aarontam: Do you remember why you had to make this change? Will `this.range` ever be a non-zero falsy value?

source/TimePicker.js
```diff
@@ -554,10 +594,10 @@
 		hourPickerChanged: function (sender, event) {
 			if(this.syncingPickers) return true;
 
-			var hour = event.value;
+			var hour = this.value.getHours() + (event.originator.get('offset'));
```
aarontam: Wonder if we should document the `offset` property of `HourPicker` and/ or just use direct property access i.e. `event.originator.offset`.

source/TimePicker.js
```diff
@@ -585,14 +625,33 @@
...
 			if(this.syncingPickers) return true;
 
-			var hour = this.$.hour.get('value'),
-				value = event.value;
+			var meridiemPicker = event.originator,
+				meridiems = meridiemPicker.get('meridiems'),
+				meridiem = meridiems[meridiemPicker.get('value')],
+				start = meridiem.start.split(':'),
+				startHour = parseInt(start[0], 10),
+				offset = meridiemPicker.get('offset'),
```
aarontam: Similar to above, we should consider documenting the `offset` property of `MeridiemPicker` and/ or just use direct property access i.e. `meridiemPicker.offset`.

source/TimePicker.js
```diff
@@ -585,14 +625,33 @@
 		meridiemPickerChanged: function (sender, event) {
 			if(this.syncingPickers) return true;
 
-			var hour = this.$.hour.get('value'),
-				value = event.value;
+			var meridiemPicker = event.originator,
+				meridiems = meridiemPicker.get('meridiems'),
+				meridiem = meridiems[meridiemPicker.get('value')],
```
aarontam: Rather than do this, it might make more sense to expose a method on `MeridiemPicker` that returns the current meridiem, or returns a meridiem based on a given value, where this value defaults to the current value.

source/TimePicker.js
DavidUM: Ya, I remember that it could be null at starting time but can't remember exact situation.
Let me find history.

PR #2397, opened 2015-08-10
ENYO-2259: ExpandablePicker have no margin at bottom when there is no current value.

KunmyonChoi:
Issue:
- ExpandablePicker current value have margin-top as 12px by default.

Fix:
- We add conditional margin-top only enabled when there is content.


KunmyonChoi:
Close this PR per other solution is fixing this issue. https://jira2.lgsvl.com/browse/PLAT-11101

PR #2399, opened 2015-08-11
ENYO-2273-Scroller abnormal behavior on holding paging controls

rajyavardhanp:
Issue: Scroll bounds are calculated before rendering the content.
'bottomBoundary' property is dependent on scroll bounds, which is used
for comparison onpaginate scrollTo, methods . It's not updated during or
before the hold, if we use only 5-way to move on paging controls.
Fix: In scroller 'onSpotlightContainerEnter'  is handled  to
update the scrollerbounds in 5-way


graynorton:
Closing in favor of https://github.com/enyojs/moonstone/pull/2449

PR #2409, opened 2015-08-13
ENYO-2284: spotlight jumps to unexpected position on auto collapse

kyo504:
### Issue

When drawer is automatically collapsed by 5-way key('UP') drawer's closing animation starts before finding nearest neighbor to give spot. At that time, spotlight finds incorrect neighbor so it gives spot to unexpected control.

### Fix

Due to regression, I reverted the previous change. And to fix ENYO-2131 I added some conditions to close drawer when spot moves from current picker to others by pressing repetitively 5-way 'UP' keys in a very short time.

Enyo-DCO-1.1-Signed-off-by: Hunkyo Jung hunkyo.jung@lge.com

### DIFF
--- a/lib/ExpandableListItem/ExpandableListItem.js
+++ b/lib/ExpandableListItem/ExpandableListItem.js
@@ -179,7 +179,6 @@ module.exports = kind(
 	* @private
 	*/
 	handlers: {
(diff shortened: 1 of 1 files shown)


kyo504:
This has been integrated into ENYO-2144

PR #2410, opened 2015-08-13
Enyo-2123 Prevent toggling selected property

DavidUM:
when we tap same button sequentialy.


lib/Slider/Slider.js
```diff
@@ -898,6 +898,7 @@ module.exports = kind(
 	*/
 	previous: function () {
 		this.set('value', this.value - this._jumpIncrementAmount);
+		if (this.popup && this.$.popup.getShowing()) return true;
```
aarontam: Hi @DavidUM, thanks for your update. I wasn't clear - is there a specific reason we want to allow the `onSpotlightSelect` event to bubble up when pressing one of the jump increment buttons, as opposed to always returning true here?

DavidUM: Hi @aarontam 
Because we'd like to execute `spotSelect()`.
It shows popup when `onSpotlightSelect` event triggered.
If we return true every time, we cannot show popup using `onSpotlightSelect` event

aarontam: Ah ha, I was unclear on the UX and whether we wanted the popup to display at all when pressing the jump increment buttons. If that is the case, should the popup be displayed when you 5-way hold on one of the jump increment buttons? Currently with this change, the popup only displays after the 5-way key is released.

DavidUM: Holding matter makes sense. 
I think to keep consistency, behavior of pointer mode need to be same as 5-way mode.

Let ask it to Robert.G and @Djspaceg

aarontam: Hi @DavidUM, just spoke with @Djspaceg and Robert - let's make it consistent and display the popup when using the pointer to press the jump increment buttons.

DavidUM: Hi @aarontam 
Currently popup is hided when we release mouse pointer pressing.
Do you mean popup should not hide?

aarontam: The popup does not display when we press and hold with the pointer on the jump increment buttons - is this what you are seeing?

DavidUM:
Hi @aarontam 
In chmod 644, I cannot write and build.
Does it really matter?

aarontam:
Yeah -- we don't want the permission set of our files being inconsistent and haphazard, and these changes shouldn't really belong in commits, as they are unintentional. We should really figure out what's going on with your environment, but in the meantime, you can have git ignore these changes so they are not included as part of the change diff, via this config command:
```
git config core.fileMode false
```

lib/Slider/Slider.js
```diff
@@ -269,19 +269,14 @@ module.exports = kind(
 	* @private
 	*/
 	jumpWrapperComponents: [
-		{name: 'buttonLeft', kind: IconButton, backgroundOpacity: 'transparent', classes: 'moon-slider-button left', icon: 'arrowlargeleft', onSpotlightKeyDown: 'configureSpotlightHoldPulse', onSpotlightSelect: 'previous', ondown: 'downLeft', onholdpulse: 'holdLeft', ondragstart: 'preventDrag', defaultSpotlightDisappear: 'buttonRight'},
+		{name: 'buttonLeft', kind: IconButton, backgroundOpacity: 'transparent', classes: 'moon-slider-button left', icon: 'arrowlargeleft', onSpotlightKeyDown: 'showKnobStatus', onSpotlightKeyUp: 'hideKnobStatus', onSpotlightSelect: 'previous', ondown: 'downLeft', onup: 'hideKnobStatus', onholdpulse: 'holdLeft', ondragstart: 'preventDrag', defaultSpotlightDisappear: 'buttonRight'},
```
aarontam: By specifying an `onSpotlightKeyDown` handler, any key press will cause the handler to execute, not just the enter key. Also, is there a way we can reduce the number of handled events; it is a very large set and I have a feeling we can consolidate some of the logic i.e. making calls to `showKnobStatus` in `next` and `previous` and getting rid of the `ondown` handlers, to start.

DavidUM: Oh, I didn't catch the first feedback.
And, reducing event handlers is what I'd like to do.
I also moved {{showKnobStatus}} in {{next}} and {{previous}} but it could not fulfill new requirement.
You said, when okay key is released, popup should be closed.
But unfortunately, {{onSpotlightSelect}} event which call {{next}} and {{previous}} comes
after {{onSpotlightKeyUp}} event.
So we cannot match new requirement.

I'm still considering how can I match new requirement with reducing handlers

aarontam: Hi @DavidUM, I was referring to the fact that pressing down on any key would trigger `showKnobStatus`, regardless of what key it is, because it is assigned to be the handler for `onSpotlightKeyDown` events. I realize we need to handle all of these events - I tried to consolidate the handlers as much as possible so that we only have one handler (`jumpButtonTriggered`) for "start" events, and then we use `hideKnobStatus` for the "end" events. Let me know what you think: https://github.com/enyojs/moonstone/compare/2.6.0-dev...ENYO-2123-aarontam.

DavidUM: Hi @aarontam 
Sorry, I expressed 'didn't catch'. That's my mistake.
I'd like to say "I understand. I didn't know it" haha.

Second thing is what I really mis-understood.
I thought you'd like to reduce number of handlers and events too!
I read you branch and it is better to understand.
Thank you.

Let me update my PR based on your advise.

aarontam: Ah ha I see. :smile: I think I needed to tweak one thing with my branch - with the way it is implemented, there would be no need to `return true` from the `next` and `previous` methods. Let me know how it works out for you.

DavidUM: Hi @aarontam, I think we do not need to secure `holdConfig` for iconButton.
Every other control which had it has removed this feature in these days.
Is there any specific reason to keep `holdConfig`?

DavidUM: And, as you expected, we do not need `return true` for `previous and `next`

aarontam: Hi @DavidUM, yes that seems to have been legacy code, before we applied a global holdpulse configuration to Moonstone (see https://github.com/enyojs/moonstone/blob/2.6.0-dev/index.js#L13). I (or the reviewer) can double-check this to make sure it is no longer needed.

DavidUM: Oh, I didn't expect that you reply at this moment :+1: 
I'm so happy if you check it. Thank you!

src/Slider/Slider.js
```diff
@@ -284,18 +284,42 @@ module.exports = kind(
 	* @private
 	*/
 	jumpWrapperComponents: [
-		{name: 'buttonLeft', kind: IconButton, backgroundOpacity: 'transparent', classes: 'moon-slider-button left', icon: 'arrowlargeleft', onSpotlightKeyDown: 'configureSpotlightHoldPulse', onSpotlightSelect: 'previous', ondown: 'downLeft', onholdpulse: 'holdLeft', ondragstart: 'preventDrag', defaultSpotlightDisappear: 'buttonRight'},
-		{name: 'slider', classes: 'moon-slider', spotlight: true, accessibilityLive: 'polite'},
```
clinuz: Hi @DavidUM, it looks like you removed the `accessibilityLive: 'polite'` property from your current rewrite below, was that intentional?

src/Slider/Slider.js
```diff
@@ -838,58 +862,25 @@ module.exports = kind(
...
-	/**
-	* @private
-	*/
-	configureSpotlightHoldPulse: function (sender, e) {
-		if (e.keyCode === 13) {
-			e.configureHoldPulse(this.holdConfig);
+	jumpButtonTriggered: function (sender, ev) {
+		var isValidEvent = true;
+		if (!sender.disabled) {
+			if (ev.keyCode != 13 && (ev.type == 'onSpotlightKeyDown' || ev.type == 'ondown')) {
```
clinuz: @DavidUM, when testing this, I only ever see 3 types of events, `onSpotlightKeyDown`, `down` (from mouse), and `onholdpulse`. Do we need to be handling the `ondown` event at all or event mapping it from the components to this handler?

src/Slider/Slider.js
```diff
@@ -284,18 +284,42 @@ module.exports = kind(
...
+		{
+			name: 'buttonLeft',
+			kind: IconButton,
+			backgroundOpacity: 'transparent',
+			classes: 'moon-slider-button left',
+			icon: 'arrowlargeleft',
+			onSpotlightSelect: 'preventEvent',
+			onSpotlightKeyDown: 'jumpButtonTriggered',
+			onSpotlightKeyUp: 'hideKnobStatus',
+			ondown: 'jumpButtonTriggered',
```
clinuz: See my comment below.

src/Slider/Slider.js
DavidUM: Hi @clinuz 
No, i think it is my mistake when I fix merge conflict.
Thanks :+1:

src/Slider/Slider.js
DavidUM: Thanks @clinuz I will test again and update PR.

src/Slider/Slider.js
```diff
@@ -838,58 +862,25 @@ module.exports = kind(
...
-	/**
-	* @private
-	*/
-	configureSpotlightHoldPulse: function (sender, e) {
-		if (e.keyCode === 13) {
-			e.configureHoldPulse(this.holdConfig);
+	jumpButtonTriggered: function (sender, ev) {
+		var isValidEvent = false;
+		if (!sender.disabled) {
+			if ((ev.type == 'down' || ev.type == 'holdpulse') || (ev.keyCode == 13 && ev.type == 'onSpotlightKeyDown')) {
```
aarontam: Hi @DavidUM, I think the alternative I proposed was more exploratory and a proof-of-concept, and not a "this is what it should be" type of thing. The original intention was to attempt to consolidate the event handlers that were all performing the same action with very little variation (three handlers per event, multiplied by two controls per event, resulted in six handlers). The only case where the behavior had to be different was the `onSpotlightKeyDown` case, where the action should be triggered only if we had keyCode 13 - this is why we originally assume the event is valid by default, and nullify this if there is a keyCode that is not 13 for the `onSpotlightKeyDown` case. The current change kind of defeats the purpose of a general handler. I wonder if it would be better to give each of these events its own handler, which would call this generic `jumpButtonTriggered` method. In the case of `onSpotlightKeyDown`, we could check the keyCode first before making this call. Thoughts?

DavidUM: Hi @aarontam Thank you for your comment.
I understand what was your intention. I totally agree with that.
Yes, we can use `down` instead of `ondown` with your original code.

I have a class now so will update it after 9 hours. Thank :+1:

aarontam: Hi @DavidUM, no problem! I think that we do not actually need to check for `down` at all - we are only concerned about the `onSpotlightKeyDown` case where the key code is *not* 13. Let me know if there is a case that I am not thinking of. Thanks!

DavidUM: Oh, we should consider `holding` case.

clinuz:
Closed in favor of https://github.com/enyojs/moonstone/pull/2525 (squashed version of these commits and merges).

PR #2411, opened 2015-08-13
ENYO-2243: Fix reading input value twice on ExpandableInput

kbs12e:
## Issue 
When ExpandableInput is expanded, screen reader reads the value in the input twice

## Cause
When ExpandableInput is expanded, input is focused(webkit focus) and this focus() event is bubble up.
Because InputDecorator has onfocus handler, which calls Spotlight.spot, InputDecorator is spotlight focused after input is focused.
In spotLightFocusedHandler of inputDecorator, it is implemented that reading placeholder and 'editBox' string, so this issue is happend.

## Fix
Set accessibilityLabel to null for preventing reading a placeholder and 'editBox' string.


kbs12e:
accessibility source is refactoring, so this PR needs more changes. Thus , I will close this PR and create new PR

PR #2413, opened 2015-08-17
ENYO-1994 : Add Text to Speech Accessibility support to LightPanels

kbs12e:
Initial add accessibility feature to LightPanels.

## Requirement
WebOS TV should read 'new page' string when LightPanel is opened or moved according to UX requirements.

## Implementation
We added LightPanelsAccessibilitySupport to support reading 'new page' string.
LightPanel is spotlight container, so first child component or last spotlight component is focused when panel is opened.
Then, Screen reader first reads this focused component content. However, UX requirements request that Screen reader first should read 'new page' string when panel is opened or moved. 
TV should read 'new page' string as quickly as possible before reading focused component, so we added readAlert function to applyTransitions(). The applyTransitions is called when panel is opened or moved.

## Limitation
We used readAlert() of enyo-webOS library and overrided applyTransitions to read 'new page' string as quickly as possible, but If panel is opened without animation, 'new page' sound will cut for reading focused child component of panel 

### DIFF
--- a/lib/LightPanels/LightPanels.js
+++ b/lib/LightPanels/LightPanels.js
@@ -5,10 +5,12 @@
 
 var
 	kind = require('enyo/kind'),
-	LightPanels = require('enyo/LightPanels');
+	LightPanels = require('enyo/LightPanels'),
+	options = require('enyo/options');
 
 var
-	LightPanel = require('./LightPanel');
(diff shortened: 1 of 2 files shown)


kbs12e:
accessibility source is refactoring, so this PR needs more changes. Thus , I will close this PR and create new PR

PR #2418, opened 2015-08-18
ENYO-2330: Scroll thumb displaced vertically

JayCanuck:

### DIFF
--- a/css/moonstone-rules.less
+++ b/css/moonstone-rules.less
@@ -62,8 +62,6 @@ html {
 }
 
 .matrix3dsurface {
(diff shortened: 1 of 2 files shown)


css/moonstone-rules.less
```diff
@@ -62,8 +62,6 @@ html {
 }
 
 .matrix3dsurface {
-    -webkit-transform-origin: center center;
```
dmikey: This seems like a weird change to me? Are we sure this will not have impact else where? Maybe a solid commit message for this would be beneficial too.

PR #2423, opened 2015-08-19
ENYO-2315: JumpIncrement button status in slider is not updated

kyo504:
### Issue

When enableJumpIncrement is set to true, slider has left and right buttons to move towards left or right based on jumpIncrementAmount. However if we use MRCU, the status of the two buttons are not updated.

### Fix

While dragging the knob, we check the status of jumpIncrement buttons based on the current value.

Enyo-DCO-1.1-Signed-Off-By: Hunkyo Jung (hunkyo.jung@lge.com)

### DIFF
--- a/lib/Slider/Slider.js
+++ b/lib/Slider/Slider.js
@@ -410,8 +410,7 @@ module.exports = kind(
 	*/
 	disabledChanged: function () {
 		if (this.enableJumpIncrement) {
-			this.$.buttonLeft.set('disabled', this.disabled);
-			this.$.buttonRight.set('disabled', this.disabled);
+			this.updateButtonStatus();
 			this.$.slider.set('disabled', this.disabled);
 		}
 		this.addRemoveClass('disabled', this.disabled);
@@ -535,10 +534,7 @@ module.exports = kind(
 
 		this.value = v;
 		this.updatePopup(v);
-
-		if (this.enableJumpIncrement) {
-			this.updateButtonStatus(this.value);
-		}
+		this.updateButtonStatus();
 
 		if (this.lockBar) {
 			this.setProgress(this.value);
@@ -557,9 +553,11 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	updateButtonStatus: function (val) {
-		this.$.buttonLeft.set('disabled', this.disabled || val == this.min);
-		this.$.buttonRight.set('disabled', this.disabled || val == this.max);
+	updateButtonStatus: function () {
+		if (this.enableJumpIncrement) {
+			this.$.buttonLeft.set('disabled', this.disabled || this.value == this.min);
+			this.$.buttonRight.set('disabled', this.disabled || this.value == this.max);
+		}
 	},
 
 	/**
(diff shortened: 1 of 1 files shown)


lib/Slider/Slider.js
```diff
@@ -579,6 +579,11 @@ module.exports = kind(
 			this.updatePopup(this.elasticFrom);
 			this.set('value',this.elasticFrom);
 
+			if (this.enableJumpIncrement) {
```
otisnate: Please extract bellowing code as another function like ```updatePopup``` function
```
if (this.enableJumpIncrement) {
	this.$.buttonLeft.set('disabled', this.disabled || v == this.min);
	this.$.buttonRight.set('disabled', this.disabled || v == this.max);
}
```

lib/Slider/Slider.js
```diff
@@ -609,6 +614,12 @@ module.exports = kind(
 		this.dragging = false;
 		Spotlight.unfreeze();
 		this.set('value',v);
+
+		if (this.enableJumpIncrement) {
+			this.$.buttonLeft.set('disabled', this.disabled || v == this.min);
+			this.$.buttonRight.set('disabled', this.disabled || v == this.max);
+		}
+
```
otisnate: I think you can remove this routines

lib/Slider/Slider.js
kyo504: Thanks for the comment. I've wrapped redundant codes in a named function.

lib/Slider/Slider.js
kyo504: Thanks for your feedback. I've tested again and it's unnecessary so I removed it.

lib/Slider/Slider.js
```diff
@@ -525,6 +524,14 @@ module.exports = kind(
 	/**
 	* @private
 	*/
+	updateButtonStatus: function(val) {
```
aarontam: Minor: let's add a space between `function` and `(` here.

kyo504: Oops, you're right. Thanks for the feedback. Fixed.

lib/Slider/Slider.js
```diff
@@ -579,6 +586,10 @@ module.exports = kind(
 			this.updatePopup(this.elasticFrom);
 			this.set('value',this.elasticFrom);
 
+			if (this.enableJumpIncrement) {
+				this.updateButtonStatus(v);
```
aarontam: I was curious about whether we should only update the button status when dragging has finished, as opposed to checking/updating while dragging. I spoke with @Djspaceg and Robert about the UX here and they are indifferent to the behavior (so as long as the buttons are updated after dragging has finished). As a result, I think we should go with the more performant path and only update the button status when dragging has finished.

otisnate: Hi @aarontam, it's good feedback considering performance. But before moving this to ```dragfinish```, I want to get UX team's feedback if there is no issue although position is min/max and user is still draging (not dragfinish)

kyo504: I also thought of performance issue about updating status while dragging. But I didn't know which is better way. If updating the status when drag is finished, we can get better performance but the status is not changed while dragging. I think we can go with your suggestion because Robert doesn't have objection about this. Right?

aarontam: Yes that's right. As additional color on the UX thinking, the user cannot tap on any of the jump increment buttons while dragging, so we are never in a scenario where it is a necessity to have these buttons disabled while dragging.

ryanjduffy:
We also need to use this new method in `disabledChanged` as it appears that a disabled slider at either extreme would have both buttons enabled if the slider was enabled. This PR will also likely be stomped by Blake's vertical slider PR so I can make the change and reconcile the merge conflict at the same time.

PR #2424, opened 2015-08-19
ENYO-1958: Implemented vertical slider

Djspaceg:
This change alters the way that value is treated. Un-unit-ed values are treated as slider units, while values which explicitly mention percentages are treated as percentages: `50%` vs `50` (units/ticks)


KunmyonChoi:
@Djspaceg , we are not able to alter the way that value is treated. Because, it will needs changes for every app. I'm suggesting that introduce separate property of 'unit' beside of value. And set default behavior as the same as previous.

Djspaceg:
1. Not "every app" uses a slider
2. Of the apps that use a slider, not all of them use the new Increment Buttons
3. If an app that uses the new Increment Buttons, has a min of 0 and a max of 100, no change is necessary (since a value of 1 will have the same effect as a % of 1)
4. There are 1, maybe 2 apps that would need to change maybe 3 lines their code. A simple multi-file search would easily identify all cases: search for "enableJumpIncrement"
5. The jump increment value doesn't even need to change much: a `5` simply becomes `'5%'`. This could even be done by a simple regular expression: `s/(jumpIncrement\s*:\s*)(\d+)(\s*,)/$1'$2%'$3/`
6. The jumpIncrement feature is still very new and this is a change that completes the original feature-set of the control. It was always supposed to work this way but was overlooked during initial implementation.

_Sometimes apps need to change a few characters in their code to become better and allow everyone else to benefit; they need to be able to do that. This is not a breaking change. If the apps simply refuse to change, the control will still work, it will just work slightly differently._

aarontam:
Hi @KunmyonChoi, we believe that only the Settings app is utilizing the `jumpIncrement` feature. I can make the necessary change for them and submit a change request. Are there other apps utilizing this feature?

ryanjduffy:
* There are quite a few `get()` calls for properties of Slider and ProgressBar. Suggest that we use direct access (`this.orientation`) rather than `get('orientation')` to skip the unnecessary function call.
* Should `iconIncreaseButton` and `iconDecreaseButton` support images as well as icons? We could take a similar approach to moonstone/Overlay and support both on a single property to minimize the API surface area. On that topic, the property names are a bit awkward. I'd have expected `increaseButtonIcon` or perhaps `incrementIcon`.
* Regarding the `jumpIncrement` change, while it might not be a "breaking" change, it is a change in the public API. Technically, we are free to do that because it was added in pre.7 and we're on pre.15 so we're still in the same pre-release cycle. In reality, we have consumers treating our pre-releases as actual releases so what we can *technically* do is kinda irrelevant. What was the driver of the change in this PR?

Djspaceg:
1. We need to have this `get()` discussion offline and decide on an official policy. This matter should not hold-up this PR as this is a general discussion not directly related to only these controls and set of changes.
2. While I appreciate the idea of having both icon and images supported, it would mean adding an additional pair of properties to slider: `srcIncreaseButton` etc. `moon.Icon` uses two different properties to achieve support for both, so it doesn't have to differentiate between an icon name, a path and HTML/text. `iconIncreaseButton` seemed like the most descriptive and least overloaded-english at the time. `increaseButtonIcon` sounds like a method name and `incrementIcon` while short, also sounds like a method, i.e. some operation we're doing to "Icon". The counter to "increment" would be "decrement" which sounds funky too. I'm open to changing it, but I'd want an ideal name before we do.
3. The driver is that JumpIncrement supporting multiple units (direct and percent) was part of the original design of the control. It wasn't implemented initially. @aarontam has already created a patch for the one app that uses the feature. This is a special case –not normal procedure– in that it was supposed to be implemented before the control was merged, but nobody noticed, so now it needs to be updated to work as intended.

otisnate:
Hi all, although enyo 2.6 has ```pre.x``` tag name, we are already under QE phase (after FC phase). If ```pre.x``` name could make misunderstanding, I suggest to change name as ```rc.x```

ryanjduffy:
@otisnate - I suggested the same. I didn't mean to imply I think we're free to make public API changes, only that according to SemVer, we could. In this case, if we're confident that the impact is manageable, I don't object to taking the change.

@Djspaceg 
1. Okay. We are inconsistent in this regard so some further discussion is reasonable.
2. If we don't have requirements to support `src` and you feel adding it would be too cumbersome, I'm good with that. Regarding the name, I don't agree with your conclusion. `iconIncreaseButton` is inconsistent with similar pass-through properties in the framework in which the 'thing' precedes the property (e.g. `headerType` in moonstone-extra/Panel, `scrimClass` in moonstone/Popup, and several in moonstone/ProgressBar -- `barClasses`, `bgBarClasses`, `popupColor`, etc). I agree that since the 'thing' is the 'increaseButton', any property that starts with that will sound method-y. That said, I don't know what `increaseButtonIcon()` or `incrementIcon()` would do if such a thing existed so I don't know that users would make the jump that they're methods rather than properties.
3. I couldn't find any documentation in the original issue (ENYO-1144) that states that it should support both absolute and relative units for `jumpIncrement`. That very well may have been the intention but it doesn't appear to be documented. Whether we intended to implement it like this or not, it's still a public API change and I think @otisnate's and @KunmyonChoi's concerns are valid. As I said above though, if we are confident that the change can be absorbed by the apps and they're on board, we can take the change.

graynorton:
@ryanjduffy, @Djspaceg, I agree with Ryan on the naming in this case. I understand the reasoning on both sides, but I find `iconDecreaseButton` really hard to parse, and I think it's because I expect the word at the end to be the primary object. In this case we are specifying an icon, so I favor something like `incrementIcon`.

@ryanjduffy, @Djspaceg, @otisnate, @KunmyonChoi, @aarontam, I agree with Blake on the API change. In an ideal world, we would never change an API once in use, but let's be honest -- we're working in nothing resembling an ideal world. The fact that the API for this feature is not yet stable results directly from the fact that this feature was requested and delivered WAY too late in the development cycle. If TVL wants us to keep adding features up to the last minute, the price they must pay is the possibility that they'll need to make minor changes right up to the end of the schedule as well as those features stabilize.

We need to handle cases like this on a case-by-case basis, but in this case it is absolutely clear to me that it's better to have a very small number of apps (one app, as far as we know) make a very minor change to avoid ending up with an inferior API that we will have to live with for releases far into the future.

And again, the way to avoid cases like this is for TVL to finalize their framework feature requests FAR earlier, so that we can deliver a release at the beginning of their product development cycle in which the requested features have already been implemented, tested and refined.

Djspaceg:
I'm content with `incrementIcon` and `decrementIcon`. I'll update the PR.

Regarding `src` It's my preference that we not add the pair of `src` properties. _(We could always add them later if needed)_ Reasoning being that the most common icons to be used will be the < > ^ V + - which are all covered by our icons font.

Djspaceg:
PR (and strawman PR) updated with new names.

KunmyonChoi:
@Djspaceg, Sorry I would need to change "every app" to "every app that is using jumpToIncrement".
And, actually settings would be the app. So, maybe I was over sensitive on the API changes. And it looks like manageable change.

PR #2426, opened 2015-08-19
Enyo 2337 Display positive value for slowRewind

DavidUM:
Issue
------
slowRewind shows '-' value

Cause
-------
feedback() does not make absolute value

Fix
----
Remove '-' from msg string


ryanjduffy:
Agree with @aarontam that we don't need to update moonstone.

PR #2435, opened 2015-08-20
Enyo-2314 Change disabled Icon color and opacity to make differences

DavidUM:
Issue
------
It is not easy to distinguish disabled and enabled Icon button with translucent status

Cause
--------
Background color and opacity are same in both

Fix
----
Button and IconButton are now more translucent to allow for easier differentiation, as well as colors between them being synced-up.

This change includes:
* Several LESS variable name changes to be more consistent with other names
* Several deletions of abandoned deprecated LESS variables
* A correction to button's disabled attribute reference (it doesn't rely on a class)
* Addition and implementation of new `.vendor-opacity` mixin
* Simplification of the translucent/transparent CSS rules for Button, IconButton

### DIFF
--- a/css/moonstone-variables-light.less
+++ b/css/moonstone-variables-light.less
@@ -20,6 +20,7 @@
 @moon-spotlight-background-color: @moon-accent;
 @moon-spotlight-border-color: @moon-accent;
 @moon-spotlight-color: @moon-accent;
+@moon-button-text-color: @moon-light-text-color;
 @moon-button-background-color: @moon-white;
 @moon-divider-border-color: @moon-dark-gray;
 @moon-active-border-color: #a6a6a6;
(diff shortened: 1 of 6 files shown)


css/moonstone-variables-light.less
```diff
@@ -95,6 +95,8 @@
 @moon-input-decorator-bg-color: white;
 
 @moon-icon-font-color: @moon-dark-gray;
+@moon-icon-font-disabled-color: rgba(187, 187, 187, 0.6);
+@moon-icon-bg-disabled-color: rgba(77, 77, 77, 0.2);
 @moon-icon-button-font-color: #999;
```
DavidUM: Hi @Djspaceg 
That's not a proper value for light theme.
To prevent build fail, I put same value with dark theme but we need proper value for light theme.
Could you advise it?

lib/Icon/Icon.less
```diff
@@ -54,7 +54,7 @@
 
 .disabled .moon-icon,
 .moon-icon.disabled {
-	opacity: @moon-disabled-opacity;
-	filter: alpha(opacity=@moon-disabled-opacity-ie);
+	color: @moon-icon-font-disabled-color;
+	background-color: @moon-icon-bg-disabled-color;
```
Djspaceg: We'll need to continue to use the opacity property so that icon-images (`src` properties) will properly fade out too when set to disabled.

PR #2442, opened 2015-08-21
ENYO-1994 : Add Text to Speech Accessibility support to LightPanels

kbs12e:
Initial add accessibility feature to LightPanels.

## Requirement
WebOS TV should read panel title or panels accessibilityLabel when LightPanel is opened or moved according to UX requirements.

## Implementation
We added LightPanelsAccessibilitySupport to support reading panel title.
LightPanel is spotlight container, so first child component or last spotlight component is focused when panel is opened.
Then, Screen reader first reads this focused component content. However, UX requirements request that Screen reader first should read panel title when panel is opened or moved. 
TV should read panel title as quickly as possible before reading focused component, so we added readAlert function to setupTransitions(). The setupTransitions is called when panel is opened or moved.

## Limitation
We used readAlert() of enyo-webOS library and overrided setupTransitions to read panel title as quickly as possible, but If panel is opened without animation, panel title sound will cut for reading focused child component of panel. In this case, user can replace title to short title using accessibilityLabel. Then, TV first reads accessiblityLabel string.


kbs12e:
I will re-create PR with the latest 2.6.0-dev branch

PR #2448, opened 2015-08-21
ENYO-2234: Tooltip Does Not Display on Input

kranga-lge:
### Issue:
Tooltip Does Not Display on Input
### Fix:
Removed code in InputDecoratorAccessibilitySupport::spotlightFocusedHandler that stopped bubbling of the onSpotlightFocused event. That code was added to prevent spotlight accessibility code from incorrectly focusing the input when inputDecorator was spotted. Replaced the stop-bubbling code by adding an accesibilityHandled flag to the event, and having spotlight check for this flag before it calls focus()

Enyo-DCO-1.1-Signed-off-by: Krishna Rangarajan krishna.rangarajan@lge.com

### DIFF
--- a/lib/InputDecorator/InputDecoratorAccessibilitySupport.js
+++ b/lib/InputDecorator/InputDecoratorAccessibilitySupport.js
@@ -25,11 +25,11 @@ module.exports = {
 			}
 			this.set('accessibilityLabel', text);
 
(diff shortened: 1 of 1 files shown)


lishammel:
Closing out after discussion with Krishna.  Jira this PR was resolving was fixed a different way instead (via PLAT-8920)

PR #2452, opened 2015-08-24
Enyo 2292: Password input should read 'n-values' instead of real value.

jaewonjang:
In case password input we need to hide password.
According to ux guide it reads 'n values edit box' when the password input
is focused.

https://jira2.lgsvl.com/browse/ENYO-2292


jaewonjang:
replaced with https://github.com/enyojs/moonstone/pull/2499

PR #2459, opened 2015-08-25
ENYO-2022: Add Text to Speech Accessibility support to VideoFullscreenToggleButton

kbs12e:
Initial add accessibility feature to VideoFullscreenToggleButton.

## Requirement
WebOS TV should read 'original screen' or 'full screen' when VideoFullscreenToggleButton is focused.

## Implementation
VideoFullscreenToggleButton is iconButton, which default has 'exitfullscreen' as icon value, so screen reader should default read 'original screen'. In the other hands, if user click VideoFullscreenToggleButton, its icon value is changed to 'fullscreen', so screen reader should read 'full sreen'.
Thus, I add accessibilityLabel according to icon.


kbs12e:
Close for applying the latest code. I will re-create again.

PR #2461, opened 2015-08-25
ENYO-1907: Add Text to Speech Accessibility support to Drawers

kbs12e:
Initial add accessibility feature to Drawers.

## Requirement
Screen reader should read 'open drawer' or 'close drawer' when Drawer is focused.

## Implementation
Screen reader should read 'open drawer' or 'close drawer' when activator of drawers is focused according to drawer open or close state. 
Whenever drawer is opened or closed, updateActivator is called and Drawers could know whether or not drawer is open using isDrawerOpen function 
Thus, I overrided updateActivator function in new added DrawerAccessibilitySupport and in this function, added accessibilityLabel. 


kbs12e:

PR #2462, opened 2015-08-26
ENYO-1908: Add Text to Speech Accessibility support to ListActions

kbs12e:
Initial add accessibility feature to ListActions

## Requirement
1) Screen reader should read ListAction IconButton's tooltip text when it is focused.
2) Screen reader should read ListAction Item content and state when ListAction's item is focused.
3) Screen reader should read close string when ListActions close button is focused.

## Implementation
Only No.3 requirement is applied because the others are already applied in each component.
To read 'close' when close button is focused, we add accessibilityLabel to IconButton. 


kbs12e:
I will re-create PR with latest code

PR #2463, opened 2015-08-26
Enyo 2391 Prevent to close popup with back key when spotlightModal is true.

DavidUM:
Requirement from HE UX

Issue
-------
When spotlightModal is true, we should block to close popup although back key is pressed.

Cause
--------
Currently popup will be closed when back key is pressed

Fix
----
Add guard code to prevent closing popup

### DIFF
--- a/src/ContextualPopup/ContextualPopup.js
+++ b/src/ContextualPopup/ContextualPopup.js
@@ -596,9 +596,8 @@ module.exports = kind(
 	* @private
 	*/
 	backKeyHandler: function() {
-		if (this.showing) {
-			this.hide();
-		}
+		if (this.spotlightModal) this.pushBackHistory();
+		else if (this.showing) this.hide();
 
 		if (this.spotlight && !this.spotlightDisabled) {
 			this.doRequestSpot();
--- a/src/Popup/Popup.js
+++ b/src/Popup/Popup.js
@@ -592,5 +592,21 @@ module.exports = kind(
 	destroy: function() {
 		this.showHideScrim(false);
 		Popup.prototype.destroy.apply(this, arguments);
-	}
+	},
+
+	/**
+	* If [spotlightModel]{@link module:moonstone/Popup~Popup#spotlightModal} property is true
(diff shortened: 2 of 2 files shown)


lib/Popup/Popup.js
```diff
@@ -592,5 +592,21 @@ module.exports = kind(
...
+	* If [spotlightModel]{@link module:moonstone/Popup~Popup#spotlightModal} property is true
+	* back key could not close popup.
+	* Because every event out-boundary of popup should be ignored when it is true.
+	*
+	* @method
+	* @public
+	*/
+	backKeyHandler: kind.inherit(function (sup) {
+		return function () {
+			if (this.spotlightModal) this.pushBackHistory();
```
Djspaceg: What does adding `this.pushBackHistory();` do?

ryanjduffy: It undoes the back action. If you hit the back key (or the back button in the browser), that'll remove the popup's entry from the history stack and normally hide it (the default backKeyHandler behavior). Since it's not supposed to close though, that pushes back onto the stack so it can be removed properly in `showingChanged`.

Djspaceg: Ah ha. Thanks for the description @ryanjduffy.

DavidUM: Thank you for your detailed description :+1:

DavidUM:
Hi @aarontam 
I found that there was a miss communication from HE UX.
What they intended was `modal` not `spotlightModal`
I will revert this and implement again for modal.
Sorry for making inconvenience.

aarontam:
Hi @DavidUM, no problem, thanks for the clarification! I think that if we condition this logic on `modal`, we may run into the same problem, though, as `moonstone/Popup` defaults to true for that property, and will prevent subkinds such as `moonstone/Dialog` from hiding the control when the back key is pressed. We should confirm the UX here and make sure that it makes sense to prevent the use of the back key when the control is `modal:true`.

PR #2464, opened 2015-08-26
ENYO-2172: Add Text to Speech Accessibility support to PagingControls

kbs12e:
Initial add accessibility feature to PagingControls

## Requirement
1) Screen reader should read scroll IconButton string when IconButton is focused.
2) Screen reader should read scroll direction string whenever this scroll iconButton is pressed.

## Implementation
The IconButton of Scroller is PagingControls component. 
I added PagingControlsAccessibilitySupport to override noop and depress function.

The noop() is onSpotlightFocused event handler, and simply returns true. so, Even If pagingControls is spotlight focused, screen reader should not read anything because pagingControls does not get the webkit focus. To prevent this, we directly add webkit focus before noop() is called.

We observer side property and when side is defined or changed, add accessibilityLabel to pagingControls.

We override depress() for reading scroll direction (ex. 'up' or 'down' ) whenever pagingControls is pressed. In this function, we use readAlert() api in enyo-webos because pagingControl's focus is not moved and screen reader should read string whenever it is pressed

## Limitation
We used readAlert() of enyo-webOS library, so No.2 requirement works on only webOS TV. 


kbs12e:
I will re-create PR with latest code

PR #2465, opened 2015-08-26
Enyo 2171 Add tabindex for Item to prevent accessibility timing issue

kbs12e:
## Issue 
 screen reader sometimes does not read child component's content when Item is focused.

## Cause
Item may have components as child. However, Sometimes screen reader does not read child component's content because time to receive tabindex is late than child.
To prevent this timing issue, I add tabindex to Item.

## Fix
Add tabindex in ariaObservers. 


kbs12e:

PR #2466, opened 2015-08-26
ENYO-2165: ScrollerVertical: Selectable Item's Icon Offsets with Text…

alischuk:
… in Different Spacing

### DIFF
--- a/lib/SelectableItem/SelectableItem.less
+++ b/lib/SelectableItem/SelectableItem.less
@@ -3,20 +3,15 @@
 	padding-left: (@moon-item-indent + @moon-spotlight-outset);
 
 	&:before {
-		content: '';
+		content: '\2022';
 		position: absolute;
 		left: @moon-spotlight-outset;
-		top: 50%;
-		-webkit-transform: translateY(-50%);
-		transform: translateY(-50%);
-		width: @moon-selectable-item-indicator-width;
-		height: @moon-selectable-item-indicator-height;
-		border-radius: (@moon-selectable-item-indicator-width / 2);
-		background-color: @moon-accent;
+		color: @moon-accent;
+		font-size: 56pt;
 	}
 
 	&.spotlight:before {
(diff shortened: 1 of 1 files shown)


lib/SelectableItem/SelectableItem.less
```diff
@@ -3,20 +3,15 @@
 	padding-left: (@moon-item-indent + @moon-spotlight-outset);
 
 	&:before {
-		content: '';
+		content: '\2022';
```
dragongears: I see a few issues with using a bullet character in this instance, with regards to its size. First, the bullet character here is larger than the original bullet character, ignoring @moon-selectable-item-indicator-width and @moon-selectable-item-indicator-height. Second, the size of the bullet is inconsistent with the size of Radio Item indicator. Maybe a UX person can provide some input?

lib/SelectableItem/SelectableItem.less
```diff
@@ -3,20 +3,15 @@
...
 		left: @moon-spotlight-outset;
-		top: 50%;
-		-webkit-transform: translateY(-50%);
-		transform: translateY(-50%);
-		width: @moon-selectable-item-indicator-width;
-		height: @moon-selectable-item-indicator-height;
-		border-radius: (@moon-selectable-item-indicator-width / 2);
-		background-color: @moon-accent;
+		color: @moon-accent;
+		font-size: 56pt;
```
Djspaceg: Change the font size to 150% instead of points (we don't use points anywhere).

lib/SelectableItem/SelectableItem.less
Djspaceg: Change this line to be `content: '\0F0013';` and add: `font-family: "Moonstone Icons";`
Using the circle glyph from Moonstone Icons will generate consistent results on multiple platforms with various fonts installed.

alischuk:
Closed this request. New one - https://github.com/enyojs/moonstone/pull/2474

PR #2467, opened 2015-08-26
ENYO-2396-New DataList:Unwanted behavior on onSpotlightLeft

rajyavardhanp:
Issue: OnSpotlightLeft is not working as expected
Cause: 5-way guard code for the scroll prevents the scroll to roll
Fix: guard code is added to handle special cases.


lib/NewDataList/NewDataList.js
```diff
@@ -38,6 +38,10 @@ module.exports = kind({
 
 		e = event.type;
 
+		if((e === 'onSpotlightUp' && this.direction == 'horizontal') || (e === 'onSpotlightLeft' && this.direction == 'vertical')) {
```
ryanjduffy: * Add a space between `if` and `(`
* Let's combine this condition with the one above for `isAccelerating()`
* We need to check for either direction (up or down with horizontal and left or right with vertical). You can still repro the bug by going to the *last* row, highlighting the first item, and holding down the right key.

ryanjduffy:
You'll also need to rebase on 2.6.0-dev due to the path change from lib to src

lib/NewDataList/NewDataList.js
rajyavardhanp: updated the code as per review comments.
But not sure about combining the conditions as  acceleration is relatively less frequent  I guess (in performance perspective).
Thankyou.

PR #2470, opened 2015-08-27
ENYO-2384: Remove unnecessary transform.

aarontam:
### Issue
Now that we are transitioning the panel container, we no longer need to apply the transform to the individual panels.

### Fix
The unnecessary transform was removed form the panels themselves and we utilize the `transitioning` class applied to the `LightPanels` kind. Also, removed a deprecated module and fixed a bug with the `preTransition` logic (the current spotted item needed to be checked for spottability).

This change coincides with https://github.com/enyojs/enyo/pull/1260.


lib/LightPanels/LightPanel.less
```diff
@@ -22,12 +22,8 @@
...
 	// the DOM during ".transitioning". This was altered to improve marquee animation performance.
 	// Pseudo-elements without a content property cannot receive styling.
-	&.transitioning {
-		-webkit-transform: translateZ(0);
-		transform: translateZ(0);
-
-		&:after {
-			content: '';
-		}
+	.moon-light-panels.transitioning &:after {
```
aarontam: I think this looks a little awkward, but wanted to take advantage of the parent kind having this class already, and it having the same meaning in that context. Without using `moon-light-panels`, I was afraid this would not be specific enough.

PR #2471, opened 2015-08-27
ENYO-2294: Prevent reading time information in player control

kbs12e:
## Requirement
Screen reader should not read time information of player bottom control, according to UX requirements.

## Implementation
VideoTransportSlider has time information control, so we add accessibilityDisabled to it.


kbs12e:
Close by CI build fail.

PR #2472, opened 2015-08-27
Enyo 2294 Prevent reading a time information in VideoPlayer control

kbs12e:
## Requirement
Screen reader should not read time information of player bottom control, according to UX requirements.

## Implementation
VideoTransportSlider has time information control, so we add accessibilityDisabled to it.


ryanjduffy:
Included in #2476

PR #2473, opened 2015-08-27
ENYO-2293: When IntegerPicker, SimpleIntegerPicker focused, reads Hint message with value.

jaewonjang:
According to UX
1. IntegerPicker
- When IntegerPicker is focused, it reads 'value' +
'change a value with up/down button'
- When the current value is changed, it reads 'value'
2. SimpleIntergerPicker
- When SimpleIntergerPicker is focused. it reads 'value' +
'change a value with left/right button'
- When the current value is changed, it reads 'value'

To meet UX we set the role spinbutton and aria-valuenow
on repeater because if we set it on IntergerPicker itself
screen reader always read 'value' + 'change a value with up/down button'

https://jira2.lgsvl.com/browse/ENYO-2293
nyo-DCO-1.1-Signed-off-by: Jaewon Jang <jaewon98.jang@lgepartner.com>


jaewonjang:
replaced with https://github.com/enyojs/moonstone/pull/2487

PR #2474, opened 2015-08-27
ENYO-2165: ScrollerVertical: Selectable Item's Icon Offsets with Text…

alischuk:
Enyo-DCO-1.1-Signed-off-by: Serge Alischuk serhei.alyshchuk@lge.com

### DIFF
--- a/lib/SelectableItem/SelectableItem.less
+++ b/lib/SelectableItem/SelectableItem.less
@@ -3,16 +3,13 @@
 	padding-left: (@moon-item-indent + @moon-spotlight-outset);
 
 	&:before {
-		content: '';
+		content: @moon-icon-circle;
 		position: absolute;
+		top: 0;
 		left: @moon-spotlight-outset;
-		top: 50%;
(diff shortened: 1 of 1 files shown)


lib/SelectableItem/SelectableItem.less
```diff
@@ -3,16 +3,13 @@
 	padding-left: (@moon-item-indent + @moon-spotlight-outset);
 
 	&:before {
-		content: '';
+		content: @moon-icon-circle;
 		position: absolute;
+		top: 0;
```
Djspaceg: Do we need to specify `top` here? The rest of the changes look great. Thanks for finding and using the variables too. (You can just commit and push to this branch, and the PR will update. No need to close and reopen. 😃)

alischuk:
For 'top' the default value is 'auto'. Some browsers handle it differently. For example in Сhrome (for the desktop) shows that the bullet will be shifted down (for SeleсtableItem Simple). If you assign top: 0, everything works correctly.

lishammel:

PR #2480, opened 2015-08-28
ENYO-2295: Accessibility scenario of the slider is changed

jaewonjang:
Chaged slider accessibility code to meet ux.
- When the slider focused, it reads value(including popup content
or value percentage).
- When the slider selected, it reads the hint
'change a value with left right button'
- When the value is changed, it reads current value.
- If there are left/right(up/down) button and when they are focused
it reads 'value' + 'press ok button to decrease/increase the value'
- When the value is changed with left/right button, it reads
current value

Additional
- set 'close' as accessibilityLabel on close button in dialog

https://jira2.lgsvl.com/browse/ENYO-2295
nyo-DCO-1.1-Signed-off-by: Jaewon Jang <jaewon98.jang@lgepartner.com>


jaewonjang:
replaced with https://github.com/enyojs/moonstone/pull/2492

PR #2481, opened 2015-08-28
Enyo 2020 Add Text to Speech Accessibility support to CaptionDecorator

kbs12e:
## Requirement
WebOS TV should read captionDecorator content before reading child component's content.

## Implementation
Generally, CaptionDecorator wraps child control and it is not spottable component, so when child component is focused, Screen reader read child component's content or label including captionDecorator's content. To implement this, I add aria-label, which captionDecorator content is appended to child's content.

## Limitation
CaptionDecorator may have user-defined component as child. When aria-label set to child, It can happen exception case. In current version, support CaptionDecorator, which has Button or ToggleButton as child.  


kbs12e:
This requirement could implement in app side. we provide guide documentation for developer.

PR #2483, opened 2015-08-28
Enyo 2401 artdahm

dragongears:


webOS101:
This PR is against master, not 2.6.0-dev.

PR #2485, opened 2015-08-28
ENYO-2240: Tooltip: Disabled Button Prevents Marquee to get 'Start' Event

kranga-lge:
### Issue:
Disabled buttons intermittently do not marquee or tooltip
### Fix:
re-inserting the button overlay child node. The overlay had been replaced with a :before tag (via the moon-taparea less mixin) to save the overhead of an extra node. Pseudo-selectors do not get mouse events when the button is disabled, while the overlay child node does.

Enyo-DCO-1.1-Signed-off-by: Krishna Rangarajan <krishna.rangarajan@lge.com>

### DIFF
--- a/src/Button/Button.js
+++ b/src/Button/Button.js
@@ -8,6 +8,7 @@ require('moonstone');
 var
 	kind = require('enyo/kind'),
 	util = require('enyo/utils'),
+	Control = require('enyo/Control'),
 	Button = require('enyo/Button');
 
 var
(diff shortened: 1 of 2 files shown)


Djspaceg:
I have several concerns with this approach. Adding a new DOM element in the way this is implemented means that this new element will be receiving mouse events all the time, not just when the control is disabled and not able to handle them on its own. This raises issues where events are handled by children of this control without direct access to the button-overlay, and events fired from it will now have an event.originator of button-overlay and not button itself. The new button-overlay element will also only appear if there are no client controls, which means that buttons who have components or a client area in them won't work the same way normal disabled buttons do.

clinuz:
Closing in favor of https://github.com/enyojs/moonstone/pull/2490. I have the same concerns as @Djspaceg on this one.

PR #2489, opened 2015-08-31
ENYO-2411 IntegerPicker does not update range at creation

dragongears:

### DIFF
--- a/src/IntegerPicker/IntegerPicker.js
+++ b/src/IntegerPicker/IntegerPicker.js
@@ -248,6 +248,7 @@ module.exports = kind(
 	create: function () {
 		Control.prototype.create.apply(this, arguments);
 		this.verifyValue();
+		this.calculateRange();
 		this.updateOverlays();
 	},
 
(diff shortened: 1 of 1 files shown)


ryanjduffy:
`range` isn't public; it's private. See [L243](https://github.com/enyojs/moonstone/blob/2.6.0-dev/src/IntegerPicker/IntegerPicker.js#L243). Also, while the range isn't calculated during `create`, it is called by `rendered()`. What's the actual bug here?

dragongears:
I missed that it was being calculated in rendered(). This is probably a non-bug then. Is there a reason it's being done in rendered() instead of create()? Does it have something to do with the (width) buffer?

ryanjduffy:
I don't see an obvious reason. No part of the logic (now at least) requires the node to exist. I tracked it back to being there in Sept 2013 so it predates me by quite a bit.

dragongears:
Which node are you talking about? The buffer? It confused me for a little while then I realized that it's to keep the control's pink background at a constant width while the width of the numbers change.

ryanjduffy:
Any node, afaik. I didn't see anything in `rangeChanged` or the methods it calls that needed a node to exist.

PR #2500, opened 2015-09-02
ENYO-1997: Add Text to Speech Accessibility support to Overlay

jaewonjang:
We need to apply checkbox role on SelectionOverlaySupport.
Set the checkbox role and aria-cheched attributes in
SelectionOverlaySupport.

https://jira2.lgsvl.com/browse/ENYO-1997
Enyo-DCO-1.1-Signed-off-by: Jaewon Jang jaewon98.jang@lgepartner.com

### DIFF
--- a/src/Overlay/Overlay.js
+++ b/src/Overlay/Overlay.js
@@ -435,6 +435,13 @@ module.exports.Selection = {
 	/**
 	* @private
 	*/
+	_selectionOverlay_Handlers: {
+		onSpotlightFocused: '_selectionOverlay_spotlightFocused'
+	},
+
+	/**
+	* @private
+	*/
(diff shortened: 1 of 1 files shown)


src/Overlay/Overlay.js
```diff
@@ -431,6 +431,13 @@ module.exports.Selection = {
 	/**
 	* @private
 	*/
+	handlers: {
+		onSpotlightFocused: 'spotlightFocused'
```
ryanjduffy: It's really not safe to use the handlers block from a mixin because you'll likely stomp on the target instances handlers. That'll happen here with GridListImageItem which already has a onSpotlightFocused handler. You'll probably need to do something like MarqueeSupport does to be safe.

PR #2508, opened 2015-09-02
Enyo-2003 Marquee is not starting for "Revoke all pairing authorizations" pop-up.

dmikey:
Issue: 

webOS Webkit, is misreading font widths without font-kerning applied to them.

Fix:

Add the font-kerning mixin


clinuz:
Closing because we are taking the update for 2.6.0

PR #2513, opened 2015-09-02
ENYO-2411 IntegerPicker does not update range at creation

dragongears:

### DIFF
--- a/src/IntegerPicker/IntegerPicker.js
+++ b/src/IntegerPicker/IntegerPicker.js
@@ -247,7 +247,7 @@ module.exports = kind(
 	*/
 	create: function () {
 		Control.prototype.create.apply(this, arguments);
-		this.verifyValue();
+		this.rangeChanged();
 		this.updateOverlays();
 	},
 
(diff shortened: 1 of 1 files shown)


src/IntegerPicker/IntegerPicker.js
```diff
@@ -247,7 +247,7 @@ module.exports = kind(
 	*/
 	create: function () {
 		Control.prototype.create.apply(this, arguments);
-		this.verifyValue();
```
Djspaceg: We no longer need to run `verifyValue`? That sets the value and allows the control to already be set to the value without animating.

dragongears: verifyValue() gets called inside rangeChanged(), so I removed it.

Djspaceg: Super. Didn't see that.

src/IntegerPicker/IntegerPicker.js
```diff
@@ -247,7 +247,7 @@ module.exports = kind(
...
+		this.rangeChanged();
```
clinuz: Calling this in `create` causes the `onChange` event to be fired during create. Which is something we REALLY don't want. It was being avoided in the current implementation and may cause problems by changing this.

PR #2516, opened 2015-09-03
ENYO-2449: ListAction: skip to destroy controls at creation time

kyo504:
### Issue
This is an additional work for ENYO-2218. As you see code change, client controls were destroyed no matter what they exist or not but it's not necessary at creation time

### Fix
Add guard code to skip destroying client controls at creation time

Enyo-DCO-1.1-Signed-off-by: Hunkyo Jung <hunkyo.jung@lge.com>

### DIFF
--- a/src/ListActions/ListActions.js
+++ b/src/ListActions/ListActions.js
@@ -484,7 +484,9 @@ var ListActions = module.exports = kind(
 		var listAction, i;
 
 		this.listActionComponents = [];
-		this.$.listActions.destroyClientControls();
+		if (this.generated) {
+			this.$.listActions.destroyClientControls();
+		}
 
 		for (i = 0; (listAction = this.listActions[i]); i++) {
 			this.listActionComponents.push(this.createListActionComponent(listAction, owner));


kyo504:
Hi @aarontam , could you review this? This is what we talked about ENYO-2218.

src/ListActions/ListActions.js
```diff
@@ -484,7 +484,9 @@ var ListActions = module.exports = kind(
 		var listAction, i;
 
 		this.listActionComponents = [];
-		this.$.listActions.destroyClientControls();
+		if (this.generated) {
```
clinuz: This can't be taken because `generated` isn't a safe-guard for actual client controls existence. I'm not sure that this bug is actually a bug? It is harmless/optimizable function calls that are essentially a nop when no controls exist. Why is this even being considered?

aarontam: There was no real bug here, just a fairly minor optimization for create time that I had forgotten to update when reviewing a previous PR. The guard isn't necessarily for the existence of client controls, but more that we we don't need to do this at create time.

PR #2528, opened 2015-09-11
ENYO-2501: Lost focus when set initialFocusIndex on disabled item in DataList

KunmyonChoi:
Issue:
DataList is give focus on specific item when initialFocusIndex is set.
But, when that item is disabled, Spotlight fails give focus.

Fix:
Check spottability on item before give focus, then give focus on list
instead when that item is not focusable. This workaround will give focus
to the first focusable item on list.

### DIFF
--- a/src/DataList/DataList.js
+++ b/src/DataList/DataList.js
@@ -68,7 +68,10 @@ var DataListSpotlightSupport = {
 				child = this.childForIndex(inIndex);
 			}
 			subChild = inSubChild ? Spotlight.getChildren(child)[inSubChild] : child;
-			Spotlight.spot(subChild);
+			if (Spotlight.isSpottable(subChild)) 
+				Spotlight.spot(subChild);
+			else
+				Spotlight.spot(this); 
 		} else {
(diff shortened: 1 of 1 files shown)


src/DataList/DataList.js
```diff
@@ -68,7 +68,10 @@ var DataListSpotlightSupport = {
 				child = this.childForIndex(inIndex);
 			}
 			subChild = inSubChild ? Spotlight.getChildren(child)[inSubChild] : child;
-			Spotlight.spot(subChild);
+			if (Spotlight.isSpottable(subChild))
```
ryanjduffy: This logic here is sound. Add `{}` around the `if`/`else` blocks.

An alternative, though perhaps not any better, would be to inline the two `spot()` commands. Since `spot()` calls `isSpottable()` internally and will return false if not, you can do the following to the same effect:

```
Spotlight.spot(subChild) || Spotlight.spot(this);
```

KunmyonChoi: That sounds better :)

PR #2529, opened 2015-09-11
ENYO-2502: Lost focus when create DataList dynamically on LightPanel

KunmyonChoi:
Issue:
When datalist is created dynamically and there is no focusable item,
lightPanels is give focus to spotlightPlaceholder.
But, datalist is fails to give focus when didRender is called because
current focus is not datalist and datalist is not trying to transfer
focus to inner item.

Fix:
Add condition to see the current focus is on spotlightPlaceholder or
spotlightDummy.


KunmyonChoi:
Close this PR per the issue is closed as not a bug.

PR #2533, opened 2015-09-13
Enyo-2516 Update header text when locale is changed

DavidUM:
Signed-off-by: David Joonho Um <david.um@lge.com>


ryanjduffy:
Closing in favor of #2540

PR #2534, opened 2015-09-14
Enyo-2522 Reflow each picker when timePicker is opening

DavidUM:
`this.pickers` represents a set of picker like hour and minute.
In setupPicker(), we compose it.
Then when timePicker open, it will reflow each picker using `this.pickers`


DavidUM:
Thx @webOS101 
I amend it. Looks better! :+1:

PR #2535, opened 2015-09-14
ENYO-2521: Make viewport a spotlight container without remembering last focus

KunmyonChoi:
Issue:
When there is a disabled item between items on scroller, there is a
chance to return paging control as a best match even when there is
focusable item below. This is because paging control is having smaller
y distance than below item.

Fix:
It is risky to alter _getPrecedenceValue logic on spotlight.
But, we have new option that is not remembering last focus on container.
We set viewport of scroller as spotlight container and set option
spotlightRememberFocus as false. So, spotlight can consider items first
on 5way focus move.


graynorton:
Closing in favor of https://github.com/enyojs/moonstone/pull/2548

PR #2539, opened 2015-09-14
ENYO-2468: Accessibility: Screen reader does not read the unit of Dat…

jaewonjang:
…aPicker and TimePicker.

Issue
It doesn't read unit(day, month, hour etc) on DatePicker and TimePicker

Cause
aria-label is overrided in IntegerPicker after it is set by units
in TimePicker and DatePicker.

Fix
On IntergerPicker we just handle only accessibilityHint and aria-valuenow
to keep the accessibilityLabel which is set by TimePicker and DatePicker.
To do this we set the accessibilityRole of IntegerPicker to spinbutton and
only set the hint when spotlight is focused and then remove hint during
value changed.

https://jira2.lgsvl.com/browse/ENYO-2468

### DIFF
--- a/src/TimePicker/TimePicker.js
+++ b/src/TimePicker/TimePicker.js
@@ -171,7 +171,7 @@ var MeridiemPicker = kind(
 		{path: 'spotted',  method: function () {
 			// When spotlight is focused, it reads value with hint
 			if (this.spotted) {
-				this.set('accessibilityHint', $L('change a value with up/down button'));
+				this.set('accessibilityHint', $L('change a value with up down button'));
 				this.setAriaAttribute('aria-valuetext', this.meridiems[this.value].name);
 			}
 		}}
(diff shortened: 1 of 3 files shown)


src/TimePicker/TimePicker.js
```diff
@@ -162,7 +162,18 @@ var MeridiemPicker = kind(
...
+			// When value is changed, it reads only value
+			if (this.spotted) {
+				this.set('accessibilityHint', null);
+				this.setAriaAttribute('aria-valuetext', this.meridiems[this.value].name);
+			}
+		}},
+		{path: 'spotted',  method: function () {
+			// When spotlight is focused, it reads value with hint
+			if (this.spotted) {
+				this.set('accessibilityHint', $L('change a value with up/down button'));
```
seunghoh: Slash of up/down could cause a problem for TTS. Ask UI to change string to up and down

PR #2540, opened 2015-09-14
Fixes ENYO-2516

ryanjduffy:
For subkinds of DateTimePickerBase, the value of currentValueText is
formatted for the locale. The local locale member is deprecated by
ENYO-1859 so we must manually notify observers of currentValueText to
refetch its value when the locale changes.


clinuz:
This is blocked by https://github.com/enyojs/moonstone/pull/2531. The additional commits will disappear once that has been merged due to base commit in history.

PR #2544, opened 2015-09-14
ENYO-2496: Updated Drawers logic to use a single universal icon method…

Djspaceg:
… independent of property states.


ryanjduffy:
Closing in favor of #2547

PR #2546, opened 2015-09-15
ENYO-2521: Add spotlightPrecedence option handler

KunmyonChoi:
Issue:
Spotlight is having fixed calculation weight factor on
_getPrecedenceValue. But, there is some cases that angle is more
important than distance like this issue.

Fix:
We add an option to setup the weight. So, developer can set option on
each control.


KunmyonChoi:
Close this PR per we decided will not fix this issue.

PR #2547, opened 2015-09-15
Fixes ENYO-2496

ryanjduffy:
fix changing drawer activator icon at runtime


Djspaceg:
if `updateActivator()` is executed and there is no change to `_activated`, then the `changed` method won't fire. `updateActivator()` is fired from many places for multiple reasons.

ryanjduffy:
The other alternative is to force the notification. I considered that but it didn't seem necessary at first review. That seems like the better solution however. Updating this PR.

Djspaceg:
Cool. Thanks!

PR #2549, opened 2015-09-16
ENYO-2505 ItemOverlay: Text String is Truncated in Auto Hide

dragongears:

### DIFF
--- a/src/Marquee/Marquee.js
+++ b/src/Marquee/Marquee.js
@@ -950,7 +950,7 @@ var MarqueeItem = {
 		var node = this.$.marqueeText ? this.$.marqueeText.hasNode() : this.hasNode(),
 			rect;
 
-		if (node && this._marquee_distance == null && this.getAbsoluteShowing()) {
+		if (node && this.getAbsoluteShowing()) {
 			rect = node.getBoundingClientRect();
 			this._marquee_distance = Math.floor(Math.abs(node.scrollWidth - rect.width));
 


src/Marquee/Marquee.js
```diff
@@ -950,7 +950,7 @@ var MarqueeItem = {
 		var node = this.$.marqueeText ? this.$.marqueeText.hasNode() : this.hasNode(),
 			rect;
 
-		if (node && this._marquee_distance == null && this.getAbsoluteShowing()) {
+		if (node && this.getAbsoluteShowing()) {
```
Djspaceg: Won't this mean that the marquee always re-measures? Seems like a major performance benefit loss. May be good to look into a way to invalidate `_marquee_distance` before we get here in those special cases where it's needed, so we keep the optimization.

dragongears: Where you say "always re-measures" I say "consistently re-measures". :-) Is it really that much of a performance hit? I was thinking of calling reflow() on the marquee but don't know the best way to trigger that given that ItemOverlay uses CSS rules to toggle.

Djspaceg: "measures every single time, even when it doesn't need to"

On the initial render alone, `_marquee_calcDistance` is called 7 times by every marquee. If you have a normal DataGridList, where each of the 21 items on screen has a caption and subCaption, each of which marquee, that's 294 calls for one screen update. `_marquee_calcDistance` preforms a costly DOM reading operation `getBoundingClientRect`, and the numbers add up quickly. We need to keep this function in particular as lean as possible with as much pre-caching as possible and exit quickly if possible.

`ItemOverlay` probably needs to get refactored to use the showing property instead of basic CSS hiding, so marquee can be aware of its state change.

PR #2553, opened 2015-09-17
PLAT-6939: Add Text to Speech Accessibility support to Scroller, Pagi…

jaewonjang:
…ngControl

Requirement

1) Screen reader should read scroll IconButton string when IconButton is focused.
2) Screen reader should read scroll direction string whenever this scroll iconButton is pressed.

Implementation

The IconButton of Scroller is PagingControls component.
Add private property spotted and pressed.
When spotted is true, set accessibilityLabel 'scroll up' and call focus api because
return true in noop.
When pressed is true, set accessibilityAlert and accessibilityLabel to read
changed label automatically.

https://jira2.lgsvl.com/browse/PLAT-6939

### DIFF
--- a/src/PagingControl/PagingControl.js
+++ b/src/PagingControl/PagingControl.js
@@ -11,7 +11,8 @@ var
 	animation = require('enyo/animation');
 
 var
-	Spotlight = require('spotlight');
+	Spotlight = require('spotlight'),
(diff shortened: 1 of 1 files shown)


ryanjduffy:
Closing in favor of #2565

PR #2557, opened 2015-09-18
ENYO-2505 ItemOverlay: Text String is Truncated in Auto Hide

dragongears:

### DIFF
--- a/src/ItemOverlay/ItemOverlay.js
+++ b/src/ItemOverlay/ItemOverlay.js
@@ -84,8 +84,10 @@ var ItemOverlay = module.exports = kind(
 	/**
 	* @private
 	*/
-	autoHideChanged: function() {
-		this.addRemoveClass('auto-hide', this.get('autoHide'));
+	autoHideChanged: function(was, is) {
+		if (this.autoHide) {
+			this.set('showing', (was || this.spotFocused));
+		}
 	},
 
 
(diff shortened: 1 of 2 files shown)


src/ItemOverlay/ItemOverlay.js
```diff
@@ -163,6 +165,42 @@ ItemOverlay.ItemOverlaySupport = {
...
+			this.$.ending.set('showing', showing);
+		}
+	},
+
+	/**
+	* @private
+	*/
+	spotFocused: function() {
+		this.overlayShowing(true);
+		this.resize();
```
aarontam: We would only need to resize if `autoHideBeginning` or `autoHideEnding` were true, right?

src/ItemOverlay/ItemOverlay.js
```diff
@@ -84,8 +84,10 @@ var ItemOverlay = module.exports = kind(
 	/**
 	* @private
 	*/
-	autoHideChanged: function() {
-		this.addRemoveClass('auto-hide', this.get('autoHide'));
+	autoHideChanged: function(was, is) {
+		if (this.autoHide) {
+			this.set('showing', (was || this.spotFocused));
```
aarontam: I'm not quite sure I follow where `this.spotFocused` is coming from and why it is being checked here.

aarontam: Also, if the intent is to only hide when we are enabling `autoHide` (and not vice versa as the overlay should remain hidden until focused), we could simplify this to

`if (this.autoHide) this.set('showing', false);`

as we know `was` would necessarily have a value of `false` in this case, since it is a change handler. We would still need to account for the case of the overlay currently being focused, which I am guessing was the intent of `this.spotFocused`, but that property is never set anywhere.

src/ItemOverlay/ItemOverlay.js
dragongears: You're probably right. I'll move the resize into overlayShowing().

src/ItemOverlay/ItemOverlay.js
dragongears: Yeah. I don't know what drugs I was on. :-) The amount of time that has passed between when I worked on this fix and when it was reviewed was long enough that I don't remember what I was thinking. Maybe I planned on setting spotFocused from the parent and forgot? Anyway, I reworked it and hopefully it's better now.

PR #2563, opened 2015-09-22
Fix for: ENYO-2539

clinuz:
This is an alternative fix from https://github.com/enyojs/moonstone/pull/2552 so as to avoid mis-using the binding's transform callback.

Enyo-DCO-1.1-Signed-Off-By: Cole Davis (cole.davis@lge.com)


clinuz:
Closing as we should accept https://github.com/enyojs/moonstone/pull/2552 as the least intrusive complete solution to the issue.

PR #2564, opened 2015-09-22
ENYO-2547: ProgressBar can now generate/destroy its popup as needed.

Djspaceg:
ProgressBar was expecting `popup` to always be true by directly referencing components that are only generated at create-time when `popup` is true. ProgressBar now automatically creates and destroys the `popupComponents` as needed based on the current setting of `popup`. Guard code has been added to updatePopup that safely isolates the methods which require `this.$.popup` to be present. Also some incidental cleanup on Slider.


aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2568.

PR #2566, opened 2015-09-24
ENYO-2573: Updated Header to accommodate ApplicationCloseButton's…

Djspaceg:
… presence when included in Panels.

An update to the `adjustTitleWidth` method was necessary to factor in the margin from the close button's rules.

### DIFF
--- a/src/Header/Header.js
+++ b/src/Header/Header.js
@@ -638,7 +638,7 @@ module.exports = kind(
 			// Measure client area's width + 40px of spacing
 			client = this.$.client ? this.$.client.hasNode() : null,
 			clientWidth = client ? client.offsetWidth : null,
-			clientStyle = client ? (client.currentStyle || window.getComputedStyle(client)) : null,		// Originally by YuC @ http://stackoverflow.com/questions/349257/detecting-true-border-padding-and-margin-from-javascript
+			clientStyle = client ? (client.currentStyle || dom.getComputedStyle(client)) : null,		// Originally by YuC @ http://stackoverflow.com/questions/349257/detecting-true-border-padding-and-margin-from-javascript
 			clientMargin = client ? (parseInt(clientStyle.marginLeft, 10) + parseInt(clientStyle.marginRight, 10)) : null,
 			clientSpaceSmall = dom.unit(clientWidth + clientMargin + ri.scale(36), 'rem'),
 			clientSpace = dom.unit(clientWidth + ri.scale(36), 'rem'),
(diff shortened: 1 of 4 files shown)


src/Header/Header.js
```diff
@@ -638,13 +638,16 @@ module.exports = kind(
 			// Measure client area's width + 40px of spacing
 			client = this.$.client ? this.$.client.hasNode() : null,
 			clientWidth = client ? client.offsetWidth : null,
+			clientStyle = client ? (client.currentStyle || window.getComputedStyle(client)) : null,		// Originally by YuC @ http://stackoverflow.com/questions/349257/detecting-true-border-padding-and-margin-from-javascript
```
aarontam: Should we use our `dom.getComputedStyle` instead?

src/ListActions/ListActions.js
```diff
@@ -172,6 +173,7 @@ var ListActionsDrawer = kind(
 		// Skip animation before render time
 		if (!this.$.client.hasNode()) { return; }
 		this.$.client.addRemoveClass('open', this.open);
+		this.doCustomizeCloseButton({showing: (this.open ? false : true)});
```
aarontam: Probably can simplify this to `{showing: !this.open}`.

src/Header/Header.js
Djspaceg: Oops. Totally.

src/ListActions/ListActions.js
Djspaceg: Yeah! Not sure why I did that super verbose way...

src/ListActions/ListActions.js
```diff
@@ -400,7 +402,7 @@ var ListActions = module.exports = kind(
 	*/
 	drawerComponents: [
 		{name: 'drawer', spotlightDisabled: true, kind: ListActionsDrawer, classes: 'list-actions-drawer', onComplete: 'drawerAnimationEnd', open: false, spotlight: 'container', spotlightModal:true, components: [
-			{name: 'closeButton', kind: IconButton, icon: 'closex', classes: 'moon-popup-close moon-list-actions-close moon-neutral', ontap: 'expandContract', accessibilityLabel: $L('Close'), backgroundOpacity: 'transparent', defaultSpotlightDown:'listActions'},
+			{name: 'closeButton', kind: IconButton, icon: 'arrowsmallup', classes: 'moon-popup-close moon-list-actions-close moon-neutral', ontap: 'expandContract', accessibilityLabel: $L('Close'), backgroundOpacity: 'transparent', defaultSpotlightDown:'listActions'},
```
aarontam: More minor: curious if we had considered making this `arrowsmalldown` to match the close animation direction.

Djspaceg: Up was specifically chosen for the direction of the list actions animation direction.

aarontam: Ah never mind! DyamicPanelsSample is way outdated... :disappointed:

PR #2567, opened 2015-09-24
ENYO-2277: Scroll to top/bottom/left/right when focus leave from scroller.

KunmyonChoi:
Issue:
If controls at the top or bottom (left or right also) are disabled inside of scroller, there
is no way to show them by 5way key.

Fix:
We force scroll to top or bottom (left or right also) on focus leave based on last 5way event.

### DIFF
--- a/src/Scroller/Scroller.js
+++ b/src/Scroller/Scroller.js
@@ -361,7 +361,9 @@ if (platform.touch) {
 		containerLeaveHandler: function () {
 			var strategy = this.getStrategy(),
 				b = strategy.getScrollBounds(),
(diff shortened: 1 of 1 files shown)


KunmyonChoi:
Hi @aarontam , thanks for the review. Updated per comments.

KunmyonChoi:
One difference on adding filter for originator is disablePageUpDownKey is called everytime, but now call only when originator is same this. I don't know how much difference will be happen with this change.

clinuz:
Closing as we will not be accepting this PR for this issue.

PR #2570, opened 2015-09-25
Enyo 2587 Apply ri.scale to adjust tooltip's position

DavidUM:
Issue
------
Tooptip has wider gap in HD resolution

Cause
--------
Some default properties has a constant value

Fix
----
Apply ri.scale to those properties

### DIFF
--- a/src/Tooltip/Tooltip.js
+++ b/src/Tooltip/Tooltip.js
@@ -351,7 +351,9 @@ module.exports = kind(
 		if (this.showing && this.hasNode()) {
 			var b = this.node.getBoundingClientRect(),
 				moonDefaultPadding = 20,
+				moonDefaultPaddingRem = ri.scale(moonDefaultPadding),
 				defaultMargin = 15,
+				defaultMarginRem = ri.scale(defaultMargin),
 				pBounds = this.parent.getAbsoluteBounds(),
 				acBounds = null;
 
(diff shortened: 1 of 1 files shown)


Djspaceg:
We should not be long-term storing a scaled value. This should be updated to live-calculate each time.

src/Tooltip/Tooltip.js
```diff
@@ -351,7 +352,9 @@ module.exports = kind(
 		if (this.showing && this.hasNode()) {
 			var b = this.node.getBoundingClientRect(),
 				moonDefaultPadding = 20,
```
Djspaceg: Wouldn't this be an easier fix than changing all the variables?
```
moonDefaultPadding = ri.scale( 20 ),
```
Also, having "rem" in the name doesn't make sense because we aren't storing "rem" units, but storing scaled pixels instead.

src/Tooltip/Tooltip.js
```diff
@@ -351,7 +352,9 @@ module.exports = kind(
...
+				moonDefaultPaddingRem = ri.scale(moonDefaultPadding),
 				defaultMargin = 15,
```
Djspaceg: Same for here, just the one variable is fine:
```
defaultMargin = ri.scale( 15 );
```

aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2571.

PR #2571, opened 2015-09-28
ENYO-2581: Scale default padding and margin.

aarontam:
### Issue
The spacing between the activator and the tooltip were the same amount of pixels, regardless of resolution.

### Fix
We scale the default padding and margin via `ri.scale`.


Djspaceg:
Success! :+1:

PR #2573, opened 2015-09-29
ENYO-2569: Relocate focus-at-end code from ExpandableInput to Input…

Djspaceg:
…so everyone gets consistent behavior.

### DIFF
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -102,7 +102,10 @@ module.exports = kind(
 		// Force cursor to end of text during a generic focus event. Creating the input by compiling
 		// a string of text with value="this.value" produces different initial caret position than
 		// using node.setAttribute('value', this.value), which is what would happen any time after
-		// the initial creation.
+		// the initial creation. The initial end-position of the caret is required to support
+		// Virtual keyboards because without arrow-keys because normal left/right arrow navigation
+		// in inputs is impossible, so the caret must be positioned at the end to allow for deletion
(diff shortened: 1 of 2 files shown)


src/Input/Input.js
```diff
@@ -99,6 +99,11 @@ module.exports = kind(
 			var oThis = this;
 			util.asyncMethod(this, function () {oThis._bFocused = true;});
 		}
+		// Force cursor to end of text during a generic focus event. Creating the input by compiling
+		// a string of text with value="this.value" produces different initial caret position than
+		// using node.setAttribute('value', this.value), which is what would happen any time after
+		// the initial creation.
+		this.hasNode().selectionStart = this.value.length;
```
graynorton: I think maybe we should retain more of the original comment here, since the stated reason for the original change (to prevent the cursor from being in an untenable position in 5-way mode) seems more important than the consistent-caret-positioning reason stated in the new issue.

Djspaceg: I took it out because it's my understanding that the original comment and assumption was incorrect, at best, incomplete.

Djspaceg: I understand now. Comment updated.

webOS101: This fails for inputs of type 'number'.

"Uncaught InvalidStateError: Failed to set the 'selectionStart' property on 'HTMLInputElement': The input element's type ('number') does not support selection."

PR #2584, opened 2015-10-06
ENYO-2650: It doesn't read complete value of ProgressBar because of c…

jaewonjang:
…hanging fast

Issue
It doesn't read complete value of ProgressBar because of changing fast

Cause
ProgressBar is used on like channel tuning, downloading and so on, the value is changed very fast.
It tried to read every changed value, so it can't read complete value.

Fix
Remove observing popup content and value and only support custom value
through accessibilityValueText.
If developer want to read something during on progress, can set
accessibilityValueText to read it.

https://jira2.lgsvl.com/browse/ENYO-2650

### DIFF
--- a/src/ProgressBar/ProgressBar.js
+++ b/src/ProgressBar/ProgressBar.js
@@ -777,6 +777,15 @@ module.exports = kind(
 	accessibilityRole: 'progressbar',
 
 	/**
+	* Custom value for accessibility (ignored if `null`).
+	*
+	* @type {String|null}
+	* @default null
+	* @public
+	*/
+	accessibilityValueText: null,
+
+	/**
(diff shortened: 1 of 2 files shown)


ryanjduffy:
Please resolve the merge conflict (likely caused by merging the other Slider a11y change). Also, we should address `accessibilityValueText` in Slider as well since it will inherit the property from ProgressBar but will ignore it in its `ariaValue()`.

jaewonjang:
The code handling accessibilityValueText in Slider already merged. In Slider if accessibilityValueText is set, it just return in ariaValue, and handle it in the function observing accessibilityValueText.

jaewonjang:
replaced with https://github.com/enyojs/moonstone/pull/2606 to fix merge conflict

PR #2585, opened 2015-10-06
ENYO-2651: Deburred the Spinner Balls

Djspaceg:
Due to low-quality anti-aliasing settings, the spinner had an occasional extra pixel where it visually probably shouldn't. This change subtracts 1px from the size of the spinner to alter the Blink compositor and trick it into rendering a higher-quality edge for each ball, eliminating the blip. 🙅


aarontam:
@Djspaceg Excellent PR title :+1:

PR #2587, opened 2015-10-08
ENYO-2666: upadate small header's title to apply allowhtml properly

sbcho:
## issue
Title and titleBelow are always {allowHtml:true} in small header even header's allowHtml property is true

## fix
Based on header's allowHtml property, escape title and titleBelow string before set title of small header.

Enyo-DCO-1.1-Signed-off-by: Sungbae Cho sb.ch@lge.com

### DIFF
--- a/src/Header/Header.js
+++ b/src/Header/Header.js
@@ -676,6 +676,8 @@ module.exports = kind(
 					: (this.get('title') || this.get('content')),
 			subtitle = this.get('titleBelow');
 		if ((this.get('type') == 'small') && subtitle) {
+			title = this.allowHtml? title: dom.escape(title);
+			subtitle = this.allowHtml? subtitle: dom.escape(subtitle);
 			this.$.title.set('allowHtml', true);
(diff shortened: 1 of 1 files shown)


src/Header/Header.js
```diff
@@ -676,6 +676,8 @@ module.exports = kind(
 					: (this.get('title') || this.get('content')),
 			subtitle = this.get('titleBelow');
 		if ((this.get('type') == 'small') && subtitle) {
+			title = this.allowHtml? title: dom.escape(title);
+			subtitle = this.allowHtml? subtitle: dom.escape(subtitle);
```
dragongears: Is there a reason for using a ternary operation on these two lines instead of just a regular 'if' statement? If this.allowHtml is true then they set title = title and subtitle = subtitle. Why not `if (!this.allowHtml) {title =  dom.escape(title)}` ?

Also, as an aside, the ternary operations should probably have a space before the question marks and colons.

sbcho: Thanks for your comment. I'll update it

PR #2592, opened 2015-10-14
DRD4TV-3571: Added custom font support for en-JP locale (English in Japan)

Djspaceg:

### DIFF
--- a/src/fonts.js
+++ b/src/fonts.js
@@ -55,7 +55,7 @@ function funLocaleSpecificFonts () {
 			'ja': {
 				regular: 'LG Display_JP',
 				unicodeRanges:
-					'U+0000-U+00FF, ' +
+					'U+0000-U+00FF,' +
 					'U+2E80-U+2EFF,' +
 					'U+2F00-U+2FDF,' +
 					'U+3000-U+303F,' +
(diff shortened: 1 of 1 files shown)


src/fonts.js
```diff
@@ -55,37 +55,51 @@ function funLocaleSpecificFonts () {
...
-					'U+3000-U+303F,' +
-					'U+3040-U+309F,' +
-					'U+30A0-U+30FF,' +
-					'U+3200-U+33FF,' +
-					'U+3400-U+4DBF,' +
-					'U+4E00-U+9FFF,' +
-					'U+E000-U+FAFF,' +
-					'U+FF00-U+FFEF'
+					'U+0-FF,' +
+					'U+2E80-2EFF,' +
```
yeram: Hi @Djspaceg, just curious, is it same behavior with excluding 'U+' from the end range?

webOS101: https://drafts.csswg.org/css-fonts-3/#descdef-font-face-unicode-range

It looks like it was wrong before.

Djspaceg: Yes @yeram, it was wrong before.
We fixed the glitch

yeram:
Got it, thanks!

PR #2593, opened 2015-10-14
ENYO-2676 IntegerPicker 'range' does not update when 'step' is changed

dragongears:

### DIFF
--- a/src/IntegerPicker/IntegerPicker.js
+++ b/src/IntegerPicker/IntegerPicker.js
@@ -341,10 +341,17 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	rangeChanged: function () {
-		this.verifyValue();
+	prepareRange: function () {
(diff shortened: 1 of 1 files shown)


webOS101:
Not super happy with all the code in IntergerPicker but now that I understand the fix a bit better I'm happy enough to merge!

src/IntegerPicker/IntegerPicker.js
```diff
@@ -341,13 +341,20 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	rangeChanged: function () {
-		this.verifyValue();
+	prepareRange: function () {
 		this.range = this.valueToIndex(this.max) - this.valueToIndex(this.min) + 1;
```
viodragon2: `this.valueToIndex(this.min)` is always 0. We may want to take this away.

webOS101: Good catch!  There is a lot to clean up in IntegerPicker.

PR #2598, opened 2015-10-16
ENYO-2643: Each Scrollable needs its own options!

graynorton:
Oops... `scrollIntoViewOptions` were being specified as an object
literal in the Scrollable mixin definition, which meant they were
inadvertently being shared by all instances. We'll initialize them
in the `create()` method instead.

### DIFF
--- a/src/Scrollable/Scrollable.js
+++ b/src/Scrollable/Scrollable.js
@@ -60,15 +60,14 @@ var Scrollable = {
 	*/
 	create: kind.inherit(function (sup) {
 		return function () {
+			var opts = {
+				block: 'farthest',
+				behavior: 'smooth'
+			};
+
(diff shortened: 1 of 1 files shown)


src/Scrollable/Scrollable.js
```diff
@@ -53,7 +61,15 @@ var Scrollable = {
 	create: kind.inherit(function (sup) {
 		return function () {
 			sup.apply(this, arguments);
-			// Save default options so they can be restored after runtime changes
+
+			if (!this.scrollIntoViewOptions) {
+				this.scrollIntoViewOptions = {
```
ryanjduffy: Should the user-provided value be merged with the defaults or do they have to specify both values?

PR #2600, opened 2015-10-19
ENYO-2701: ListActions is now aware of when its parent panel is onscreen

Djspaceg:
So it can execute necessary changes that may have happened at a higher level without its knowledge.


Djspaceg:
`showCloseButton` is reserved for future usage.

webOS101:
No reason to leave dead code around in the source.

PR #2614, opened 2015-10-28
ENYO-2746 Adjust margin top for sub-title to prevent overlapping of c…

DavidUM:
…ontents.

Current CSS make an overlapped area in title and sub-title

### DIFF
--- a/src/Header/Header.less
+++ b/src/Header/Header.less
@@ -83,9 +83,6 @@
 		.moon-header-title-below,
 		.moon-header-sub-title-below {
 			line-height: 39px;
-			.enyo-locale-non-latin & {
-				line-height: @moon-non-latin-medium-header-line-height;
-			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Djspaceg:
@DavidUM That change was made to prevent subTitleBelow from falling off the bottom of the header when the title's line height is taller.

src/Header/Header.less
```diff
@@ -87,6 +83,9 @@
 		.moon-header-title-below,
 		.moon-header-sub-title-below {
 			line-height: 39px;
+			.enyo-locale-non-latin & {
+				line-height: @moon-non-latin-medium-header-line-height;
```
Djspaceg: Why set the `@moon-non-latin-medium-header-line-height` to the `.moon-header-title-below` and `.moon-header-sub-title-below` classes?

DavidUM: Because test letter ''First Panel  ฟิ้  ไั  ஒ  து'' could not display property with 39px;
They are taller than 39px

Djspaceg: Sorry @DavidUM, I should have been more clear. I think it's inappropriate to use the *medium header's* line height to set the line height of the `title-below` and `sub-title-below`. They aren't logically related and don't even share the same font.

PR #2615, opened 2015-10-28
ENYO-2736 Don't call unspot() manually in pointerMode

DavidUM:
Issue
-----
When we select item of ExpandablePicker with pointerMode,
we lose spotlight focus if we do not move mouse cursor.

Cause
-------
{{closeDrawerAndHighlightHeader}} unspot its focus manually and close picker
Due to that, it does not execute onDisappear routine of Spotlight.

Fix
-----
Remove unspot(). 
Selected item will lose its focus naturally according to disappear routine of Spotlight


ryanjduffy:
Closing in favor of #2625

PR #2643, opened 2015-11-10
ENYO-2789: Add ariaObserver for preventing reading a password type in…

kbs12e:
## Issue 
It reads the real value of password in ExpandableInput

## Cause
ExpandableInput binds client input value to header ( LabeledTextItem ) text, so screen reader reads this text and accessibilityHint ( 'Text Input' ). 
 
## Fix
Set _accessibilityText to 'password' if input type is password, because screen reader reads _accessibilityText value instead of header text

https://jira2.lgsvl.com/browse/ENYO-2789


seunghoh:
The parent of header dom has an unused aria-label

PR #2648, opened 2015-11-11
ENYO-2790: Accessibility: The slider value does not change while no p…

jaewonjang:
…opup.

Issue
The slider value does not change while no popup.

Cause
After the popup property is changed from true to false, both
'aria-valuetext' and 'aria-valuenow' are exist.
Web engine reads only 'aria-valuetext' if two of them are in a dom node.

Fix
Before setting 'aria-valuetext' or 'aria-valuenow', set them to null.

https://jira2.lgsvl.com/browse/ENYO-2790

### DIFF
--- a/src/ProgressBar/ProgressBar.js
+++ b/src/ProgressBar/ProgressBar.js
@@ -793,6 +793,8 @@ module.exports = kind(
 		// of the label determination could help here - rjd
 		{path: ['progress', 'popup', '$.popupLabel.content'], method: 'ariaValue'},
 		{path: ['accessibilityValueText'], method: function () {
+			this.setAriaAttribute('aria-valuetext', null);
+			this.setAriaAttribute('aria-valuenow', null);
 			this.setAriaAttribute('aria-valuetext', this.accessibilityValueText);
 		}}
 	],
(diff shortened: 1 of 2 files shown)


src/ProgressBar/ProgressBar.js
```diff
@@ -793,6 +793,8 @@ module.exports = kind(
 		// of the label determination could help here - rjd
 		{path: ['progress', 'popup', '$.popupLabel.content'], method: 'ariaValue'},
 		{path: ['accessibilityValueText'], method: function () {
+			this.setAriaAttribute('aria-valuetext', null);
+			this.setAriaAttribute('aria-valuenow', null);
```
ryanjduffy: We shouldn't null `aria-valuenow` here because that could result in no a11y value if `accessibilityValueText` were changed from a valid value to `null`. Also, there's no need to null `aria-valuetext` just before setting it to another value as it'll incur an extra fn call and DOM access.

src/Slider/Slider.js
```diff
@@ -999,6 +999,12 @@ module.exports = kind(
 		this.set('accessibilityRole', !this.enableJumpIncrement ? 'spinbutton' : null);
 		this.set('accessibilityLive', null);
 		this.set('accessibilityHint', null);
+		this.setAriaAttribute('aria-valuetext', null);
+		this.setAriaAttribute('aria-valuenow', null);
+		if (this.$.slider) {
```
ryanjduffy: For consistency, unless you had a specific reason otherwise let's check against `enableJumpIncrement` as we've done above.

src/ProgressBar/ProgressBar.js
ryanjduffy: Actually, it's probably better to collapse the `accessibilityValueText` observer into `ariaValue()`. We could then drop the observer in moonstone/Slider and just override `ariaValue()` there.

PR #2654, opened 2015-11-13
Fixes ENYO-2837

ryanjduffy:
Alter the CSS of LabeledTextItem without text to match Item sizing
Move logic into observer from binding transform and remove showing
logic to prevent overriding ExpandableListItemHeader's textShowing
Update ExpandableListItem to override .with-text for specificity

Issue: ENYO-2837

### DIFF
--- a/src/ExpandableListItem/ExpandableListItem.less
+++ b/src/ExpandableListItem/ExpandableListItem.less
@@ -3,7 +3,10 @@
 	// Header
 	.moon-labeledtextitem-header {
 		.padding-start-end(0, @moon-spotlight-outset + 30px);
-		margin-bottom: 0px;
+		
+		&.with-text {
+			margin-bottom: 0px;
+		}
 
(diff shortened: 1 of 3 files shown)


src/ExpandableListItem/ExpandableListItem.less
```diff
@@ -3,7 +3,10 @@
...
 	.moon-labeledtextitem-header {
 		.padding-start-end(0, @moon-spotlight-outset + 30px);
-		margin-bottom: 0px;
+
+		&.with-text {
+			margin-bottom: 0;
+		}
 
 		&:after {
 			position: absolute;
```
Djspaceg: I did not change the line below: `top: @moon-spotlight-outset - 12px;` since the 12px used to here doesn't refer to the spotlight-outset, but rather the vertical offset of the icon/glyph, which may be different from spotlight-outset.

PR #2661, opened 2015-11-17
ENYO-2850: update multiSelectCurrentValue function to handle wrong odrder

sbcho:
## Issue
In some case, currentValueText is different from header text because selectedIndex's oder is changed sunddenly. (header text is right result)

## Fix
Update multiSelectCurrentValue function to handle wrong selectedIndex's odrder

Enyo-DCO-1.1-Signed-off-by: Sungbae Cho (sb.cho@lge.com)

### DIFF
--- a/src/DayPicker/DayPicker.js
+++ b/src/DayPicker/DayPicker.js
@@ -211,7 +211,7 @@ module.exports = kind(
 		if (str) {
 			return str;
 		}
-
+		this.selectedIndex.sort();
 		for (var i=0; i < this.selectedIndex.length; i++) {
 			if (!str) {
 				str = this.days[this.selectedIndex[i]];


src/DayPicker/DayPicker.js
```diff
@@ -211,7 +211,7 @@ module.exports = kind(
 		if (str) {
 			return str;
 		}
-
+		this.selectedIndex.sort();
```
aarontam: I realize that this line exists in `ExpandablePicker` as well, but I'm curious if we know how the ordering of the selected indices becomes incorrect? It would be nice to avoid sorting here, in general, if we can correct the core logic.

sbcho: Yes, indeed, I didn't insert this line because I thought there is no case that selectedIndex's order is changed. I also would like to know when and why selectedIndex is changed even rebuildSelectedIndices is called.

PR #2666, opened 2015-11-20
ENYO-2866: [LightPanel] during back key process, spotlight disappears

suhlee:
## Issue

During back key process, there is a postTransition step. There is a checkSpottability, but there is nothing to spot.

## Fix

we added more condition to spot active panel
(!current.isDescendantOf(this) -> for current is not control in active panel(e.g. control in deactivated panel - previous panel)
current.isDescendantOf(this.owner) -> for control outside panels. In this case we do not need to spot active panel

Enyo-DCO-1.1-Signed-off-by: Suhyung.Lee (suhyung2.lee@lge.com)


aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2668.

PR #2672, opened 2015-12-09
PLAT-10884: Remodify Accessibility string for multi language support

jaewonjang:
We had modified temporarily accessibility string with substitution string
witch has been translated already.
Now we can request translation for 2016 TV MR and restore it to the original
string.

https://jira2.lgsvl.com/browse/PLAT-10884


jaewonjang:
I'll create new PR to fix conflict

PR #2677, opened 2015-12-30
ENYO-2784: When the LightPanels are switched, screen reader doesn't read the focused item.

jaewonjang:
Fix for ENYO-2784 and ENYO-2485

https://jira2.lgsvl.com/browse/ENYO-2784?filter=-1


seunghoh:
Close as outdated PR

PR #2678, opened 2015-12-31
ENYO-2907: Ignore text directionality of CheckboxItem

viodragon2:
Issue
---
Checkbox icon is right aligned when contents are RTL text even in LTR locale. Its base kind, `moon/Item` which mixins MarqueeItem, checks for text directionality and sets its `direction` style according to its text. We only need to run that against client's MarqueeItem and not the entire CheckboxItem.

Fix
--- 
Ignore text directionality for CheckboxItem.

### DIFF
--- a/src/CheckboxItem/CheckboxItem.js
+++ b/src/CheckboxItem/CheckboxItem.js
@@ -227,9 +227,7 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	detectTextDirectionality: function () {
-		this.applyStyle('direction', null);
-	},
+	detectTextDirectionality: function () {},
 
 	/**
 	* @private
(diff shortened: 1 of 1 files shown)


src/CheckboxItem/CheckboxItem.js
```diff
@@ -227,6 +227,13 @@ module.exports = kind(
 	/**
 	* @private
 	*/
+	detectTextDirectionality: function () {
+		this.applyStyle('direction', null);
```
aarontam: Curious if we need to execute `applyStyle`, or if we could just make this method a no-op.

viodragon2: no-op would do!

ryanjduffy:
Replaced by #2696

PR #2680, opened 2016-01-08
ENYO-2992: Allow for horizontal scrolling when there is also vertical scrolling.

aarontam:
### Issue
This change might be wholly unnecessary if we do not plan on supporting horizontal scroll wheels for Moonstone. That being said, horizontal scrolling does not function properly if there is also vertical scrolling enabled (i.e. in the Scroller2DSample). The scroll position resets to the origin in the horizontal plane in this case.

### Fix
This occurred because we were not computing the value of `x` in the case where the `event.wheelDeltaX` value was 0 and we were allowing for vertical scrolling. The conditionals have been tweaked so that this case is handled.


aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2682.

PR #2689, opened 2016-01-26
ENYO-3036: End hold job when control is blurred.

aarontam:
### Issue
Due to some refactoring for accessibility, we separated out the logic for handling `onSpotlightKeyUp` and `onSpotlightBlur`. As a result, the call to `endHold` was only being made when handling `onSpotlightKeyUp`, and pressing a 5-way key to move off of the held `moonstone/PagingControl` control did not stop the scrolling.

### Fix
We have added a call to `endHold` when handling `onSpotlightBlur`, and have moved/consolidated the clean-up logic into the `endHold` method.

### DIFF
--- a/src/PagingControl/PagingControl.js
+++ b/src/PagingControl/PagingControl.js
@@ -244,7 +244,6 @@ module.exports = kind(
 	undepress: function (sender, event) {
 		this.set('pressed', false);
 		IconButton.prototype.undepress.apply(this, arguments);
-		this.downCount = 0;
 		this.endHold(sender, event);
 	},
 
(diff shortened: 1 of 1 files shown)


src/PagingControl/PagingControl.js
```diff
@@ -256,6 +255,7 @@ module.exports = kind(
 			return;
```
webOS101: Is there any reason we're receiving the sender, event here?  Should we clean this up and revert the spotlightBlurred below?

aarontam: That crossed my mind, but I decided to preserve it as it was pre-existing. We would still need the call to `endHold` in `spotlightBlurred`, regardless. I guess the only case where this would break something would be if someone had subkinded `PagingControl` and had come to rely on the parameters sent to their override of `endHold`.

webOS101: Reasonable!  Let's leave it as is.

PR #2691, opened 2016-01-28
PLAT-15389: Add back key support to Expandables.

aarontam:
### Issue
We want to support the back key in `Expandable*` controls (basically, anything that subkinds `moonstone/ExpandableListItem`), such that pressing the back key closes the control if it is open and has Spotlight focus.

### Fix
We have added this functionality primarily to `moonstone/ExpandableListItem`. Specifically, we have converted this control to be a Spotlight container to allow for easier handling of both the header + drawer as a whole, and we maintain an internal property to ensure we only add/remove history entries as appropriate. We have also added some logic to `moonstone/DateTimePickerBase` and `moonstone/ExpandableInput` to reset their respective values when the back key is pressed.

### Notes
I considered adding an API called `resetOnBack` (or something to that effect) that would be defaulted to `false`, and set to `true` for `moonstone/DateTimePickerbase` and `moonstone/ExpandableInput`. This would allow for the reset logic to be written once in `moonstone/ExpandableListItem`'s `backKeyHandler`, but in the interest of being more opinionated and not unnecessarily expanding the API surface area, have decided against it, especially because the shared logic is a single call. Also, the setting of the initial value varies between the controls and was not easily generalizable. I can probably be convinced otherwise, though.

This PR is dependent on https://github.com/enyojs/enyo/pull/1366.

### DIFF
--- a/src/DateTimePickerBase/DateTimePickerBase.js
+++ b/src/DateTimePickerBase/DateTimePickerBase.js
@@ -263,9 +263,13 @@ module.exports = kind(
 	* @private
 	*/
 	expandContract: function () {
-		// if currently closed and without a value, set it to now before opening
-		if (!this.open && !this.value) {
-			this.set('value', new Date());
+		if (!this.open) {
+			// if currently closed and handling the back key, keep track of the initial value in
+			// case we need to restore this value
+			if (this.allowBackKey) this._initialValue = this.value && new Date(this.value.getTime());
+
+			// if currently closed and without a value, set it to now before opening
+			if (!this.value) this.set('value', new Date());
 		}
 		ExpandableListItem.prototype.expandContract.apply(this, arguments);
 	},
@@ -306,15 +310,23 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	refresh: function (){
+	refresh: function () {
 		this.destroyClientControls();
 		this.pickers = null;
 		delete this._tf;
 		if (this.value) {
-			this.localeValue = dateFactory({unixtime: this.value.getTime(), timezone: "local"});
+			this.localeValue = dateFactory({unixtime: this.value.getTime(), timezone: 'local'});
 		}
 		this.set('open', false);
 		this.initDefaults();
 		this.render();
+	},
+
+	/**
+	* @private
+	*/
+	backKeyHandler: function () {
+		this.set('value', this._initialValue);
+		ExpandableListItem.prototype.backKeyHandler.apply(this, arguments);
 	}
 });
--- a/src/DateTimePickerBase/DateTimePickerBase.js
+++ b/src/DateTimePickerBase/DateTimePickerBase.js
@@ -209,10 +209,7 @@ module.exports = kind(
 		this.syncingPickers = true;
 		this.setChildPickers(inOld);
 		this.syncingPickers = false;
-
-		if (this.value) {
-			this.doChange({name:this.name, value:this.value});
-		}
+		this.doChange({name:this.name, value:this.value});
 	},
 
 	/**
--- a/src/DateTimePickerBase/DateTimePickerBase.js
+++ b/src/DateTimePickerBase/DateTimePickerBase.js
@@ -326,7 +326,7 @@ module.exports = kind(
 	* @private
 	*/
 	backKeyHandler: function () {
-		this.set('value', this._initialValue);
+		if (this.open) this.set('value', this._initialValue);
 		ExpandableListItem.prototype.backKeyHandler.apply(this, arguments);
 	}
 });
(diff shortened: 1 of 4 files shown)


src/ExpandableInput/ExpandableInput.js
```diff
@@ -213,10 +216,12 @@ module.exports = kind(
...
 	inputSpotBlurred: function (inSender, inEvent) {
-		var eventType;
+		var eventType, lastEvent;
 		if (this.open && Spotlight.getPointerMode()) {
-			eventType = Spotlight.getLastEvent().type;
-			if (eventType !== 'onSpotlightFocus' && eventType !== 'mouseover') {
+			lastEvent = Spotlight.getLastEvent();
+			eventType = lastEvent.type;
+			if (eventType !== 'onSpotlightFocus' && eventType !== 'mouseover'
+				&& (!lastEvent.dispatchTarget || !lastEvent.dispatchTarget.isDescendantOf(this))) {
```
ryanjduffy: Why was this condition added?

src/ExpandableListItem/ExpandableListItem.js
```diff
@@ -9,18 +9,23 @@ var
...
 var
+	HistorySupport = require('../HistorySupport'),
 	Item = require('../Item');
 
 var
 	ExpandableListItemHeader = require('./ExpandableListItemHeader'),
 	ExpandableListItemDrawer = require('./ExpandableListItemDrawer');
 
+var
+	hasHistoryEntry = false; // used to track whether or not we have pushed a history entry for ourselves
```
ryanjduffy: Not sure this should be shared. What was the rationale for it here instead of an instance member?

src/ExpandableInput/ExpandableInput.js
aarontam: Good catch - this is cruft from trying to handle an `ExpandableInput` edge case that went away by keeping track of an instance's history status, rather than trying to assume we can rely on all of the add/remove cases always pairing up.

src/DateTimePickerBase/DateTimePickerBase.js
```diff
@@ -306,15 +310,23 @@ module.exports = kind(
...
 		this.set('open', false);
 		this.initDefaults();
 		this.render();
+	},
+
+	/**
+	* @private
+	*/
+	backKeyHandler: function () {
+		if (this.open) this.set('value', this._initialValue);
```
webOS101: This needs to fire an onChange event if the value has changed

src/ExpandableListItem/ExpandableListItem.js
```diff
@@ -373,6 +420,27 @@ module.exports = kind(
...
+		this.hasHistoryEntry = false;
+
+		// In the case where Spotlight focus is not on one of the items in the expandable, but there
+		// was still an entry history from this control, we must be in a situation where the pointer
+		// has moved away from the control and we have yet to spot another item. We should then
+		// effectively "pass on" the back action by calling the "pop" method of History.
+		if (current && current.isDescendantOf(this)) {
+			this.closeDrawerAndHighlightHeader();
+		} else {
+			EnyoHistory.pop();
```
webOS101: Should we maybe just let this bubble out?  (not return true)

aarontam: No, the default behavior of `HistorySupport` will cause the control to hide if we don't indicate we have handled the event.

src/DateTimePickerBase/DateTimePickerBase.js
aarontam: Good catch! Let me tinker with this...

src/ExpandablePicker/ExpandablePicker.js
```diff
@@ -466,6 +466,11 @@ module.exports = kind(
 		}
 
 		if (this.getAutoCollapseOnSelect() && this.$.drawer.hasRendered && this.getOpen()) {
+			// this is needed due to the select and close delay, during which other history entries
+			// could have been pushed - we drop the entry immediately to prevent dropping the wrong
+			// entry later
+			this.removeHistoryEntry();
```
webOS101: Should we check `this.hasHistoryEntry` first?

aarontam: This property is being checked here: https://github.com/enyojs/moonstone/pull/2691/files#diff-a273ea1fb0277d3dc6b4ae55158e2c63R391

src/DateTimePickerBase/DateTimePickerBase.js
aarontam: There is a long-standing guard against firing the change event for a falsy value via this commit: https://github.com/enyojs/moonstone/commit/a721bdec0cc124bae8b5e2cf72e3437e5d05b432. It seems like this guard is no longer necessary as we are no longer handling the `onChange` event ourselves, so I've removed it to address your concern. Let me know what you think!

PR #2696, opened 2016-02-03
Fixes ENYO-2907

ryanjduffy:
## Issue
When the locale was LTR but the CheckboxItem's text directionality is RTL, the checkbox item would
be right-aligned causing it to overlap with the content.

## Fix
Use `right: auto` for left-handed checkboxes to override the default `right` value set for the checkbox.

### DIFF
--- a/src/CheckboxItem/CheckboxItem.less
+++ b/src/CheckboxItem/CheckboxItem.less
@@ -19,6 +19,7 @@
 		// Checkbox
 		.moon-checkbox {
 			left: @moon-spotlight-outset - 3;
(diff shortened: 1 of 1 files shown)


src/CheckboxItem/CheckboxItem.less
```diff
@@ -19,6 +19,7 @@
 		// Checkbox
 		.moon-checkbox {
 			left: @moon-spotlight-outset - 3;
+			right: auto;
```
aarontam: (nice) :+1:

PR #2702, opened 2016-02-09
ENYO-2954: API and layout for Dialog established

Djspaceg:
* Title, SubTitle, and Divider are all optional now. Buttons relocated to the bottom.
* Popup had a minor change to provide safe-padding around the X-Close button (things were running into it)
* Layout updated with the latest sizing/positioning specs from GUI.

### DIFF
--- a/src/Dialog/Dialog.less
+++ b/src/Dialog/Dialog.less
@@ -1,30 +1,27 @@
 /* Dialog.css */
 .moon-dialog {
-	padding: 24px 42px;
+	padding-bottom: 0; // Override popup, so our optional components won't be stuck with too much padding
+
+	.moon-dialog-title-wrapper,
+	.moon-dialog-content,
+	.moon-dialog-client {
+		margin-bottom: 39px;
+	}
 
 	.moon-dialog-title-wrapper {
-		margin: @moon-spotlight-outset 0;
-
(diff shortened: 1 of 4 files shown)


src/Dialog/Dialog.less
```diff
@@ -1,34 +1,27 @@
...
+	padding-bottom: 0; // Override popup, so our optional components won't be stuck with too much padding
+
+	.moon-dialog-title-wrapper,
+	.moon-dialog-content,
+	.moon-dialog-client {
+		margin-bottom: 39px;
+	}
+
+	.moon-dialog-title-wrapper {
+		&.use-divider {
```
otisnate: @Djspaceg , do you want not to use```divider``` directly? If yes, this approach could break css dependency between ```divider``` and dialog's ```divider```

Curent value of @moon-divider-border-width is ```3px```  (see https://github.com/enyojs/moonstone/blob/2.6.0-dev/css/moonstone-variables.less#L265)

Djspaceg: Correct. We no longer want to use moon.Divider here. Dialog will now no longer require Divider. It's just an optional line now. Also, due to new GUI specs, they'd prefer this line to be 1px, not 3px.

PR #2704, opened 2016-02-12

KunmyonChoi:
### Issue:

The components outside the contextual popup were not spottable even after the spotlightModal property set to false for the contextual popup. The spotlightModal property doesn't use for show/hide the scrim which caused this issue.

### Fix:

Enyo-DCO-1.1-Signed-off-by: Kunmyon Choi kunmyon.choi@lge.com

### DIFF
--- a/src/ContextualPopup/ContextualPopup.js
+++ b/src/ContextualPopup/ContextualPopup.js
@@ -572,7 +572,7 @@ module.exports = kind(
 			} else {
 				this._scrim.hideAtZIndex(this._scrimZ);
 			}
-			util.call(scrim, 'addRemoveClass', [this.scrimClassName, this._scrim.showing]);
+			util.call(this._scrim, 'addRemoveClass', [this.scrimClassName, this._scrim.showing]);
 		} else {
 			// clean up scrim here when showHideScrim is not called from showingChanged
 			this._scrim && this._scrim.hideAtZIndex(this._scrimZ);
--- a/src/ContextualPopup/ContextualPopup.js
+++ b/src/ContextualPopup/ContextualPopup.js
@@ -239,8 +239,7 @@ module.exports = kind(
 	*/
 	syncProperties: function () {
 		if (this.modal == this.spotlightModal) return;
-		if (this.hasOwnProperty('spotlightModal')) this.modal = this.spotlightModal;
-		if (this.hasOwnProperty('modal')) this.spotlightModal = this.modal;
+		this.modal = this.spotlightModal;
 	},
 
 	/**
@@ -574,21 +573,24 @@ module.exports = kind(
 	* @private
 	*/
 	showHideScrim: function (show) {
-		if (this.floating && (this.scrim || (this.modal && this.scrimWhenModal))) {
-			this._scrim = this.getScrim();
-			if (show) {
-				// move scrim to just under the popup to obscure rest of screen
-				var i = this.getScrimZIndex();
-				this._scrimZ = i;
-				this._scrim.showAtZIndex(i);
-			} else {
-				this._scrim.hideAtZIndex(this._scrimZ);
-			}
+		var scrim = this.getScrim();
+
+		if (this._scrim && scrim != this._scrim) {
+			// hide if there was different kind of scrim
+			this._scrim.hideAtZIndex(this._scrimZ);
 			util.call(this._scrim, 'addRemoveClass', [this.scrimClassName, this._scrim.showing]);
-		} else {
-			// clean up scrim here when showHideScrim is not called from showingChanged
-			this._scrim && this._scrim.hideAtZIndex(this._scrimZ);
 		}
+
+		if (show && this.floating && (this.scrim || (this.modal && this.scrimWhenModal))) {
+			// move scrim to just under the popup to obscure rest of screen
+			var i = this.getScrimZIndex();
+			this._scrimZ = i;
+			scrim.showAtZIndex(i);
+		} else {
+			scrim.hideAtZIndex(this._scrimZ);
+		}
+		util.call(scrim, 'addRemoveClass', [this.scrimClassName, scrim.showing]);
+		this._scrim = scrim;
 	},
 
 	/**
--- a/src/ContextualPopup/ContextualPopup.js
+++ b/src/ContextualPopup/ContextualPopup.js
@@ -125,14 +125,14 @@ module.exports = kind(
 
 		/**
 		* If `true`, focus cannot leave the constraints of the popup unless the
-		* popup is explicitly closed. This property is copied to
-		* [modal]{@link module:enyo/Popup~Popup#modal} on initiation time.
-		* And, these two proverties are synced whenever one of
+		* popup is explicitly closed. This property's value is copied to
+		* [modal]{@link module:enyo/Popup~Popup#modal} at initialization time.
+		* Additionally, these two properties are synced whenever one of the following properties changes:
 		* [spotlightModal]{@link module:moonstone/ContextualPopup~ContextualPopup#spotlightModal},
 		* [modal]{@link module:enyo/Popup~Popup#modal},
 		* [modal]{@link module:enyo/Popup~Popup#scrim},
 		* [modal]{@link module:enyo/Popup~Popup#scrimWhenModal},
-		* [modal]{@link module:enyo/Popup~Popup#floating} property changes.
+		* [modal]{@link module:enyo/Popup~Popup#floating}.
 		*
 		* @type {Boolean}
 		* @default false
--- a/src/ContextualPopup/ContextualPopup.js
+++ b/src/ContextualPopup/ContextualPopup.js
@@ -125,7 +125,14 @@ module.exports = kind(
 
 		/**
 		* If `true`, focus cannot leave the constraints of the popup unless the
-		* popup is explicitly closed.
+		* popup is explicitly closed. This property is copied to
+		* [modal]{@link module:enyo/Popup~Popup#modal} on initiation time.
+		* And, these two proverties are synced whenever one of
+		* [spotlightModal]{@link module:moonstone/ContextualPopup~ContextualPopup#spotlightModal},
+		* [modal]{@link module:enyo/Popup~Popup#modal},
+		* [modal]{@link module:enyo/Popup~Popup#scrim},
+		* [modal]{@link module:enyo/Popup~Popup#scrimWhenModal},
+		* [modal]{@link module:enyo/Popup~Popup#floating} property changes.
 		*
 		* @type {Boolean}
 		* @default false
@@ -229,16 +236,6 @@ module.exports = kind(
 	initComponents: function () {
 		this.createChrome(this.tools);
 		Popup.prototype.initComponents.apply(this, arguments);
-		this.syncProperties();
-	},
-
-	/**
-	* Sync spotlightModal and modal properties on creation time.
-	*
-	* @private
-	*/
-	syncProperties: function () {
-		if (this.modal == this.spotlightModal) return;
 		this.modal = this.spotlightModal;
 	},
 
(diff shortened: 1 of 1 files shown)


KunmyonChoi:
FYI, these are related commits.
https://github.com/enyojs/moonstone/commit/6e4df4c7dc819b73b2b261d03d4dccd1b1d4bebb 
https://github.com/enyojs/enyo/commit/96a39d0057c1652af6455ee2129beb47745bacc6

src/ContextualPopup/ContextualPopup.js
```diff
@@ -538,18 +545,37 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	showHideScrim: function (inShow) {
+	updateScrim: function (old, value, source) {
+		if (this._updateScrimMutex) return;
```
KunmyonChoi: Hi @aarontam, 
I think I can have two choices for preventing multiple re-entrance.

1) Use mutex "this._updateScrimMutex"
2) Use assign "this.modal = this.spotlightModal;"

Which one do you prefer?

src/ContextualPopup/ContextualPopup.js
```diff
@@ -580,7 +606,6 @@ module.exports = kind(
 	showingChanged: function () {
 		Popup.prototype.showingChanged.apply(this, arguments);
 		this.alterDirection();
-		this.showHideScrim(this.showing);
```
KunmyonChoi: I think this is redundant call because we already call it on enyo popup.

src/ContextualPopup/ContextualPopup.js
aarontam: Hi @KunmyonChoi, I think the mutex seems reasonable.

src/ContextualPopup/ContextualPopup.js
```diff
@@ -222,6 +229,18 @@ module.exports = kind(
...
+	},
+
+	/**
+	* Sync spotlightModal and modal properties on creation time.
+	*
+	* @private
+	*/
+	syncProperties: function () {
+		if (this.modal == this.spotlightModal) return;
+		if (this.hasOwnProperty('spotlightModal')) this.modal = this.spotlightModal;
```
aarontam: Do you know why we need to use `hasOwnProperty` here? I think we would still want to sync the (default) properties of the kind if they were not set on the instance.

src/ContextualPopup/ContextualPopup.js
```diff
@@ -568,10 +606,10 @@ module.exports = kind(
 		// show a transparent scrim for modal popups if
 		// {@link module:moonstone/ContextualPopup~ContextualPopup#scrimWhenModal} is `true`, else show a
 		// regular scrim.
-		if (this.modal && this.scrimWhenModal) {
+		if (this.modal && this.scrimWhenModal && !this.scrim) {
```
aarontam: I think we could potentially have two scrims appear if we changed the value of `this.scrim` and then proceeded to change the value of `spotlightModal`, for example.

src/ContextualPopup/ContextualPopup.js
```diff
@@ -538,18 +557,37 @@ module.exports = kind(
...
 				var i = this.getScrimZIndex();
 				this._scrimZ = i;
-				scrim.showAtZIndex(i);
+				this._scrim.showAtZIndex(i);
 			} else {
-				scrim.hideAtZIndex(this._scrimZ);
+				this._scrim.hideAtZIndex(this._scrimZ);
 			}
-			util.call(scrim, 'addRemoveClass', [this.scrimClassName, scrim.showing]);
+			util.call(this._scrim, 'addRemoveClass', [this.scrimClassName, this._scrim.showing]);
```
aarontam: Do we also need to perform this class clean-up when hiding the scrim in the other block, below?

src/ContextualPopup/ContextualPopup.js
```diff
@@ -538,18 +557,37 @@ module.exports = kind(
...
+
+	/**
+	* @private
+	*/
+	showHideScrim: function (show) {
 		if (this.floating && (this.scrim || (this.modal && this.scrimWhenModal))) {
-			var scrim = this.getScrim();
-			if (inShow && this.modal && this.scrimWhenModal) {
+			this._scrim = this.getScrim();
+			if (show) {
```
aarontam: Would it work to combine this with the conditional above, so that we have a single if/else for showing/hiding the scrim? We could consolidate all of the clean-up logic and remove some redundancy. Also, I don't know if we necessarily need to cache the scrim reference as an instance member, as the scrim is a singleton anyways.

src/ContextualPopup/ContextualPopup.js
KunmyonChoi: I'm copying comment from Jira. 

---
Hi Aaron Tam,

Current default properties are:
- enyo.Popup.modal = false
- moon.ContextualPopup.spotlightModal = false

If user can specify both of modal and spotlightModal on components block. I can think of below four cases.
1) User can override modal -> sync from modal to spotlightModal
2) User can override spotlightModal -> sync from spotlightModal to modal
3) User can override both modal and spotlightModal with same value -> No sync needed
4) User can override both modal and spotlightModal with different value -> ???

If modal and spotlightModal is doing the same, I think we don't even need to open spotlightModal as published property. But, just change it internally. How do you think?

There is another way of syncing. Lookup hasOwnProperty and decide the direction of sync.
But, in case of 4), we need to decide the default sync direction like 'sync from spotlightModal to modal' with some warning to developer.
---

The code is comes from the last comment.

aarontam: Thanks for the clarification! I don't think we can make `spotlightModal` private, as that affects the enabling of the close button, and developers would still need a mechanism for controlling the close button. I think for that reason, and because `spotlightModal` is the more "specific" property (`modal` is more general and is inherited from `enyo/Popup`), it seems to make the most sense to give `spotlightModal` priority in terms of sync direction (i.e. sync from `spotlightModal` to `modal`).

It wasn't clear to me that we needed or wanted to support "overriding" of one or both of the properties at create time, as the original issue could then reoccur. Are there specific cases you know of where we need to support this? If not, we can simplify the syncing logic to `this.modal = this.spotlightModal`. I think we would want to be as consistent as possible in our handling of these properties, if we are going to make a change.

src/ContextualPopup/ContextualPopup.js
KunmyonChoi: Yes, and I think that is caused because current scrim is creating separate instance for transparent and translucent. If we don't have a plan to have single instance for both, then I think we need to hide previous scrim before showing other kind of scrim. 

No problem in below case:
1) Popup A is showing transparent scrim then hide.
2) Popup B is showing translucent scrim then hide.

Problem in below case.
1) Popup A is showing transparent scrim.
2) Popup A is change this.scrim property to show translucent scrim (new instance).

src/ContextualPopup/ContextualPopup.js
KunmyonChoi: I don't have any actual cases for overriding both properties at creation time. So, simply sync like 'this.modal = this.spotlightModal' then add some documentation about its sync direction would be fine.

src/ContextualPopup/ContextualPopup.js
```diff
@@ -222,6 +229,17 @@ module.exports = kind(
...
 		Popup.prototype.initComponents.apply(this, arguments);
+		this.syncProperties();
+	},
+
+	/**
+	* Sync spotlightModal and modal properties on creation time.
+	*
+	* @private
+	*/
+	syncProperties: function () {
```
aarontam: Not sure we need a separate method for this anymore, after pruning some of its operations. Could we move the assignment operation into our initialization routine, instead, and drop the early return code?

src/ContextualPopup/ContextualPopup.js
```diff
@@ -538,19 +556,41 @@ module.exports = kind(
...
+
+		if (show && this.floating && (this.scrim || (this.modal && this.scrimWhenModal))) {
+			// move scrim to just under the popup to obscure rest of screen
+			var i = this.getScrimZIndex();
+			this._scrimZ = i;
+			scrim.showAtZIndex(i);
+		} else {
+			scrim.hideAtZIndex(this._scrimZ);
 		}
+		util.call(scrim, 'addRemoveClass', [this.scrimClassName, scrim.showing]);
```
aarontam: I'm curious about the original intent of this code (this was pre-existing). It seems like if we hide the scrim associated with this control, we could potentially re-add (or at least not remove) the scrim class for our current control, if the scrim is still showing (for another control).

src/ContextualPopup/ContextualPopup.js
aarontam: Hiding the previous scrim makes sense to me. I'm on the fence about how much work we want to do to fix up some of the misguided logic, as you mentioned this also exists in `moonstone/Popup`. My comment above is along those lines; let's fix things where we can, but if it involves some major refactoring I think it would be best to hold off.

src/ContextualPopup/ContextualPopup.js
aarontam: Also, let's make sure we document this behavior in the public-facing documentation, either in the description of the kind, or in the description of the published `spotlightModal` property. It might not hurt to explicitly document the `modal` property as well so that it is clear that the default value of both properties is `false`, with some details on how they are synced and in what direction.

src/ContextualPopup/ContextualPopup.js
```diff
@@ -538,19 +556,41 @@ module.exports = kind(
...
+	/**
+	* @private
+	*/
+	showHideScrim: function (show) {
+		var scrim = this.getScrim();
+
+		if (this._scrim && scrim != this._scrim) {
+			// hide if there was different kind of scrim
+			this._scrim.hideAtZIndex(this._scrimZ);
+			util.call(this._scrim, 'addRemoveClass', [this.scrimClassName, this._scrim.showing]);
```
aarontam: Can we simplify this as:

```
if (this.scrimClassName) this._scrim.removeClass(this.scrimClassName);
```

This is assuming that the intention was to actually remove the scrim class when hiding the scrim associated with this control (rather than relying on the scrim's showing/hiding state).

KunmyonChoi: Good catch!

src/ContextualPopup/ContextualPopup.js
KunmyonChoi: And, I don't understand the intention of using util.call instead of below just like moon.Popup.
`scrim.addRemoveClass(this.scrimClassName, scrim.showing);`

src/ContextualPopup/ContextualPopup.js
```diff
@@ -538,19 +553,41 @@ module.exports = kind(
...
-			}
-			util.call(scrim, 'addRemoveClass', [this.scrimClassName, scrim.showing]);
+	updateScrim: function (old, value, source) {
+		if (this._updateScrimMutex) return;
+		this._updateScrimMutex = true;
+
+		// We sync modal and spotlightModal here because binding doesn't
+		// guarantee sequence when it is used with observers.
+		if (source == 'modal') this.set('spotlightModal', this.modal);
+		if (source == 'spotlightModal') this.set('modal', this.spotlightModal);
```
webOS101: Is this necessary?

KunmyonChoi: We can have two way for setting spotlightModal value.
1) Set property on inherited kind.
- The modal value is copied from spotlightModal on initComponents. 
2) Set binding on inherited kind.
- The updateScrim is called from observer, but not copied from spotlightModal yet in this call.

So, below code is the related use case.

```
var myC = kind({
	name: 'myC',
	kind: ContextualPopup,
	bindings: [
		{from: 'bs', to: 'spotlightModal'}
	]
})
```

PR #2705, opened 2016-02-12
ENYO-3074 Focus to wrong item in NewDataList

changgilee:
Issue
: NewDataList restore worng focus

Fix
: Before focus to first visible item, add condition that the item is visible or not in screen

Enyo-DCO-1.1-Signed-off-by:Changgi Lee changgi.lee@lge.com

### DIFF
--- a/src/NewDataList/NewDataList.js
+++ b/src/NewDataList/NewDataList.js
@@ -124,5 +124,14 @@ module.exports = kind({
 			Spotlight.spot(fv);
 			return true;
 		}
+	},
+	isVisibleChild : function(control){
+		var controls = this.orderedChildren.slice(this.firstFullyVisibleI, this.lastFullyVisibleI + 1);
+		for(var i=0;i<controls.length;i++){			
+			if( control.isDescendantOf(controls[i]) ){
+				return true;
+			}
+		}
+		return false;
 	}
 });
--- a/src/NewDataList/NewDataList.js
+++ b/src/NewDataList/NewDataList.js
@@ -87,7 +87,7 @@ module.exports = kind({
 	* Spotlight focused. Currently, setting `focus: true` forces the
 	* scroll behavior to be `instant`, meaning that the scroller will
 	* jump to the item (with no animation).
-	* 
+	*
 	* @see module:enyo/NewDataList~NewDataList.scrollToItem
 	* @param {number} index - The (zero-based) index of the item to scroll to
 	* @param {Object} opts - Scrolling options (see module:enyo/Scrollable~Scrollable.scrollTo)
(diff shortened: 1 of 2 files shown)


src/Scrollable/Scrollable.js
```diff
@@ -138,8 +138,8 @@ var Scrollable = {
 				// default, we short-circuit the action because the child is
 				// potentially not in view, and focusing would cause it to
 				// scroll it into view for no reason apparent to the user.
-				// Instead, we should focus a child we know to be visible.
-				if (this.eventIsFromScrollingChild(event)) {
+				// Instead, we should focus a child we know to be visible.				
+				if (this.eventIsFromScrollingChild(event) && !this.isVisibleChild(event.originator) ) {
```
germboy: For consistency & safety, it may be best to pass `event` as the argument of `isVisibleChild()` here. In doing that, the method name `eventIsFromVisibleChild()` could make more sense.

src/NewDataList/NewDataList.js
```diff
@@ -124,5 +124,14 @@ module.exports = kind({
 			Spotlight.spot(fv);
 			return true;
 		}
+	},
+	isVisibleChild : function(control){
```
germboy: For style consistency, could you remove the white-space before the `:` here?

germboy: Could you also include comment blocks, identifying this as a private event (similar to `eventIsFromScrollingChild`)?

src/Scrollable/Scrollable.js
```diff
@@ -147,6 +147,16 @@ var Scrollable = {
 				return false;
 		}
 	},
+	isVisibleChild: kind.inherit(function (sup) {
```
germboy: Could you also include a comment block, identifying this as a private event?

src/NewDataList/NewDataList.js
```diff
@@ -124,5 +124,14 @@ module.exports = kind({
...
+		var controls = this.orderedChildren.slice(this.firstFullyVisibleI, this.lastFullyVisibleI + 1);
+		for(var i=0;i<controls.length;i++){			
+			if( control.isDescendantOf(controls[i]) ){
```
germboy: For style consistency, could you change the white-space to reflect `if (control.isDescendantOf(controls[i])) {`

src/NewDataList/NewDataList.js
```diff
@@ -124,5 +124,19 @@ module.exports = kind({
 			Spotlight.spot(fv);
 			return true;
 		}
+	},
+	/**
+	* Called by moonstone/Scrollable.filterFocus()
+	* @private
+	*/
+	isVisibleChild: function(event){
```
graynorton: For consistency with `eventIsFromScrollingChild()`, and to match the nature of the argument, I'd recommend that we name this `eventIsFromVisibleChild()`.

@germboy, rather than kick it back again, I think you can go ahead and make this change as part of integrating.

src/Scrollable/Scrollable.js
```diff
@@ -150,6 +150,23 @@ var Scrollable = {
 
 	/**
 	* This check is factored out of `filterFocus()` so that the logic can be
+	* overridden by the kind that includes the `Scrollable` mixin.
+	*
+	* @private
+	*/
+	isVisibleChild: kind.inherit(function (sup) {
```
graynorton: See note above about naming.

PR #2707, opened 2016-02-16
PLAT-16199: Change default value of spotlightModal and showCloseButton

viodragon2:


otisnate:
I think you need to add ```useDivider``` property

viodragon2:
@otisnate This PR is for the default value change only. Layout change can be found in #2702.

PR #2709, opened 2016-02-19
PLAT-16900: Updated neutral theme colors to resemble light theme.

Djspaceg:


css/moonstone-variables-dark.less
```diff
@@ -1,7 +1,8 @@
 // Moonstone Dark Theme
 // ---------------------------------------
-@moon-dark-gray:   #4b4b4b;
-@moon-light-gray:  #ededed;
+@moon-dark-gray:   #4d4d4d;
+@moon-gray:        #a6a6a6;
```
otisnate: Is ```@moon-gray```  just for reference of ```@moon-dark-gray``` and ```@moon-light-gray``` ?

Djspaceg: We were missing a 'middle gray' color that we used in many places below.

otisnate: Then, will you add 'middle gray' color to ```moonstone-variables-light.less``` file, too?

Djspaceg: Sure. Why not.

PR #2711, opened 2016-02-22
Fixes ENYO-3101

ryanjduffy:
If the client code relies on the data in the collection to drive the
initial selection, it will not do so on create because the repeater has
not rendered yet so the onActivate event doesn't fire. So, if the
collection for an ExpandableDataPicker changes and the list hasn't
rendered yet, force a refresh to trigger the onActivate event, update
the selected item, and display the correct text in currentValueText.

### DIFF
--- a/src/ExpandableDataPicker/ExpandableDataPicker.js
+++ b/src/ExpandableDataPicker/ExpandableDataPicker.js
@@ -70,6 +70,7 @@ module.exports = kind(
 	*/
 	collectionChanged: function () {
 		this.$.list.set('collection', this.collection);
+		if (this.collection && !this.$.list.hasNode()) this.$.list.refresh();
 	},
 	
 	/**


src/ExpandableDataPicker/ExpandableDataPicker.js
```diff
@@ -70,6 +70,7 @@ module.exports = kind(
 	*/
 	collectionChanged: function () {
 		this.$.list.set('collection', this.collection);
+		if (this.collection && !this.$.list.hasNode()) this.$.list.refresh();
```
otisnate: ```this.$.list.refresh()``` will render each items in collection. I think an advantage of ExpandableDataPicker comparing to ExpandablePicker is to not create/render child controls before picker is expanded

Is there another fix not creating every child controls of picker?

ryanjduffy: The only alternative I can come up with would be to iterate over the collection and find the first entry with `active: true`. That'd be more performant but it does make the pretty key assumption that   that property will be used to trigger the `onActivate` event. If a dev used a custom property or a custom control (other than CheckboxItem), it wouldn't work as expected.

At a minimum, we'd need to document that constraint or look for a more comprehensive change.

ryanjduffy: I think the benefit of ExpandableDataRepeater is that it leverages a Collection to generate the children rather than requiring a static list (or manually managed list). I'll add that this change *does not* force a render but it will create the controls. There is still a cost, of course, but it's minimized. The rendering delay is managed by the drawer set to `renderOnShow: true`.

PR #2733, opened 2016-02-26
PLAT-16501 moon.Input provides validator to prevent inputting invalid number

Sangwook-Lee:
### Issue
This is new UX requirement in 2017 TV. 
User can input the invalid value out of correct range. So moon.input provides validator to prevent inputting invalid number.

### Fix
I added published variables(min, max) to make range from default values.
Also I implemented change handler to make it dynamically.
If moon.input has number type, min and max value, contexturePopup is shown by validity when user input the wrong number.

Enyo-DCO-1.1-Signed-off-by: Sangwook Lee <sangwook1203.lee@lge.com>

### DIFF
--- a/src/InputDecorator/InputDecorator.js
+++ b/src/InputDecorator/InputDecorator.js
@@ -18,6 +18,8 @@ var
 	RichText = require('../RichText'),
 	TextArea = require('../TextArea');
 
+var
+	ContextualPopup = require('../ContextualPopup');
 /**
 * {@link module:moonstone/InputDecorator~InputDecorator} is a control that provides input styling. Any controls
 * in the InputDecorator will appear to be inside an area styled as an input. Usually,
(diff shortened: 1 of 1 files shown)


src/InputDecorator/InputDecorator.js
```diff
@@ -87,6 +89,8 @@ module.exports = kind(
 	*/
 	handlers: {
 		onDisabledChange    : 'disabledChangeHandler',
+		onenabledValidity   : 'enabledValidityHandler',
```
KunmyonChoi: The 'onEnabledValidity' would be better to custom event name.

KunmyonChoi:
Please create one more PR or 2.6.0-dev.

KunmyonChoi:
It looks like the popup is only show when input changes.
Hidden on blur and not shown on focus.

src/InputDecorator/InputDecorator.js
```diff
@@ -104,6 +108,15 @@ module.exports = kind(
 	_oInputControl: null,
 
 	/**
+	* @private
+	*/
+	tools: [
+		{name: 'rangeMsgPopup', kind: ContextualPopup, direction: 'right', components: [
+			{content: 'Invalid value !'}	// To DO : This content will be changed from UX. It is temporary content.
```
KunmyonChoi: Please make the content as a published property.

PR #2735, opened 2016-02-27
ENYO-2959: Change ListActions with new design

viodragon2:
We're going to change `ListActions` from `Drawer` to a `ContextualPopup` for 2017 TV.

Following are some of the notable changes from existing ListActions
- The base kind for `ListActions` is `ContextualPopupDecorator`.
- `open`, `autoCollapse` and `proportionalWidth` properties are removed.
- No longer stacks. (assumes there would be maximum of 3 columns horizontally)
- `Header` no longer handles events from `ListActions`.
- `listActions` block no longer use `FittableRowLayout`.
- Fix the height of scrollable area to `height: 300px`.
- Unlike `ContextualPopup`, the new ListActionPopup sets `spotlightModal: true` and `showClosebutton: true` by default.
- Popup is anchored to the right and points arrows aligned to the center of `IconButtons`. If the width is not wide enough to touch the edge of screen, then it would align center.

### DIFF
--- a/src/ListActions/ListActions.js
+++ b/src/ListActions/ListActions.js
@@ -2,39 +2,26 @@ require('moonstone');
 
 /**
 * Contains the declarations for the {@link module:moonstone/ListActions~ListActions}
-* and {@link module:moonstone/ListActions~ListActionsDrawer} kinds, and the
+* and {@link module:moonstone/ListActions~ListActionsPopup} kinds, and the
 * {@link module:moonstone/ListActions~ListActionActivationSupport} mixin.
 * @module moonstone/ListActions
 */
 
 var
-	kind = require('enyo/kind'),
 	dom = require('enyo/dom'),
-	ri = require('enyo/resolution'),
-	dispatcher = require('enyo/dispatcher'),
-	Control = require('enyo/Control'),
-	EnyoHistory = require('enyo/History'),
-	GroupItem = require('enyo/GroupItem');
+	ri = require('enyo/resolution');
 
 var
-	FittableLayout = require('layout/FittableLayout'),
-	FittableRowsLayout = FittableLayout.Rows;
-
-var
-	Spotlight = require('spotlight');
-
-var
-	$L = require('../i18n'),
-	Scroller = require('../Scroller'),
-	IconButton = require('../IconButton'),
-	HistorySupport = require('../HistorySupport');
+	ContextualPopup = require('moonstone/ContextualPopup'),
+	ContextualPopupDecorator = require('moonstone/ContextualPopupDecorator'),
+	IconButton = require('moonstone/IconButton');
 
 /**
 * An internally-used support mixin added to a {@link module:moonstone/ListActions~ListActions}
 * menu, which decorates `activate` events with the menu's `action` property.
 *
 * @mixin ListActionActivationSupport
-* @protected
+* @private
 */
 var ListActionActivationSupport = {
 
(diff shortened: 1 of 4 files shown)


src/ListActions/ListActions.js
```diff
@@ -2,32 +2,19 @@ require('moonstone');
 
 /**
 * Contains the declarations for the {@link module:moonstone/ListActions~ListActions}
-* and {@link module:moonstone/ListActions~ListActionsDrawer} kinds, and the
+* and {@link module:moonstone/ListActions~ListActionPopup} kinds, and the
```
aarontam: Very minor: this should probably be `ListActionsPopup`.

src/ListActions/ListActions.less
```diff
@@ -1,115 +1,109 @@
...
 
-	&.open {
-		-webkit-transform: translateZ(0) translateY(0);
-		tranform: translateZ(0) translateY(0);
-	}
-	&.animated {
-		-webkit-transition: -webkit-transform .225s cubic-bezier(0.25, 0.1, 0.25, 1);
-		transition: transform .225s cubic-bezier(0.25, 0.1, 0.25, 1);
+	&.right {
+		&.list-actions-1h {
```
aarontam: Because we made the `list-actions-left-hunits` mixin very specific to the `ListActions` use case, what do you think about also incorporating the `before` and `after` pseudo element selectors as part of the mixin? And maybe we could consider passing the position property name as an additional parameter i.e. `.list-actions-h(1, left);` since the mixin is the same for both left and right, save for the property name. If we want to get really fancy, we could consider creating all of these selectors in a loop too: http://lesscss.org/features/#loops-feature.

src/ListActions/ListActions.js
```diff
@@ -555,267 +285,33 @@ var ListActions = module.exports = kind(
...
-		if (this.resetScroller) {
-			this.$.listActions.scrollTo(0, 0);
-			this.resetScroller = false;
-		}
-		if (this.generated && !rendered) {
-			Spotlight.spot(this.$.closeButton);
-		}
+	popupShown: function() {
+		ContextualPopupDecorator.prototype.popupShown.apply(this,arguments);
 		this.bubble('onRequestMuteTooltip');
```
aarontam: Do you think we still need this if the popup will never overlap with the tooltip? Is there a case where this isn't true? And lastly, does it make sense to potentially display both?

src/ListActions/ListActions.js
```diff
@@ -58,202 +45,87 @@ var ListActionActivationSupport = {
...
+* {@link module:moonstone/ListActions~ListActionsPopup} is a
+* [control]{@link module:moonstone/ContextualPopup~ContextualPopup} used by
 * {@link module:moonstone/ListActions~ListActions} to house a menu of selectable options.
 *
-* @class ListActionsDrawer
-* @extends module:enyo/Control~Control
+* @class ListActionsPopup
+* @extends module:moonstone/ContextualPopup~ContextualPopup
 * @ui
 * @public
```
aarontam: This might have been improperly documented before, but I'm thinking this should be `@private`. cc @webOS101

src/ListActions/ListActions.js
viodragon2: good catch!

src/ListActions/ListActions.js
viodragon2: Yes I believe it's still needed. `ListActions` which is a subkind of `ContextualPopupDecorator` can be also wrapped in `TooltipDecorator`. If we don't mute tooltip, every spotlight focus inside the `ListActionsPopup` will show tooltip. It may look alright, but spotlight blur will disappear tooltip, which means that every spot move in both 5-way and pointer mode will cause to hide a show tooltip with delays. I don't think that would be okay.

aarontam: Gotcha! Makes sense.

src/ListActions/ListActions.js
```diff
@@ -58,202 +45,87 @@ var ListActionActivationSupport = {
...
-	},
-
-	/**
-	* We override `getBubbleTarget()` here so that events emanating from a
-	* [ListActionsDrawer]{@link module:moonstone/ListActions~ListActionsDrawer} instance will bubble to the owner
-	* of the associated [ListActions]{@link module:moonstone/ListActions~ListActions} instance, as expected. This
-	* is necessary because events normally bubble to a control's DOM parent, but we have
-	* sneakily arranged for the DOM parent of a `ListActionsDrawer` instance to be not the
-	* `ListActions` instance, but the containing [Header]{@link module:moonstone/Header~Header} instance.
+	* If `true`, {@glossary Spotlight} (focus) cannot leave the area of the popup unless the
```
aarontam: Instead of copying this description from `ContextualPopup`, we can replace this with

```
@see enyo/ContextualPopup~ContextualPopup#spotlightModal
```

src/ListActions/ListActions.less
viodragon2: Good idea! Didn't think you could pass CSS property as a parameter.

src/ListActions/ListActions.js
viodragon2: Since we're exporting `ListActionsPopup` in L309, I think it could as is.

src/ListActions/ListActions.js
```diff
@@ -58,202 +45,89 @@ var ListActionActivationSupport = {
...
 	},
 
 	/**
-	* We override `getBubbleTarget()` here so that events emanating from a
-	* [ListActionsDrawer]{@link module:moonstone/ListActions~ListActionsDrawer} instance will bubble to the owner
-	* of the associated [ListActions]{@link module:moonstone/ListActions~ListActions} instance, as expected. This
-	* is necessary because events normally bubble to a control's DOM parent, but we have
-	* sneakily arranged for the DOM parent of a `ListActionsDrawer` instance to be not the
-	* `ListActions` instance, but the containing [Header]{@link module:moonstone/Header~Header} instance.
+	* Prevent waterfalling 'onresize'.
```
ryanjduffy: Can we document the 'why' here as well?

src/ListActions/ListActions.js
```diff
@@ -263,166 +137,93 @@ var ListActionsDrawer = kind(
...
+	* CSS classes to apply to adjust width of each actions.
+	*
+	* @type {String}
+	* @default
+	* @public moon-list-actions-popup-width
 	*/
-	components:[
-		{name:'activator', kind: IconButton, classes: 'moon-list-actions-activator', ontap: 'expandContract'}
-	],
+	actionWidthClasses: 'moon-list-actions-popup-width',
```
ryanjduffy: Should this really be a public, configurable property? I don't think we'd really want anyone setting a different value for this since we've included it in our CSS. If the width is configurable, should it be set by a LESS variable instead?

src/ListActions/ListActions.less
```diff
@@ -1,115 +1,54 @@
...
 }
 
-/* Drawer */
-.moon-list-actions-drawer {
-	overflow: hidden;
-	z-index: 50;
-	pointer-events: none;
-	* {
-		pointer-events: auto;
+.generate-columns(@n, @i: 1) when (@i =< @n) {
```
ryanjduffy: Suggest moving this within `.moon-list-actions-popup`. As it stands, it'd a global LESS function and could be stomped by LESS elsewhere. Since it's truly private to this control, we should scope it as such.

src/ListActions/ListActions.js
webOS101: Also, it should be jsdocced, @private or not.

ryanjduffy: It's docced (up a couple lines) although the default value follows the scope pragma instead of default (oops)

src/ListActions/ListActions.js
```diff
@@ -58,202 +45,89 @@ var ListActionActivationSupport = {
...
+			var clientRect = this.getBoundingRect(this.node),
+				viewPortWidth = dom.getWindowWidth(),
+				offsetWidth = (clientRect.width - this.activatorOffset.width) / 2,
+				popupMargin = ri.scale(24),
+				iconButtonWidth = this.activatorOffset.width + popupMargin,
+				bounds = {top: null, left: null},
+				c;
+
+			this.resetDirection();
+			this.addClass('below');
```
ryanjduffy: Any reason we can't include this in the classes for the kind since we're always 'below'?

ryanjduffy:
Overall, the CSS for spacing seems unnecessarily complex. We're overriding the spacing specified by moon-hspacing for this scenario and then further customizing it in ListActions.less. I took a cut at a much simplified version but I may be missing edge cases of which I wasn't aware.

27ba4b4

src/ListActions/ListActions.js
viodragon2: (doh) Wrong line!

re: public / configurable. I am not super opposed to changing it to private. The reasoning behind making it a public was that there's no easy way for developers to adjust width in case if any app wanted to do so as action width is defined by `controlClasses` of the wrapper. Plus, if those cases do happen, it's more than likely that those widths won't fall under what we provide in the framework and apps will come up with their own numbers. I wanted make sure they have some capability instead of restricting it to one width.

@otisnate can help us to confirm whether this is needed or not.

src/ListActions/ListActions.js
viodragon2: Good point.

viodragon2:
I agree that CSS spacing is very ugly and complicated. However, to respect 12px margin left/right of icon button, we can not simply set `margin: 0` for listaction icon buttons as moon-hspacing will add 9px margin left/right. I'll definitely look into this further to cleanup a bit, but I still think contextual-popup-decorator and tooltip-decorator should set their margin: 0 when inside the moon-hspacing.

src/ListActions/ListActions.js
ryanjduffy: Since ListActions is a specialized version of a ContextualPopup, I don't think we need to bake in too much design flexibility into the API. When the design calls for ListActions, you should use it mostly as is for UX consistency. If you're doing something else, you'll have to roll your own. Rails over rope.

That said, there are several options for overriding the width with existing APIs/technologies (overriding the CSS rules with the already applied classes, adding `classes` to a list action with a custom CSS class, applying inline style to a list action using `style`). I don't think we're adding a lot of value by introducing another.

ryanjduffy:
Based on the redline, the distance between the right sides of the icons is 80px. We can't have that since it's not divisible by three so we should aim for 84px or 78px. 81px isn't really possible since the IconButton is 60px and we're adding 2 margins.

The 18px margin (9px per side) introduced by `.moon-hspacing` with no `.moon-icon` margin gives us 78px between edges. With `.moon-icon`'s 12px margin and no `.moon-hspacing` margin, we'd have 84px between edges.

So, we need to decide which set of rules we want to enforce:
* the 12px margin between icon buttons,
* the 9px margin for horizontal spacing, or
* the ListActions-specific redline (altered to be divisible by 3).

cc: @Djspaceg

viodragon2:
One of the reasons why I wanted to enforce 12px margin between icon buttons was that the tap area would overlap each other and making it to spot the previous item even after mouse cursor is over the one next to ti. However, the tap area was wrong and we were able to fix that in #2755. So 9px margin between icon buttons does not give us that problem anymore. Consequently, we can make the less file structure simple now.

PR #2749, opened 2016-03-07
ENYO-3093: Fix video-slider focus bug

germboy:
### Issue

The VideoPlayer can lose window/spotlight focus while holding/pressing the VideoTransportSlider. At this point, the video controls can be hidden (via the autoHide job). At this point, when you attempt to tap the video, you receive an error in the console. When the VideoTransportSlider loses window/spotlight focus while being held/pressed, it never receives an `up` event (even when regaining focus), so the slider remains in a semi-unreleased/drag state. Clicking the video, even when the controls are hidden, attempts to fire drag events.
### Fix

Issue an `up` event in the specific case where the slider loses focus while an unreleased `downEvent` exists.

Enyo-DCO-1.1-Signed-off-by: Jeremy Thomas jeremy.thomas@lge.com


Djspaceg:
Could this change be relevant to `moonstone/Slider` as well?

src/VideoTransportSlider/VideoTransportSlider.js
```diff
@@ -454,6 +456,9 @@ module.exports = kind(
 		this.selected = false;
 		this.removeClass('visible');
 		this.endPreview();
+		if (gesture.downEvent) {
+			gesture.up(util.clone(gesture.downEvent));
```
aarontam: I'm not sure we need to clone the `downEvent` here, as we are effectively doing this in `up` already: https://github.com/enyojs/enyo/blob/2.6.0-dev/src/gesture/gesture.js#L82. Semantically, it seemed slightly weird to send the `downEvent` to `up`, but not sure of an alternative, as it seems to be the only property that contains the relevant properties specific to the event originator.

Also, along the lines of @Djspaceg's feedback, I think we could make this more general and move this to Spotlight's `onEvent` handler, similar to how we are calling `gesture.drag.endHold()` for the `blur` event.

germboy:
This change *could* also be relevant to `moonstone/Slider` as a similar edge-case. However, in this case, `moonstone/VideoTransportSlider` overrides the `moonstone/Slider` `spotBlur` method, which is why I've settled on where I've placed this change. So we could look at creating another JIRA/PR for that use-case.

lishammel:
Closing PR as Jira is closed.

PR #2751, opened 2016-03-08
ENYO-3189: Convert "more" icons references from SVG assets to icon font

Djspaceg:
I've left the "libPath" variable in place so it's available for future cases.


ryanjduffy:
I don't think we need to keep `libPath` around if it isn't used by VideoPlayer. No reason to willingly keep cruft.

Djspaceg:
Removed and updated.

PR #2752, opened 2016-03-09
ENYO-3183: Corrected opacity level for input Headers' placeholders

Djspaceg:
... In Firefox

### DIFF
--- a/src/Header/Header.less
+++ b/src/Header/Header.less
@@ -172,6 +172,7 @@
 					color: @moon-input-header-placeholder-color;
 					margin-top: 12px;
 					line-height: 1.25em;
+					opacity: 1;
 				});
 			}
 
--- a/src/Input/Input.less
+++ b/src/Input/Input.less
@@ -34,11 +34,11 @@
 	.moon-disabled .moon-input {
 		cursor: default;
 	}
-	> input::-webkit-input-placeholder {
-		color: @moon-input-placeholder-color;
-	}
-	> input:-moz-placeholder {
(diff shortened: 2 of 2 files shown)


src/Header/Header.less
```diff
@@ -172,6 +172,7 @@
 					color: @moon-input-header-placeholder-color;
 					margin-top: 12px;
 					line-height: 1.25em;
+					opacity: 1;
```
viodragon2: Could we instead be specific to FireFox only?
`
::-moz-placeholder {
  opacity: 1;
}
`

Djspaceg: We _could_, but it doesn't hurt webkit/others, and a separate rule would just add bulk to the already very heavy CSS. I'm not concerned about hitting other browsers with this too.
The resulting selectors code would look like this:
```
.moon-header.moon-input-header .moon-input-header-input-decorator .moon-input::-webkit-placeholder {
    color: @moon-input-header-placeholder-color;
    margin-top: 12px;
    line-height: 1.25em;
}
.moon-header.moon-input-header .moon-input-header-input-decorator .moon-input::-moz-placeholder {
    color: @moon-input-header-placeholder-color;
    margin-top: 12px;
    line-height: 1.25em;
}
.moon-header.moon-input-header .moon-input-header-input-decorator .moon-input::-moz-placeholder {
    opacity: 1;
}
```

viodragon2: Okay, no strong opposition.

PR #2755, opened 2016-03-09
ENYO-3207: IconButton now has its own tap area rules

Djspaceg:
And mixin updated/added for alternate case.

### DIFF
--- a/src/IconButton/IconButton.less
+++ b/src/IconButton/IconButton.less
@@ -12,6 +12,8 @@
 	margin: 0 @moon-spotlight-outset;
 	line-height: @moon-icon-sprite-size;
 
+	.moon-taparea(@moon-icon-button-size);
+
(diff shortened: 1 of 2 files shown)


src/IconButton/IconButton.less
```diff
@@ -12,6 +12,8 @@
 	margin: 0 @moon-spotlight-outset;
 	line-height: @moon-icon-sprite-size;
 
+	.moon-taparea(@moon-icon-button-size);
```
ryanjduffy: Is the idea here that if either `@moon-icon-button-size` or `@moon-button-small-tap-area-height` are altered such that a tap area would be required, it would automatically add it in?

Djspaceg: And, yes, if Icon size changes, this will auto-add/remove it.

PR #2757, opened 2016-03-09
ENYO-3197: Fix spotlight color and display for header text

viodragon2:

### DIFF
--- a/src/LabeledTextItem/LabeledTextItem.less
+++ b/src/LabeledTextItem/LabeledTextItem.less
@@ -8,7 +8,7 @@
 
 	// Header
 	.moon-labeledtextitem-header {
-		display: inline-block;
+		display: block;
 		box-sizing: border-box;
 		max-width: 100%;
 		position: relative;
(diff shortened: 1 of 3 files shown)


src/LabeledTextItem/LabeledTextItem.less
```diff
@@ -8,7 +8,7 @@
 
 	// Header
 	.moon-labeledtextitem-header {
-		display: inline-block;
+		display: block;
```
ryanjduffy: Changing this pushes the arrow for the expandable pickers all the way to the right. Was that your expectation? If so, we should probably document that as such in the ticket so we acknowledge it was intentional.

webOS101: Good catch.  I only tested labeledTextItem. :(

viodragon2: Ah, didn't catch that either. We should override `display: inline-block` for `moon-expandable-list-item` then.

PR #2769, opened 2016-03-15
ENYO-2709: Added support for sideways-facing tooltips

Djspaceg:
Tooltip nub also got a makeover in all cases, and now looks more like the original spec.

### DIFF
--- a/src/Tooltip/Tooltip.js
+++ b/src/Tooltip/Tooltip.js
@@ -15,6 +15,8 @@ var
 	Signals = require('enyo/Signals'),
 	ri = require('enyo/resolution');
 
+var pointerTemplate = '<path d="M0,5C0,3,1,0,3,0H0V5Z"/>';
+
 // To prevent lingering tooltips, we're monitoring spotlight changes and tooltip display
 // to ensure that only 1 tooltip is active.
 // see BHV-14524, ENYO-247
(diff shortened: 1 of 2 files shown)


src/Tooltip/Tooltip.js
```diff
@@ -314,7 +319,7 @@ module.exports = kind(
 	* @private
 	*/
 	positionChanged:function () {
-		Popup.prototype.positionChanged.apply(this, arguments);
+		Popup.prototype.positionChanged && Popup.prototype.positionChanged.apply(this, arguments);
```
Djspaceg: `positionChanged` was behaving strangely when the value was in a binding. Them _seemed_ to remedy the problem.

src/Tooltip/Tooltip.js
```diff
@@ -281,6 +284,8 @@ module.exports = kind(
 		if (this.contentUpperCase !== null) this.uppercase = this.contentUpperCase;
 
 		this.contentChanged();
+
+		this.$.point.setAttribute('viewBox', '0 0 3 5');
```
viodragon2: can we put this inside the components block instead? 
e.g. `{name: 'point', kind: Control,  ... , attributes: {viewBox: '0 0 3 5'}`

src/Tooltip/Tooltip.js
viodragon2: Surprised this never brought any problems before. There's no `positionChanged` in `enyo/Popup`. We should get rid of this line instead.

viodragon2:
There's no documentation on how to use sideway tooltip. I think editing `position` property would suffice.

viodragon2:
Also, `top` and `bottom` seemed a bit confusing to me. They represent where the tail resides in respect to the tooltip instead of direction of where tooltip is facing.

One last thing I want to mention is that it doesn't flip when the sideway tooltips go over the viewport of window like the `above` and `below` case. I don't think it's necessary, but it's worth mentioning in somewhere as part of the doc if we're not going to address those cases. (`below` position doesn't properly flip either. We should fix them all together if we do.)

src/Tooltip/Tooltip.js
Djspaceg: Of course! I didn't know you could interact with attributes that way. Will do.

Djspaceg:
All issues including documentation and inverted top/bottom addressed. (As well as a new bug found, and corrected. "auto" wasn't working as expected if chosen after a different position.)

PR #2770, opened 2016-03-15
PLAT-16501: Tooltip is added to be displayed for invalid message.

Sangwook-Lee:
### Issue
This is new UX requirement for 2017 TV. 
moonstone input set invalid-class and show tooltip with message about invalid value.

### Fix
Moonstone CSS variables is added to change text color in relation to invalid status.
Some published variables is added to set attributes for checking validity and their change handler are also added.
'invalidMessage' can be set to any strings. If it's to set to a falsy value, the tooltip is not shown.
Setting moon.Input's invalid property to true, shows tooltip and adds the invalid-class.

Enyo-DCO-1.1-Signed-off-by: Sangwook Lee <sangwook1203.lee@lge.com>

### DIFF
--- a/css/moonstone-variables.less
+++ b/css/moonstone-variables.less
@@ -251,6 +251,7 @@
 // ---------------------------------------
 @moon-input-border-width: 6px;
 @moon-input-height:60px;
+@moon-invalid-text-color: @moon-red;
 
 // Pickers
 // ---------------------------------------
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -11,6 +11,10 @@ var
 	Input = require('enyo/Input');
 
 var
+	TooltipDecorator = require('../TooltipDecorator'),
+	Tooltip = require('../Tooltip');
+
+var
 	Spotlight = require('spotlight');
 
 /**
@@ -76,7 +80,43 @@ module.exports = kind(
 		* @default false
 		* @public
 		*/
-		dismissOnEnter: false
+		dismissOnEnter: false,
+
+		/**
+		* The min attribute specifies the minimum value for an [input]{@link module:enyo/Input~Input}.
+		*
+		* @type {Number}
+		* @default null
+		* @public
+		*/
+		min: null,
+
+		/**
+		* The max attribute specifies the maximum value for an [input]{@link module:enyo/Input~Input}.
+		*
+		* @type {Number}
+		* @default null
+		* @public
+		*/
+		max: null,
+
+		/**
+		* When `true`, 'input-invalid' class is added and message tooltip is shown if it exists.
+		*
+		* @type {Boolean}
+		* @default false
+		* @public
+		*/
+		invalid: false,
+
+		/**
+		* Text to be displayed as the tooltip content if invalidMessage is set not ''.
+		*
+		* @type {String}
+		* @default null
+		* @public
+		*/
+		invalidMessage: ''
 	},
 
 	/**
@@ -89,12 +129,37 @@ module.exports = kind(
 	},
 
 	/**
+	* @private
+	*/
+	components: [
+		{kind: TooltipDecorator, autoShow: false, components: [
+			{name: 'tooltip', kind: Tooltip, floating: false, position: 'below'}	// To Do : It's position has dependency with ENYO-2709.
+		]}
+	],
+
+	/**
+	* @private
+	*/
+	bindings: [
+		{from: 'invalidMessage', to: '$.tooltip.content'}
+	],
+
+	/**
 	* Used only for [dismissOnEnter]{@link module:moonstone/Input~Input#dismissOnEnter} feature;
 	* we cannot rely on `hasFocus()` in this case due to race condition.
 	*
 	* @private
 	*/
 	_bFocused: false,
+
+	/**
+	* @private
+	*/
+	create: function () {
+		Input.prototype.create.apply(this, arguments);
+		this.minChanged();
+		this.maxChanged();
+	},
 
 	/**
 	* @private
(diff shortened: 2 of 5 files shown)


src/Input/Input.js
```diff
@@ -89,6 +129,22 @@ module.exports = kind(
...
+		{name: 'TooltipDecorator', kind: TooltipDecorator, autoShow: false, components: [
+			{name: 'tooltip', kind: Tooltip, floating: false, position: 'right top'}
+		]}
+	],
+
+	/**
+	* @private
+	*/
+	bindings: [
+		{from: '.invalidMessage', to: '$.tooltip.content'}
```
otisnate: .invalidMessage -> invalidMessage

src/Input/Input.js
```diff
@@ -89,6 +129,22 @@ module.exports = kind(
 	},
 
 	/**
+	* @private
+	*/
+	components: [
+		{name: 'TooltipDecorator', kind: TooltipDecorator, autoShow: false, components: [
+			{name: 'tooltip', kind: Tooltip, floating: false, position: 'right top'}
```
otisnate: Is Tooltip's position fixed like ``position: 'right top'`` ? I wonder if Tooltip's position in RTL locale is ``position: 'left top'`` or not

src/Input/Input.js
```diff
@@ -182,5 +247,38 @@ module.exports = kind(
...
+	* @private
+	*/
+	invalidChanged: function () {
+		if(!this.invalid) {
+			this.removeClass('input-invalid');
+			this.$.tooltip.set('showing', false);
+			return;
+		}
+
+		this.addClass('input-invalid');
```
otisnate: I think you can do refactoring to use ```addRemoveClass``` instead of using ```removeClass``` and ```addClass```

src/Input/Input.js
```diff
@@ -76,7 +80,43 @@ module.exports = kind(
...
+		invalid: false,
+
+		/**
+		* Text to be displayed as the tooltip content if invalidMessage is set not ''.
+		*
+		* @type {String}
+		* @default null
+		* @public
+		*/
+		invalidMessage: null
```
aarontam: Should we make the default value the empty string `''` instead of `null`?

src/Input/Input.js
```diff
@@ -182,5 +247,28 @@ module.exports = kind(
...
+		this.setAttribute('max', this.max);
+	},
+
+	/**
+	* @private
+	*/
+	invalidChanged: function () {
+		this.addRemoveClass('input-invalid', this.invalid);
+		this.$.tooltip.activator = this;
+		this.$.tooltip.set('showing', this.invalid && this.invalidMessage && this.invalidMessage !== '');
```
aarontam: I don't think the last part of this statement is necessary, as `this.invalidMessage` will be falsy if the value is `''`.

src/Input/Input.js
```diff
@@ -99,6 +155,15 @@ module.exports = kind(
 	/**
 	* @private
 	*/
+	create: function () {
+		Input.prototype.create.apply(this, arguments);
+		this.setAttribute('min', this.min);
```
aarontam: What do you think about changing these calls to
```
this.minChanged();
this.maxChanged();
```
to keep things DRY, and in case we decide to add additional logic to these change handlers?

src/Input/Input.js
```diff
@@ -76,7 +80,43 @@ module.exports = kind(
...
 		*/
-		dismissOnEnter: false
+		dismissOnEnter: false,
+
+		/**
+		* The min attribute specifies the minimum value for an [input]{@link module:enyo/Input~Input}.
+		*
+		* @type {Number}
+		* @default null
+		* public
```
aarontam: This should be `@public`. Same for the documentation for `max` and `invalid`.

src/Input/Input.js
```diff
@@ -89,6 +129,22 @@ module.exports = kind(
 	},
 
 	/**
+	* @private
+	*/
+	components: [
+		{name: 'TooltipDecorator', kind: TooltipDecorator, autoShow: false, components: [
```
aarontam: I don't know if we necessarily need to specify an explicit name here. If so, we should be consistent with our naming convention and make it lower camel case.

src/Input/Input.js
```diff
@@ -89,6 +129,22 @@ module.exports = kind(
...
+			{name: 'tooltip', kind: Tooltip, floating: false, position: 'right top'}	// To Do : It's position has dependency with ENYO-2709.
```
aarontam: Is there a specific reason we are setting `floating: false` here?

css/moonstone-variables.less
```diff
@@ -251,6 +251,7 @@
 // ---------------------------------------
 @moon-input-border-width: 6px;
 @moon-input-height:60px;
+@moon-input-invalid-text-color: @moon-red;
```
aarontam: It seems like it might be better to either make this variable a bit more generic in name, i.e. `@moon-invalid-text-color`, or use `@moon-red` directly in the `moonstone/Input` style, otherwise it kind of defeats the purpose of creating this variable.

src/Input/Input.js
aarontam: Also, curious if we actually need a TooltipDecorator in this case. It seems like it could be problematic if the decorator needs to wrap our activator, which is the input in this case...

src/Input/Input.js
```diff
@@ -89,6 +129,22 @@ module.exports = kind(
 	},
 
 	/**
+	* @private
+	*/
+	components: [
+		{kind: TooltipDecorator, autoShow: false, components: [
+			{name: 'tooltip', kind: Tooltip, floating: false, position: 'below'}	// To Do : It's position has dependency with ENYO-2709.
```
Sangwook-Lee: First, I changed position value from 'right top' to 'below'.
I guess, sideways tooltip is displayed on wrong position when I set floating to 'true'.
Like as I commented in Jira, I think this tooltip is not moved when scrolling up and down.

aarontam: Hi @Sangwook-Lee, that makes sense about the `floating` property, thanks!

aarontam: Taking a deeper look at this now - I thought we wanted the tooltip to display on the side of the input? Have you tried experimenting with the layout some more? It does not appear we are using `TooltipDecorator` correctly here, but since we are controlling the showing/hiding of the tooltip directly, we may not actually need to use it.

aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2782.

PR #2774, opened 2016-03-17
ENYO-3163: Fix Panels breadcrumb spotlight issue

germboy:
### Issue
When setting the index of a Panels control (in create time) to > 1, spotlight cannot focus on breadcrumbs, until you transition to another panel.

### Fix
Previous method required the breadcrumbs and their container to be rendered to calculate if they were spottable. We can use an alternate method - since we know how many breadcrumbs there are and how many can fit on the screen at a given time, we can calculate which breadcrumbs are spottable.

Enyo-DCO-1.1-Signed-off-by: Jeremy Thomas <jeremy.thomas@lge.com>

### DIFF
--- a/src/Panels/Panels.js
+++ b/src/Panels/Panels.js
@@ -1585,7 +1585,7 @@ module.exports = kind(
 		var end = this.toIndex,
 			start = end - this.getBreadcrumbs().length,
 			range = {start: start, end: end},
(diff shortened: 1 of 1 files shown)


aarontam:
Closed in favor of https://github.com/enyojs/moonstone/pull/2776.

PR #2781, opened 2016-03-21
ENYO-3242: Better non-floating tooltip support.

aarontam:
### Issue
At a general level, the tooltip tail overlaps the activator. We had increased the size of the tail due to some recent optimizations and support for sideways tooltips, but the default margin was set to a static size in the JS. Additionally, when using a non-floating tooltip, the tooltip is positioned relative to the activator, and does not take into account any potential border width of the activator.

### Fix
To resolve the first overlapping issue, we have noted the `18px` tail size of the tooltip, and increased the default margin to `21px` to accommodate this. To take into account the border width of the activator, we add the width of the left border to the position for a `left *` positioned tooltip, as the positioning is relative to the element excluding the border. For a `right *` positioned tooltip, we utilize the `clientWidth` of the activator instead of the bounds width, so that we can effectively account for `border-box` styles, and add the width of the right border. For an `above` positioned tooltip, we add the height of the top border. And for a `below` positioned tooltip, similar to `right *` positioned tooltips, we utilize the `clientHeight` of the parent (activator) instead of the bounds height, and add the height of the bottom border. We also convert to rem units instead of px for the `below` positioned tooltip, and make some various optimizations and cleanup.

### Notes
I'm not the most familiar with the inner workings of Tooltip, so would appreciate some extra eyes on this and whether or not these changes make sense in the grand scheme of Tooltip behavior. I don't believe that we had been using non-floating tooltips very frequently in Moonstone, which is why edge cases, such as activator border width, had not really been encountered.

### DIFF
--- a/src/Tooltip/Tooltip.js
+++ b/src/Tooltip/Tooltip.js
@@ -379,11 +379,12 @@ module.exports = kind(
 		if (this.showing && this.hasNode()) {
 			var b = this.node.getBoundingClientRect(),
 				moonDefaultPadding = ri.scale(18),
-				defaultMargin = ri.scale(15),
+				defaultMargin = ri.scale(21),
+				floating = this.get('floating'),
+				acNode = this.activator.hasNode(),
 				pBounds = this.parent.getAbsoluteBounds(),
-				acBounds = null;
-
-			acBounds = this.activator.getAbsoluteBounds();
+				acBounds = this.activator.getAbsoluteBounds(),
+				acBorders;
 
 			//* Calculate the difference between decorator and activating
 			//* control's top, left, right differences, position tooltip against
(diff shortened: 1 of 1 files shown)


src/Tooltip/Tooltip.js
```diff
@@ -412,7 +412,13 @@ module.exports = kind(
 				anchorLeft = window.innerWidth - moonDefaultPadding - pBounds.left - pBounds.width / 2 >= b.width;
 			}
 
-			offset = this.floating ? (acRight + moonDefaultPadding) : paRightDiff;
+			if (floating) {
+				offset = acRight + moonDefaultPadding;
+			} else {
+				offset = paRightDiff;
+				// we need to account for activator border widths if we are not floating
+				acBorders = dom.calcBoxExtents(this.activator.node, 'border');
```
aarontam: I chose to use `this.activator.node` instead of `this.activator.hasNode()` to save a function call, as we will always have called `this.activator.getAbsoluteBounds()` already, which will have already cached the node reference. I don't feel super strongly about this and can probably be convinced to change this to `hasNode()` for future safety.

PR #2786, opened 2016-03-22
ENYO-3243: Allow for popup value updating during knob dragging, fixing some latent bugs.

aarontam:
### Issue
When `lockBar` is set to `false`, dragging the knob does not result in updated popup values. Additionally, there were some latent bugs, such as setting the `showPercentage` property resulting in non-sensical popup value, and similarly when setting the `popupSide` property. A lot of this stems from a mismatch between `ProgressBar` and `Slider`, as the former relies on the value of `progress`, while the latter relies on the `value` property. 

### Fix
Instead of using `this.progress` when updating the popup, we now use the passed-in value. I didn't see any issues with this, as the value passed from `updateValues` is already clamped, and the calls to `updatePopup` in other instances are either a result of user interaction, which would pass valid values, or programmatic calls that are just passing along the current value or progress. This was also necessary to resolve the non-sensical popup values. For the issue with changing the `showPercentage` property, we actually pass a value to `updatePopup` that is either the value of progress for `ProgressBar`, or the `value` property for `Slider` via an overridden `showPercentageChanged` method. For the `popupSide` property, we do something similar with the `popupSideChanged` method.

### DIFF
--- a/src/Slider/Slider.js
+++ b/src/Slider/Slider.js
@@ -561,6 +561,22 @@ module.exports = kind(
 	}),
 
 	/**
+	* @override
+	* @private
+	*/
+	showPercentageChanged: function () {
+		this.updatePopup(this.value);
+	},
+
+	/**
+	* @override
+	* @private
+	*/
+	popupSideChanged: function () {
+		this.updatePopup(this.value);
+	},
+
+	/**
 	* @private
 	*/
 	valueChanged: function (was, is) {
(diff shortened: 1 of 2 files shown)


src/Slider/Slider.js
```diff
@@ -563,6 +563,20 @@ module.exports = kind(
 	/**
 	* @private
 	*/
+	showPercentageChanged: function () {
+		this.updatePopup(this.value);
+	},
+
+	/**
+	* @private
+	*/
```
viodragon2: Let's add @override documentation

PR #2787, opened 2016-03-23
ENYO-3239: set close button's spotlight to false when panel is changed

kbs12e:
If panels's animate is false and arranger is cardArranger, close button
is always spottable, then close button is spotlight focused when panel is
changed, so TV reads 'exit app' + panle title + focused item.

https://jira2.lgsvl.com/browse/ENYO-3239

### DIFF
--- a/src/Panels/Panels.js
+++ b/src/Panels/Panels.js
@@ -1295,7 +1295,7 @@ module.exports = kind(
 		// Turn on the close-x so it's ready for the next panel; if hasCloseButton is true
 		// and remove spottability of close button during transitions.
 		if (this.$.appClose) {
-			if (willAnimate) this.$.appClose.customizeCloseButton({'spotlight': false});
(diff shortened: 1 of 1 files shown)


aarontam:
Hi @kbs12e, we added this guard in https://github.com/enyojs/moonstone/pull/2776 to address the situation where we would have an unspottable close button when specifying a non-zero starting index at create time (the `finishTransition` method is never run to re-enable spotlight for the close button), but did not realize it would affect accessibility read order. How about instead of removing the guard altogether, we use a less-aggressive guard that more specifically addresses the situation I described, something like the following:
```
if (this.hasNode()) this.$.appClose.customizeCloseButton({'spotlight': false});
```

kbs12e:
Hi @aarontam I agree your opinion. I tested it again and #2787 issue is not reproduced

PR #2791, opened 2016-03-24
ENYO-3228: Force move focus to app close button on right key from input

KunmyonChoi:
Issue:
 App close button is a child of panels but not panel. So, when it is
 looking for nearest neighbor from input of header, it fails to find app
 close button.

Fix:
 We force move focus to app close button on right key press when focus
 is on input of header. Do the same when press left key on RTL.

### DIFF
--- a/src/Panels/Panels.js
+++ b/src/Panels/Panels.js
@@ -1031,11 +1031,17 @@ module.exports = kind(
 	},
 
 	/**
+	* Given a direction and starting control, walks up the lineage chain until a suitable control to
+	* spot has been found.
+	*
+	* @param {String} dir - The direction of movement.
+	* @param {Object} control - The starting control.
+	* @returns {Object} The target that should be spotted.
 	* @private
 	*/
 	getSpotlightTarget: function (dir, control) {
-		var target,
-			ref = control,
+		var ref = control,
+			target,
 			parent,
 			ext;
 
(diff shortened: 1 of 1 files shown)


src/Panels/Panels.js
```diff
@@ -1049,7 +1049,9 @@ module.exports = kind(
 		}
 		var orig = ev.originator,
 			idx = this.getPanelIndex(orig),
-			target = this.getSpotlightTarget('LEFT', orig);
+			target = this.getSpotlightTarget('LEFT', orig),
+			active = this.getActive(),
+			h = active ? active.getHeader() : null;
```
aarontam: Not sure we need the ternary here if all we're doing is a truthy check on `h` anyways.

src/Panels/Panels.js
```diff
@@ -1072,6 +1074,11 @@ module.exports = kind(
 				Spotlight.spot(this.$.breadcrumbs);
 				return true;
 			}
+		} else {
+			if (this.hasCloseButton && this.rtl === true && h && orig == h.$.inputDecorator) {
```
aarontam: It makes me uncomfortable that we are making an assumption about the component structure of the `Header` kind here (`h.$.inputDecorator`).

src/Panels/Panels.js
```diff
@@ -1134,6 +1148,9 @@ module.exports = kind(
 	spotlightFocus: function (oSender, oEvent) {
 		var orig = oEvent.originator;
 		var idx = this.getPanelIndex(orig);
+		if (oEvent.originator.owner === this.$.appClose) {
```
aarontam: We can use `orig` instead of `oEvent.originator` here and the next line.

src/Panels/Panels.js
KunmyonChoi: I thought that there are two separate cases.
- no active panel (No panel pushed) 
- no header (Custom panel case)

aarontam: I guess I was thinking we could do something like the following instead, but this is more of a stylistic thing and I don't feel super strongly:
```
h = active && active.getHeader();
```

webOS101:
Closing in favor of https://github.com/enyojs/moonstone/pull/2796

PR #2794, opened 2016-03-25
ENYO-3238: set panelsHanel to accessibilityDisabled when panels is hi…

kbs12e:
…dden.

When panels is hidden using left 5-way key, panelsHandle is spotlight focused,
because it could not find neighbor's spottable item in panels.
TV reads panelsHandle's aria-label when it is focused.
We added new observer property for preventing a reading panelsHanle label.

https://jira2.lgsvl.com/browse/ENYO-3238


aarontam:
Closing in favor of https://github.com/enyojs/moonstone/pull/2795.

PR #2801, opened 2016-03-28
ENYO-3238: Prevent handle readout when sender is appCloseButton

kbs12e:
If we left key prees from appCloseButton, panels handle is spotlight focused, so TV reads its label.
When appCloseButton is hidden, next spotlight target is panels handle. 

https://jira2.lgsvl.com/browse/ENYO-3238


lishammel:

PR #2813, opened 2016-04-06
Enyo 3309 yeramchoi

Djspaceg:


src/Notification/Notification.less
```diff
@@ -22,10 +27,10 @@
 		&::before {
 			content: '';
 			display: block;
-			background-color: @moon-neutral-bg-color;
-			border: 6px solid @moon-black;
-			border-radius: 30px;
-			opacity: 0.85;
```
Djspaceg: @yeram What was the reasoning for removing this element's opacity, only to replace it with the background color and border color alpha? Borders stack on top of background. Ex: a 50% alpha border on top of a 50% alpha background results in an effective 75% total opacity for the border.

src/Notification/Notification.less
```diff
@@ -1,4 +1,11 @@
 /* Popup.css */
+.moon {
+	.moon-notification {
+		padding: 27px 39px 27px 39px;
+		display: table;
+	}
+}
+
```
Djspaceg: I assume this is just for specificity precedence?

src/Notification/Notification.less
```diff
@@ -6,11 +13,9 @@
 	position: absolute;
 	left: 50%;
 	bottom: 0;
-	max-width: 70%;
-	padding: @moon-popup-padding;
+	max-width: 1083px;
```
Djspaceg: Lets make this 1080px

src/Notification/Notification.less
yeram: Hmm, I saw padding in .moon-notification is overridden by padding in .moon's.

http://10.186.120.46/~yeram.choi/dist/

src/Notification/Notification.less
yeram: got it!

src/Notification/Notification.less
yeram: Ah, I couldn't consider side effect of color alpha. I just implemented based on my experience about performance between opacity and color alpha.
I couldn't remember exact number of performance difference and don't know it is still presented, so it's better to keep your implementation for preventing side effect of color alpha.

lishammel:

PR #2818, opened 2016-04-20
ENYO-3332: Switch order of set disabled to avoid focus jump

KunmyonChoi:
Issue:
Slider have updateButtonStatus function that is set disabled
property for left/right buttons based on the value range.
When value is set from outside of control and focus is on one
of the left/right buttons, then it can trigger spotlight disappear
routine to find next focus control. But, it fails workaround focus
when next control is still disabled. Because, the sequence of
disabling controls is fixed as left to right.

Fix:
We switch disableing order by checking current disabled status.
Right button is disabled -> set disabled from right to left.
Left button is disabled -> set disabled from left to right.

### DIFF
--- a/src/Slider/Slider.js
+++ b/src/Slider/Slider.js
@@ -646,8 +646,13 @@ module.exports = kind(
 	*/
 	updateButtonStatus: function () {
 		if (this.enableJumpIncrement) {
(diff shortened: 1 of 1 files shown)


ryanjduffy:
We've merged 2.6.0-dev into master so this PR isn't applicable. We'll hold off on a master PR until it's necessary for another release.

PR #2820, opened 2016-04-20
ENYO-3260 : Change read sequence in popup, contexturePopup

changgilee:
In current status, when popup(or contexturePopup) is shown, read "button" + "title" sequence.

But UX guideline in Major MR, "title" + "button" sequence is right.

When onSpotlightContainerEnter is fired, the button in popup is not focused yet.

So, if change role to alert in this handler, "alert" + "dom.focus()" rule is applied and TTS call is not cut off.

Enyo-DCO-1.1-Signed-off-by: Changgi Lee changgi.lee@lge.com

### DIFF
--- a/src/ContextualPopup/ContextualPopup.js
+++ b/src/ContextualPopup/ContextualPopup.js
@@ -26,6 +26,8 @@ var
 	Scrim = require('moonstone/Scrim'),
 	HistorySupport = require('../HistorySupport');
 
+var options = require('enyo/options');
+
 /**
 * Fires when the contextual popup is to be shown.
 *
@@ -656,9 +658,7 @@ module.exports = kind(
 	*/
 	ariaObservers: [
 		{path: ['accessibilityReadAll', 'accessibilityRole', 'showing'], method: function () {
-			this.startJob('alert', function () {
-				this.setAriaAttribute('role', this.accessibilityReadAll && this.showing ? 'alert' : this.accessibilityRole);
-			}, 100);
+			this.updateAriaRole();
 		}}
 	],
 
@@ -666,8 +666,15 @@ module.exports = kind(
 	* @private
 	*/
 	onEnter : function(oSender, oEvent){
-		if (oEvent.originator == this){
-			this.setAriaAttribute('role', this.accessibilityReadAll && this.showing ? 'alert' : this.accessibilityRole);
+		if (options.accessibility && oEvent.originator == this) {
+			this.updateAriaRole();
 		}
+	},
+
+	/**
+	* @private
+	*/
+	updateAriaRole: function() {
+		this.setAriaAttribute('role', this.accessibilityReadAll && this.showing ? 'alert' : this.accessibilityRole);
 	}
 });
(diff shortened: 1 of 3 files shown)


src/Popup/Popup.js
```diff
@@ -596,5 +597,14 @@ module.exports = kind(
...
 			}, 100);
 		}}
-	]
+	],
+
+	/**
+	* @private
+	*/
+	onEnter : function(oSender, oEvent){
+		if (oEvent.originator == this){
```
kbs12e: I think that accessibility option checking is neccessary.

src/ContextualPopup/ContextualPopup.js
```diff
@@ -659,5 +660,14 @@ module.exports = kind(
...
 			}, 100);
 		}}
-	]
+	],
+
+	/**
+	* @private
+	*/
+	onEnter : function(oSender, oEvent){
+		if (oEvent.originator == this){
```
kbs12e: accessibility option checking is neccessary

src/Popup/Popup.js
```diff
@@ -596,5 +597,14 @@ module.exports = kind(
 				this.setAriaAttribute('role', this.accessibilityReadAll && this.showing ? 'alert' : this.accessibilityRole);
 			}, 100);
```
kbs12e: Basic popup, which has only content doesn't needs timer delay.

src/Popup/Popup.js
```diff
@@ -596,5 +597,14 @@ module.exports = kind(
 				this.setAriaAttribute('role', this.accessibilityReadAll && this.showing ? 'alert' : this.accessibilityRole);
```
kbs12e: This line is similar to onEnter() callback implementation.

src/Popup/Popup.js
changgilee: I will add it~ :)

src/ContextualPopup/ContextualPopup.js
changgilee: I will add it~ :)

src/Popup/Popup.js
changgilee: You are right~ I will remove timer delay~

src/Popup/Popup.js
changgilee: Can you give me more guidance what is wrong in this code? Thanks~

src/Popup/Popup.js
```diff
@@ -427,6 +430,7 @@ module.exports = kind(
 	* @public
 	*/
 	hide: function() {
+		this.set('_enableAlert', false);
```
kbs12e: If updateAriaRole() is called, this line is not neccessary.

src/Popup/Popup.js
```diff
@@ -591,10 +595,24 @@ module.exports = kind(
 	* @private
 	*/
 	ariaObservers: [
-		{path: ['accessibilityReadAll', 'accessibilityRole', 'showing'], method: function () {
-			this.startJob('alert', function () {
-				this.setAriaAttribute('role', this.accessibilityReadAll && this.showing ? 'alert' : this.accessibilityRole);
-			}, 100);
+		{path: ['accessibilityReadAll', 'accessibilityRole', 'showing', '_enableAlert'], method: function () {
```
kbs12e: If updateAriaRole() is called, _enableAlert observer is not neccessary.

lishammel:

PR #2823, opened 2016-04-21
ENYO-3326: Support accessibilityPreHint

kbs12e:
According to TV UX requirements, TV should support to reading a
hint message before reading control's content or aria-label, so
accessibilityPreHint API is added in enyo AccessibilitySupport,
we also applied it to Moonstone components.

https://jira2.lgsvl.com/browse/ENYO-3326

### DIFF
--- a/src/Drawers/Drawers.js
+++ b/src/Drawers/Drawers.js
@@ -523,13 +523,15 @@ module.exports = kind(
 	// Accessibility
 
 	ariaObservers: [
-		{path: ['accessibilityLabel', 'accessibilityHint', '_activated'], method: function() {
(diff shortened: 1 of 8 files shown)


KunmyonChoi:
Hi @kbs12e ,
Does video transport slider need to reset accesibilityPreHint as null?

KunmyonChoi:
Oh never mind on previous comment. 
It looks like slider itself is changing accessibilityHint and it needs to be reset as needed. But, accessibilityPreHint is never used in framework side. So, no problem.

lishammel:
Closing PR as jira has been closed

PR #2824, opened 2016-04-26
ENYO-3309: Added the new Notification Component

Djspaceg:


css/moonstone-variables.less
```diff
@@ -93,6 +93,7 @@
 @moon-expandable-value-font-size:   27px;
 @moon-expandable-client-item-font-size:   27px;
 @moon-divider-font-size:            24px;
+@moon-notification-font-size:       30px;
```
webOS101: This value appears to be duplicated below

src/Notification/Notification.js
```diff
@@ -0,0 +1,417 @@
+require('moonstone');
+
+/**
+* Contains the declaration for the {@link module:moonstone/Popup~Popup} kind.
```
webOS101: This should be Notification, not Popup

src/Notification/Notification.js
```diff
@@ -0,0 +1,417 @@
...
+
+	/**
+	* @private
+	*/
+	classes: 'moon moon-neutral enyo-unselectable moon-notification',
+
+	/**
+	* @private
+	*/
+	// floating: true,
```
webOS101: Delete this?

src/Notification/Notification.js
```diff
@@ -0,0 +1,417 @@
...
+	* @private
+	*/
+	allowDefault: true,
+
+	/**
+	* @private
+	*/
+	handlers: {
+		onRequestScrollIntoView   : '_preventEventBubble',
+		// ontransitionend           : 'animationEnd',
```
webOS101: remove

src/Notification/Notification.js
```diff
@@ -0,0 +1,433 @@
...
+	Button = require('../Button'),
+	BodyText = require('../BodyText'),
+	HistorySupport = require('../HistorySupport');
+
+/**
+* {@link module:moonstone/Notification~Notification} is a toast-like minimal popup that comes up
+* from the bottom of the screen.
+*
+* @class Notification
+* @ui
```
webOS101: Can you add an `@extends module:enyo/Popup~Popup`?

src/Notification/Notification.js
```diff
@@ -0,0 +1,433 @@
...
+		* The message that will be displayed in the notification's text area.
+		*
+		* @type {String}
+		* @default ''
+		* @public
+		*/
+		content: '',
+
+		/**
+		* When `true`, the notification will animate on/off screen.
```
webOS101: I think this description needs an update.

src/Notification/Notification.js
```diff
@@ -0,0 +1,433 @@
...
+		/**
+		* If `true`, {@glossary Spotlight} (focus) cannot leave the area of the notification unless
+		* the notification is explicitly closed; if `false`, spotlight may be moved anywhere within
+		* the viewport.
+		*
+		* @type {Boolean}
+		* @default true
+		* @public
+		*/
+		spotlightModal: true,
```
webOS101: I don't think this is hooked up to anything yet, is it?

Djspaceg: It's evidently implemented in Spotlight. It has no visible wiring in both enyo/Popup and moonstone/Popup.

css/moonstone-variables.less
Djspaceg: Got it. Good catch.

src/Notification/Notification.js
aarontam: As a general note and/or FYI, `spotlightDisabled` has a similar behavior of only being referenced in `Spotlight`.

src/Notification/Notification.js
```diff
@@ -0,0 +1,351 @@
...
+		if (anim) {
+			this.getComputedStyleValue('display');
+			this.set('animate', anim);
+		}
+	},
+
+	/**
+	* @private
+	*/
+	release: function () {
```
aarontam: Think we can get rid of this, as the implementation is the same as the base kind's.

src/Notification/Notification.js
```diff
@@ -0,0 +1,351 @@
...
+	/**
+	* @private
+	*/
+	handlers: {
+		onRequestScrollIntoView   : '_preventEventBubble'
+	},
+
+	/**
+	* @private
+	* @lends module:moonstone/Popup~Popup.prototype
```
aarontam: Should reference `Notification` here, instead. Also, we were moving away from published properties - can we just declaring these as properties directly on the kind?

src/Notification/Notification.js
```diff
@@ -0,0 +1,351 @@
...
+		}
+	},
+
+	/**
+	* @private
+	*/
+	hidden: function (sender, ev) {
+		this.respotActivator();
+
+		if (this._initialized && this.allowBackKey && !EnyoHistory.isProcessing()) {
```
aarontam: What do you think about checking for something like `this.hasNode()` here, instead of maintaining an internal property? Would also allow us to get rid of the `render` override.

Djspaceg: This code was actually copied directly from moonstone/Popup, so we should ideally update that too. I'll swap out `_initialized` for `this.hasNode()` or `this.generated` or something to that same effect.

src/Notification/Notification.js
Djspaceg: I was initially under the impression we were deprecating `published` but saw no evidence of it, outside my own work. I used it here because it seemed like nobody was deeply interested in not having it any more. I can take it out, but it will just make this control look different from others. Thoughts?

lishammel:
Closing PR as the jira has been closed.

PR #2829, opened 2016-05-12
ENYO-3348: Read full string for current value on DayPicker

KunmyonChoi:
### Issue:
Current value of DayPicker is showing abbreviation even when each item
is showing full string. This is to show as many days as possible in
small control space. When current value is 'Mon, Tue, Thu, Sat', it
reads 'Day Picker Mon, Tuesday, Thursday, Sat'.

### Fix:
We bind accessibilityText to accessibilityLabel of header. Update
accessibilityText by concatenate content of each items whenever
currentValue changes. Also, it concatenate accessibilityLabel for
item when specified. This also fix the problem of read current
value of ExpandablePicker when multipleSelection is true. (ENYO-3355)

### DIFF
--- a/src/ExpandablePicker/ExpandablePicker.js
+++ b/src/ExpandablePicker/ExpandablePicker.js
@@ -209,7 +209,7 @@ module.exports = kind(
 		{from: 'selected.content', to: 'selectedText'},
 
 		// Accessibility
-		{from: 'selected.accessibilityLabel', to: '$.header._accessibilityText'}
+		{from: 'accessibilityText', to: '$.header._accessibilityText'}
 	],
 
 	computed: {
@@ -241,7 +241,8 @@ module.exports = kind(
 	},
 
 	/**
-	*  'multiSelectCurrentValue()' can be overridden by subkinds, such as moon.DayPicker
+	*  'multiSelectCurrentValue()' can be overridden by subkinds, such as moon.DayPicker.
+	*  Returns label for selected items by concatenate content of each items.
 	*
 	* @protected
 	*/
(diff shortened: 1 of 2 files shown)


src/ExpandablePicker/ExpandablePicker.js
```diff
@@ -266,6 +267,30 @@ module.exports = kind(
...
 
 	/**
+	*  'multiSelectCurrentAccesibilityLabel()' can be overridden by subkinds, such as moon.DayPicker.
+	*  Returns accessibilityLabel for selected items by concatenate accessibilityLabel of each items.
+	*
+	* @protected
+	*/
+	multiSelectCurrentAccesibilityLabel: function () {
+		if (!this.multipleSelection) {
+			return;
```
seunghoh: We also consider single selection case. Is there any other place to take care of single selection case? I can't find

src/ExpandablePicker/ExpandablePicker.js
```diff
@@ -209,7 +209,7 @@ module.exports = kind(
 		{from: 'selected.content', to: 'selectedText'},
 
 		// Accessibility
-		{from: 'selected.accessibilityLabel', to: '$.header._accessibilityText'}
+		{from: 'accessibilityText', to: '$.header._accessibilityText'}
```
seunghoh: accessibilityText looks like public API very much. Would you please change this private val with _ prefix or use totally different name?

src/ExpandablePicker/ExpandablePicker.js
```diff
@@ -346,6 +371,15 @@ module.exports = kind(
 	},
 
 	/**
+	* Update header accessibilityLabel whenever currentValueText changed.
+	*
+	* @private
+	*/
+	currentValueTextChanged: function () {
+		this.set('accessibilityText', this.multiSelectCurrentAccesibilityLabel());
```
seunghoh: Please check accessibility option so that we can add this line only for accessibility option ON

src/ExpandablePicker/ExpandablePicker.js
KunmyonChoi: Added single selection case on currentValueTextChanged.

yeram:
moved to rc.6 branch. see #2841

PR #2837, opened 2016-05-18
ENYO-3331: Apply incremental option to popup animation

yeram:
Although there is an incremental option popup still shows unexpecting point number.
To resolve this, I override calcPopupLabel() to make corresponding values.


KunmyonChoi:
Please add signed off statement on PR.

PR #2843, opened 2016-06-01
PLAT-20930: Prevent reading a invalid tooltip message.

kbs12e:
Prevent reading a invalid message when tooltip is shown in range input.
Tooltip is popup type component, so TV reads tooltip content when tooltip is appeared.
According to UX team, TV should not read this tooltip message.

https://jira2.lgsvl.com/browse/PLAT-20930


lishammel:
Closing PR as Jira is closed.

PR #2846, opened 2016-06-10
ENYO-3366 Accessibility: aria-label is set to uppercased string

Sangwook-Lee:
### Issue
In the moonstone button case, it has not aria-label. So, aria-label is set to upper-cased string.
If uppercased content is read from TTS engine, it's pronounciation can be wrong.
For example, you can see this sentence ("Reset a device").
In this case, 'a' is article grammartically, not captial 'A'.

### Fix
It is to fix issue about above accessibility.
I modified button, tooltip and dialog for setting 'aria-label' from origin content.

ENYO-3366 Accessibility: aria-label is set to uppercased string

### DIFF
--- a/src/Button/Button.js
+++ b/src/Button/Button.js
@@ -291,5 +291,19 @@ module.exports = kind(
 		if (opacity == 'translucent' || opacity == 'transparent') {
 			this.addClass(opacity);
 		}
-	}
+	},
+
+	// Accessibility
+
+	/**
+	* @private
+	*/
+	ariaObservers: [
+		{path: ['content'], method: function () {
+			if (!this.accessibilityLabel) {
+				var content = this.content;
+				this.$.client.setAriaAttribute('aria-label', content);
+			}
+		}}
+	]
 });
--- a/src/Dialog/Dialog.js
+++ b/src/Dialog/Dialog.js
@@ -238,6 +238,12 @@ module.exports = kind(
 	},
 
 	ariaObservers: [
+		{from: 'title', method: function () {
+			if (!this.accessibilityLabel) {
+				var content = this.getTitle();
+				this.$.title.set('accessibilityLabel', content);
+			}
+		}},
 		{from: 'generated', method: function () {
 			if (!this.generated) return;
 
--- a/src/Tooltip/Tooltip.js
+++ b/src/Tooltip/Tooltip.js
@@ -102,7 +102,21 @@ var TooltipContent = kind({
 	*/
 	uppercaseChanged: function (was, is) {
 		this.contentChanged();
-	}
+	},
+
+	// Accessibility
+
+	/**
+	* @private
+	*/
+	ariaObservers: [
+		{path: ['content'], method: function () {
+			if (!this.accessibilityLabel) {
+				var content = this._content;
+				this.setAriaAttribute('aria-label', content);
+			}
+		}}
+	]
 });
 
 /**


src/Dialog/Dialog.js
```diff
@@ -184,6 +184,7 @@ module.exports = kind(
 	*/
 	titleChanged: function () {
 		var title = this.getTitle();
+		this.$.title.set('accessibilityLabel', title);
 		this.$.title.setContent( this.get('uppercase') ? util.toUpperCase(title) : title );
 	},
```
yeram: how about adding this feature as ariaObservers like others?
ex)
ariaObservers: [
    {path: ['title'], method: function () {
        var content = this.getTitle();
        this.setAriaAttribute('aria-label', content);
    }}
]

src/Button/Button.js
```diff
@@ -291,5 +291,17 @@ module.exports = kind(
...
+
+	// Accessibility
+
+	/**
+	* @private
+	*/
+	ariaObservers: [
+		{path: ['content'], method: function () {
+			var content = this.content;
+			this.$.client.setAriaAttribute('aria-label', content);
```
yeram: Does it guarantee timing in case of developer set accessibility label and content dynamically?

ex 1) if we couldn't guarantee this, we need to consider reference guide that developers should use following sequence reversely like ex2.
`
this.set('accessibilityLabel', 'a11y'); // we should pick this to use a11y label prior than following content
`
`
this.set('content', 'normal');
`

ex 2) I think it's ok for this case, but need to check
`
this.set('content', 'normal');
`
`
this.set('accessibilityLabel', 'a11y'); // we should pick this to use a11y label prior than following content
`

src/Tooltip/Tooltip.js
```diff
@@ -102,7 +102,19 @@ var TooltipContent = kind({
...
+	},
+
+	// Accessibility
+
+	/**
+	* @private
+	*/
+	ariaObservers: [
+		{path: ['content'], method: function () {
+			var content = this._content;
```
yeram: I think it's okay to use 'this.content' since 'ariaObservers' are triggered prior than 'somethingChanged'.
Could you check timing?

src/Dialog/Dialog.js
Sangwook-Lee: I'll modify like below:
{from: 'title', method: function () {
	if (!this.accessibilityLabel) {
		var content = this.getTitle();
		this.$.title.setAriaAttribute('aria-label', content);
	}
}}

src/Button/Button.js
```diff
@@ -291,5 +291,19 @@ module.exports = kind(
...
+	},
+
+	// Accessibility
+
+	/**
+	* @private
+	*/
+	ariaObservers: [
+		{path: ['content'], method: function () {
+			if (!this.accessibilityLabel) {
```
seunghoh: What if app only applys 'accessibilityHint' or 'accessibilityPreHint'?

src/Tooltip/Tooltip.js
```diff
@@ -102,7 +102,21 @@ var TooltipContent = kind({
...
+	},
+
+	// Accessibility
+
+	/**
+	* @private
+	*/
+	ariaObservers: [
+		{path: ['content'], method: function () {
+			if (!this.accessibilityLabel) {
```
seunghoh: What if app only applys 'accessibilityHint' or 'accessibilityPreHint'?

PR #2852, opened 2016-06-21
ENYO-3344: add isChrome options when creating tools

yeram:
We didn't expect client, closeButton as returns of getClientControls().
To get corresponding return, we added isChrome option when creating tools.


KunmyonChoi:
LGTM!

PR #2863, opened 2016-06-23
ENYO-3309: Change undefined less variable to real value

yeram:
moon-font-family-small wasn't defined in moonstone but moonstone-extra.
It caused less compile fail. So I changed that variable to real value temporally.


Sangwook-Lee:
Reviewed-By: Sangwook-Lee (sangwook1203.lee@lge.com)

KunmyonChoi:
LGTM!

PR #2865, opened 2016-07-01
ENYO-3382: Change focus timing of notification to 'before start anima…

KunmyonChoi:
…tion'

Issue:
We give default focus to button on notification when animation ends.
This behavior differs from popup or dialog which is giving focus when
start animation. The focus timing change gave a chance of focus lost.

Fix:
We change focus timing to 'before start animation'.

### DIFF
--- a/src/Notification/Notification.js
+++ b/src/Notification/Notification.js
@@ -86,12 +86,12 @@ module.exports = kind(
 	/**
 	* @private
 	*/
-	shownMethod: 'shown',
+	showingMethod: 'beforeShow',
 
 	/**
 	* @private
 	*/
-	hiddenMethod: 'hidden',
+	hidingMethod: 'beforeHide',
 
 	/**
 	* @private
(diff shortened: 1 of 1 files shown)


src/Notification/Notification.js
```diff
@@ -210,16 +210,27 @@ module.exports = kind(
...
-	shown: function (sender, ev) {
+	showingChanged: kind.inherit(function (sup) {
+		return function (sender, ev) {
+			if (this.showing && this.showingTransitioning) {
+				// ENYO-3382: hide animation is not finished, call popup hide directly before show again
+				this.showing = !this.showing;
+				Popup.prototype.showingChanged.apply(this, arguments);
+				this.showing = !this.showing;
+			}
+			sup.apply(this, arguments);
```
aarontam: Probably should be `Popup.prototype.showingChanged.apply(this, arguments);`.

KunmyonChoi: You mean remove this.showing = !this.showing; ?
I added these lines because popop.showingchanged with opposite showing value should be called before sup call.

KunmyonChoi: I think showingTransitionSupport.showingChanged should complete previous 'showingTransition' job on showingTransitionSupport, not here. So, closing this PR and add fix on showingTransitionSupport.

PR #2866, opened 2016-07-01
ENYO-3390: Flip start/end compare position on rtl case in moon.input

KunmyonChoi:
Issue:
When user press left or right key in moon.Input, it compares the cursor
position to 0 or length of text and move spotlight focus to other
control to that direction. Problem happens on rtl character. It compares
start position to 0 but it should compare to length.

Fix:
When rtl character is written in moon.Input, flip the start and end
position of compare. So it can safely move focus to the right direction.

### DIFF
--- a/src/Input/Input.js
+++ b/src/Input/Input.js
@@ -161,7 +161,7 @@ module.exports = kind(
 	*/
 	right: function () {
 		var end = this.rtl ? 0 : this.node.value.length;
-		if (!this.hasNode() || this.node.selectionStart == end) {
+		if (!this.hasNode() || this.node.selectionStart === end) {
 			return false;
 		}
 		return true;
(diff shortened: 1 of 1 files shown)


src/Input/Input.js
```diff
@@ -159,7 +160,8 @@ module.exports = kind(
 	* @private
 	*/
 	right: function () {
-		if (!this.hasNode() || this.node.selectionStart == this.node.value.length) {
+		var end = this.rtl ? 0: this.node.value.length;
```
yeram: can we add white space between 0 and : ?

src/Input/Input.js
```diff
@@ -159,7 +160,8 @@ module.exports = kind(
...
+		if (!this.hasNode() || this.node.selectionStart == end) {
```
yeram: what about === instead of ==

src/Input/Input.js
KunmyonChoi: done! thanks!

src/Input/Input.js
KunmyonChoi: Done!

PR #2867, opened 2016-07-04
ENYO-3391 : Marquee is not working when reflow is happen

changgilee:
Issue
:Marquee status change to ellipse but animation is not working

Fix
:Call marquee reset func in reflow handler


KunmyonChoi:
Code looks good. But, I have one comment. 
The marquee item _marquee_reset is actually calls marquee support _resetMarquee.
This will call more than what it needs when many marquee item is having one marquee support.
So, I recommend to call marquee support resetMarquee from marquee support reflow handler.

PR #2871, opened 2016-07-08
ENYO-3379: Set ExpandableListItem spotlight property as false

KunmyonChoi:
Issue:
We added Scroll to boundary feature on ScrollStrategy which is checking
5way limit with its first spottable parent. When spotlight container
control is used in scroller with spottable child inside, it is detect
there is no nearest neighbor from the child and try to jump scroll.

This behavior is okay with other case but it is not when child items are
small number like ExpandableListItem Header case. It is try to jump
scroll for every 5way key.

Fix:
Set nearest neighbor root as focusable parent that is having more than
1 children.

### DIFF
--- a/src/ScrollStrategy/ScrollStrategy.js
+++ b/src/ScrollStrategy/ScrollStrategy.js
@@ -1030,7 +1030,12 @@ var MoonScrollStrategy = module.exports = kind(
 	* @private
 	*/
 	at5WayLimit: function (control, direction) {
-		return !Spotlight.NearestNeighbor.getNearestNeighbor(direction, control, {root: Spotlight.getParent(control)});
+		// Find parent control which is having more than 1 children
+		var parent = Spotlight.getParent(control);
+		while (parent != this && Spotlight.getChildren(parent).length <= 1) {
+			parent = Spotlight.getParent(parent);
+		}
+		return !Spotlight.NearestNeighbor.getNearestNeighbor(direction, control, {root: parent});
 	},
 
 	/**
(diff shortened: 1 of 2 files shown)


ryanjduffy:
cc: @aarontam since you made the change (24ca411a92cd278400c81ac205dac5f22fee52af) to make it a container

aarontam:
Hi @KunmyonChoi, this change will cause a regression with Expandables + back key, as the history state will be out of sync, particularly when navigating in and out of these controls in 5-way mode. This is due to a quirk where multiple `onSpotlightContainerEnter` events are fired in these cases, and making the wrapping control (i.e. the entire `ExpandableListItem` control) a Spotlight container seemed to be the least-intrusive way to address this (as opposed to making a general change to how Spotlight containers behave).

KunmyonChoi:
Hi @ryanjduffy and @aarontam, thanks for checking.
I personally considered this issue as a special case that doesn't want to get advantage of Scroll to boundary feature. So, I could think of options: 1) passing by the scroll to boundary logic 2) apply scroll to boundary only when the container contains more than one focus-able items. how do you think?

src/ScrollStrategy/ScrollStrategy.js
```diff
@@ -1030,7 +1030,12 @@ var MoonScrollStrategy = module.exports = kind(
 	* @private
 	*/
 	at5WayLimit: function (control, direction) {
-		return !Spotlight.NearestNeighbor.getNearestNeighbor(direction, control, {root: Spotlight.getParent(control)});
+		// Find parent control which is having more than 1 children
+		var parent = Spotlight.getParent(control);
+		while (parent != this && Spotlight.getChildren(parent).length <= 1) {
```
aarontam: We should use an identity check here i.e. `parent !== this`.

PR #2872, opened 2016-07-12
ENYO-3396: Revert dialog layout

KunmyonChoi:
Revert client position to right side of title.
Revert divider and set default showing as true.
Revert padding and margin.

Enyo-DCO-Signed-off-by: Kunmyon Choi (kunmyon.choi@lge.com)

### DIFF
--- a/src/Dialog/Dialog.js
+++ b/src/Dialog/Dialog.js
@@ -16,6 +16,7 @@ var
 	Popup = require('../Popup'),
 	IconButton = require('../IconButton'),
 	BodyText = require('../BodyText'),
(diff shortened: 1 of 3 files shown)


src/Dialog/Dialog.js
```diff
@@ -16,6 +16,7 @@ var
 	Popup = require('../Popup'),
 	IconButton = require('../IconButton'),
 	BodyText = require('../BodyText'),
+	Divider = require('../Divider'),
```
yeram: travis-ci build failed due to this line. are we assuming to use classes of divider?

PR #2879, opened 2016-07-14
ENYO-3412: Assign allowHtml as true by default

yeram:
Previously, Dialog supported allowHtml as true by default.
It was caused by not binding Dialog's allowHtml to BodyText.

In case of Notification we binded its allowHtml to BodyText and it caused
settings BodyText's prop to false by unexpectation.


KunmyonChoi:
LGTM

PR #2883, opened 2016-07-20
ENYO-3427 : Button label did not read when Notification Popup is shown.

changgilee:
Issue
:When Notification Popup is opened,
Button focus is called straightly, but Alert is called after 100ms.
So, Alert message is cut off button's tts audio guidance.

Fix
:Change Alert called timing to onSpotlightContainerEnter event handler

### DIFF
--- a/src/Notification/Notification.js
+++ b/src/Notification/Notification.js
@@ -21,7 +21,8 @@ var
 	BodyText = require('../BodyText'),
 	HistorySupport = require('../HistorySupport');
 
-var options = require('enyo/options');
+var
+	options = require('enyo/options');
 
 /**
 * {@link module:moonstone/Notification~Notification} is a toast-like minimal popup that comes up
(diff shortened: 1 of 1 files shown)


src/Notification/Notification.js
```diff
@@ -21,6 +21,8 @@ var
 	BodyText = require('../BodyText'),
 	HistorySupport = require('../HistorySupport');
 
+var options = require('enyo/options');
```
yeram: It looks better to add indent here likely Spotlight.

src/Notification/Notification.js
```diff
@@ -417,9 +419,23 @@ module.exports = kind(
...
-			}, 100);
+			this.updateAriaRole();
 		}}
-	]
+	],
+
+	/**
+	* @private
+	*/
+	onEnter : function(oSender, oEvent) {
```
yeram: Please remove white-space between onEnter and :, and add between function and (

src/Notification/Notification.js
```diff
@@ -417,9 +419,23 @@ module.exports = kind(
...
+		if (options.accessibility && oEvent.originator == this) {
+			this.updateAriaRole();
+		}
+	},
+
+	/**
+	* @private
+	*/
+	updateAriaRole: function() {
```
yeram: add white-space between function and ( . Thanks!

yeram:
LGTM

PR #2885, opened 2016-07-21
ENYO-3428 ListActions: arrow position is wrong when it use with IconButton

Sangwook-Lee:
### Issue
In photo&video app case, arrow of contextualPopup(ListActions) has wrong position.
It is occurred when it use with IconButton. There are 2 problems.
First, in appliation side, app uses additional components block. So, 'moon-hspacing' of header clients is overrided.
Secondly, in framework side, 'moon-hspacing' does not cover the tooltip-decorator case.

### Fix
It is fixed that case about 'tooltip-decorator' is added. Also, first-cheild of decorator has margin 0.
Moreover, It is added code which can cover the case about that 'hspacing' has contact with another 'hspacing'.

ENYO-3428 ListActions: arrow position is wrong when it use with IconButton

### DIFF
--- a/css/moonstone-spacing.less
+++ b/css/moonstone-spacing.less
@@ -13,22 +13,23 @@
 	&.top > * {
 		vertical-align: top;
 	}
-	& > :first-child {
+	& > :not([style*="display:none"]):not([style*="display: none"]):first-of-type {
 		margin-left: 0;
 	}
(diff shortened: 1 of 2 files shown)


KunmyonChoi:
I'm not convinced with current adjust logic. But, I'm okay with merge this fix for now because it can cover most of the ux use cases.

PR #2887, opened 2016-07-21
ENYO-3431 Scrim isn't hidden when notification is destroyed in showing

Sangwook-Lee:
### Issue
Scrim is not hidden when notification is destroyed in showing status.

### Fix
When notification is destroyed, showHideScrim is called to hide scrim.

ENYO-3431 Scrim isn't hidden when notification is destroyed in showing

### DIFF
--- a/src/Notification/Notification.js
+++ b/src/Notification/Notification.js
@@ -194,14 +194,6 @@ module.exports = kind(
 	/**
 	* @private
 	*/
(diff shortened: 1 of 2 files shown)


yeram:
LGTM

KunmyonChoi:
It looks like destroy on enyo/popup also needs same logic. So, suggest that move this to enyo.

PR #2888, opened 2016-07-22
ENYO-3429: Fix spinner text color on neutral theme

KunmyonChoi:
Set Spinner text color as white on dark background.
Set Spinner text color as dark gray on transparent background.


yeram:
LGTM

PR #2891, opened 2016-07-22
ENYO-3389: Set overflow as hidden to client control

yeram:
Many buttons can cause overflow on Notification control.
We guard this by setting overflow as hidden to client per UX requirement.


KunmyonChoi:
LGTM

PR #2895, opened 2016-07-25
ENYO-3437: Correct to corresponding condition and add guard condition

yeram:
We got stuck in some scenario due to there is in-valid condition in
while statement. We fix in-valid condition and add guard condition
to avoid excpetional cases.


KunmyonChoi:
LGTM!

PR #2900, opened 2016-08-08
ENYO-3438 : Apply non-latin font in Notification Popup

changgilee:
Issue
: When japanese language is selected, non-latin(LG Display) has to applied in Notification Popup's body text.

Fix
: Add optional code for  apply non-latin font.

### DIFF
--- a/src/Notification/Notification.less
+++ b/src/Notification/Notification.less
@@ -61,6 +61,9 @@
 		font-weight: @moon-notification-font-weight;
 		font-style: @moon-notification-font-style;
 		font-size: @moon-notification-font-size;
+		.enyo-locale-non-latin & {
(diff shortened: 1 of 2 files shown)


src/Notification/Notification.less
```diff
@@ -61,6 +61,9 @@
 		font-weight: @moon-notification-font-weight;
 		font-style: @moon-notification-font-style;
 		font-size: @moon-notification-font-size;
+		.enyo-locale-non-latin & {
+			font-family: @moon-non-latin-font-family;
```
KunmyonChoi: We gathering enyo-locale-non-latin on moonstone-text.less file. I recommand to move this into that just like what moon-body-text did.

PR #2905, opened 2016-08-18
ENYO-3446: CSS adjustment

yeram:
There were changes on style of Notification control.
I adjusted CSS to meet latest GUI redlines.

### DIFF
--- a/css/moonstone-variables.less
+++ b/css/moonstone-variables.less
@@ -198,7 +198,7 @@
 @moon-notification-background-opacity: 95%;
 @moon-notification-border-opacity: 30%;
 @moon-notification-border-width: 6px;
-@moon-notification-border-radius: 24px;
+@moon-notification-border-radius: 15px;
 
 // Drawer
 // ---------------------------------------
--- a/css/moonstone-variables.less
+++ b/css/moonstone-variables.less
@@ -199,6 +199,7 @@
 @moon-notification-border-opacity: 30%;
 @moon-notification-border-width: 6px;
 @moon-notification-border-radius: 15px;
+@moon-notification-button-gap: 9px;
 
 // Drawer
 // ---------------------------------------
--- a/src/Notification/Notification.less
+++ b/src/Notification/Notification.less
@@ -1,31 +1,29 @@
 /* Notification.css */
-.moon {
-	.moon-notification {
-		padding: 24px 30px;
-	}
-}
-
 .moon-notification {
 	position: absolute;
 	left: 50%;
-	bottom: 0;
+	bottom: 324px;
 	min-width: 300px;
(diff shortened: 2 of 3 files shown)


KunmyonChoi:
When we assume that button can be used maximum 4, then we should set max-width as 1227px. (300+9+300+9+300+9+300)

css/moonstone-variables.less
```diff
@@ -198,7 +198,7 @@
 @moon-notification-background-opacity: 95%;
 @moon-notification-border-opacity: 30%;
 @moon-notification-border-width: 6px;
-@moon-notification-border-radius: 24px;
+@moon-notification-border-radius: 15px;
```
KunmyonChoi: I don't see any specification for border-radius on GUI redline.

src/Notification/Notification.less
```diff
@@ -20,12 +14,20 @@
 	-ms-transform:     translate3d(-50%, 100%, 0);
 	transform:         translate3d(-50%, 100%, 0);
 
+	&.wide {
+		max-width: 1227px; // 4 button case
+	}
+
+	.moon& {
```
KunmyonChoi: This should have space between .moon and &.

src/Notification/Notification.less
```diff
@@ -20,12 +14,20 @@
...
+		padding: 24px 30px;
+	}
+
 	&::before {
 		content: '';
 		display: block;
 		position: absolute;
 		border: @moon-notification-border-width solid fade(@moon-black, @moon-notification-border-opacity);
-		border-radius: (@moon-notification-border-radius + @moon-notification-border-width);
+		border-radius: @moon-notification-border-radius;
```
KunmyonChoi: How border-radius should change if we increase @moon-notification-border-width?

src/Notification/Notification.less
```diff
@@ -50,10 +52,10 @@
 	&.showing,
 	&.shown {
 		// Vertically move the full height of the notification box, plus an additional 168px
```
KunmyonChoi: Does this need to be updated?

src/Notification/Notification.less
```diff
@@ -68,10 +70,33 @@
...
+		margin-left: 9px;	/* gap between buttons should be 9px per notification spec */
+		margin-right: 0px;
+		.enyo-locale-right-to-left & {
+			margin-left: 0px;
+			margin-right: 9px;
+		}
+	}
+	.moon-hspacing {
+		& > :first-child,
+		& > :first-child.moon-contextual-popup-decorator :first-child {
```
KunmyonChoi: Do we have any use case that is using contextual popup decorator on dialog popup?

PR #2906, opened 2016-08-19
PLAT-24872: ListActions lists are empty

anishlg:
Issue: ListAction component menus are not displaying properly in mobile portrait mode.
Cause: In portrait mode the menu will be stacked and the if the menu items having list and have not set the height will not be displayed as expected.
Fix: For stacked menus, if the child is list. added the css min-height

Enyo-DCO-1.1-Signed-off-by: Anish T D(anish.td@lge.com)

### DIFF
--- a/src/ListActions/ListActions.less
+++ b/src/ListActions/ListActions.less
@@ -79,6 +79,9 @@
 
 	&.stacked .moon-list-actions-menu {
 		display: block;
+		> [role="listbox"] {
+			min-height: 250px;
+		}
 	}
 }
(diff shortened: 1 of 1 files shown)


src/ListActions/ListActions.less
```diff
@@ -79,6 +79,9 @@
 
 	&.stacked .moon-list-actions-menu {
 		display: block;
+		> [role="listbox"] {
```
KunmyonChoi: I'm wondering why select using role here.

lishammel:

PR #2908, opened 2016-08-20
ENYO-3447: Define .moon-word-break and use it in Notification

yeram:
It's better to have word-break option to display context in various locales.
At this moment, we only applied it to BodyText in Notification.
For the expandability of word-break style, we separate it as an individual
css prop to use it in other controls.

### DIFF
--- a/css/moonstone-text.less
+++ b/css/moonstone-text.less
@@ -116,6 +116,10 @@
 	font-size: @moon-divider-font-size;
 	color: @moon-divider-text-color;
 }
+.moon-word-break {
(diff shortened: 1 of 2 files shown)


css/moonstone-text.less
```diff
@@ -197,4 +201,4 @@
 		font-size: @moon-non-latin-button-small-font-size;
 		font-weight: normal;
 	}
-}
\ No newline at end of file
+}
```
Sangwook-Lee: Is new line entered?

PR #2909, opened 2016-08-23
[ENYO-3441] DayPicker: Support special separator for fa-IR locale

Sangwook-Lee:
This is requirement from 'Gold Iran'.
They want to support special delimiter to moon.DayPicker.

Add method for getting delimiter(,) depands on Locales.

ENYO-3441 DayPicker: Support special separator for fa-IR locale


KunmyonChoi:
Code looks good. But, we may need to add the move of tools control block on release note.

PR #2911, opened 2016-08-24
ENYO-3456: Set filter as blur(0) to avoid chromium bug

yeram:
There was a bug on chromium side that set blur to font forcefully in
case of div's position is odd number.
To avoid this, we set blur(0) to -webkit-filter prop.

### DIFF
--- a/src/Notification/Notification.less
+++ b/src/Notification/Notification.less
@@ -15,6 +15,7 @@
 	text-align: center;
 	border-radius: @moon-notification-border-radius;
 	background-color: fade(@moon-neutral-bg-color, @moon-notification-background-opacity);
+	-webkit-filter: blur(0);
 	-webkit-transform: translate3d(-50%, 100%, 0);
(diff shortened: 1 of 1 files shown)


src/Notification/Notification.less
```diff
@@ -15,6 +15,7 @@
 	text-align: center;
 	border-radius: @moon-notification-border-radius;
 	background-color: fade(@moon-neutral-bg-color, @moon-notification-background-opacity);
+	-webkit-filter: blur(0);
```
webOS101: Somewhat concerned about this, especially considering that it appears a blur of 0 still applies blur on Retina macbooks and iOS devices (In some cases?  Seems fixed on Blink.  Safari gives odd results:  http://stackoverflow.com/questions/20377990/why-doesnt-blur0-remove-all-text-blur-in-webkit-chrome-on-retina-screens )

KunmyonChoi: We reverted this on 2.6.4-rc.9.1.

PR #2913, opened 2016-08-25
ENYO-3446: Set margin-right as 9px if locale is right to left

yeram:
There was a missing code that adding overrides for .enyo-locale-right-to-left
case. If we don't have that overriding prop, gap between buttons is 0.
We don't need to set margin-left as 0 in case of first-child since it's
already 0.

### DIFF
--- a/src/Notification/Notification.less
+++ b/src/Notification/Notification.less
@@ -92,6 +92,7 @@
 			margin-right: 0;
 		}
 		.enyo-locale-right-to-left & > :last-child {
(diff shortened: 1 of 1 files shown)


KunmyonChoi:
I see that the right most button also having gap on right side in rtl case. Is it okay?

KunmyonChoi:
We are focusing on button case for this time. And need to review other usecases and improve next time.

KunmyonChoi:
Lgtm!

PR #2915, opened 2016-09-05
ENYO-3473: Add new feature for auto position adjustment

yeram:
A fraction (non-integer) coordinate caused several rendering issues related to
antialiasing such as blurred text and rough edges.
To resolve this, we need to set div's position to Integer numbers,
and I added new feature to enable above feature by handy option.

### DIFF
--- a/src/Notification/Notification.js
+++ b/src/Notification/Notification.js
@@ -9,6 +9,7 @@ var
 	kind = require('enyo/kind'),
 	Control = require('enyo/Control'),
 	dispatcher = require('enyo/dispatcher'),
+	dom = require('enyo/dom'),
 	EnyoHistory = require('enyo/History'),
 	Popup = require('enyo/Popup'),
 	ShowingTransitionSupport = require('enyo/ShowingTransitionSupport');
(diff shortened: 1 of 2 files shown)


src/Notification/Notification.js
```diff
@@ -258,6 +273,31 @@ module.exports = kind(
 	/**
 	* @private
 	*/
+	refineAbsoluteBounds: function () {
+		var b = this.getAbsoluteBounds(),
+			l = b ? b.left : 0,
+			t = b ? b.top : 0,
+			transforms = {};
+
+		if (l % 1 !== 0 || t % 1 !== 0) {
```
KunmyonChoi: It looks like we should avoid any modification when l and t  are 0.

src/Notification/Notification.js
```diff
@@ -258,6 +273,31 @@ module.exports = kind(
...
+			transforms.translateX = Math.round(l) + 'px';
```
KunmyonChoi: Nevermind, this condition will block 0 cases anyway.

yeram:
resize solution(#2916) is more reliable than this.

PR #2916, opened 2016-09-06
ENYO-3473: Add new feature for auto position adjustment

yeram:
Floating numbers on bounds caused several issues relate to rendering issues especially making blurred text.

I added new feature to enable above feature by handy option.

### DIFF
--- a/src/Notification/Notification.js
+++ b/src/Notification/Notification.js
@@ -176,7 +176,16 @@ module.exports = kind(
 		* @default false
 		* @public
 		*/
-		wide: false
+		wide: false,
+
+		/**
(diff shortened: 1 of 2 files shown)


src/Notification/Notification.js
```diff
@@ -176,7 +176,16 @@ module.exports = kind(
...
+		wide: false,
+
+		/**
+		* If `true`, adjust size of control to even number after rendered by default.
+		*
+		* @type {Boolean}
+		* @default true
+		* @public
+		*/
+		autoSizeAdjustment: true
```
KunmyonChoi: I vote on postSizeAdjustment as a property.

PR #2928, opened 2016-10-11
ENYO-3554 : Prevent dummy tts call in Spinner

changgilee:
Apply accessibilityDisabled property in decorator module

Add translated loading message


src/Spinner/Spinner.js
```diff
@@ -243,5 +246,13 @@ module.exports = kind(
...
-	accessibilityLive: 'off'
+	accessibilityLive: 'off',
+
+	/**
+	* @default Loading
+	* @type {String}
+	* @translated loading message
+	* @public
+	*/
+	loadingMessage: $L('Loading')
```
KunmyonChoi: If you intended to use this as a constant that can be used when set content no app side, this can be not a variable but a constant. So, I recommend to explicitly export the constant on Spinner module like below. You can reference LightPanels code that exporting Direction.

```
var
	$L = require('../i18n'),
	loadingString = $L('Loading');

mosule.exports.LOADING_STRING = loadingString;
```

src/Spinner/Spinner.js
```diff
@@ -243,5 +246,13 @@ module.exports = kind(
 	* @see enyo/AccessibilitySupport~AccessibilitySupport#accessibilityLive
 	* @public
 	*/
-	accessibilityLive: 'off'
+	accessibilityLive: 'off',
+
+	/**
+	* @default Loading
```
KunmyonChoi: Public property should have well documented comment. I think this document should also include sample code for how do you expect app developer to use this constant.

src/Spinner/Spinner.js
KunmyonChoi: App developer can use this exported constant like below code from app side.
```
var Spinner = require('moonstone/Spinner');

set('content', Spinner.LOADING_STRING);
```

src/Spinner/Spinner.js
```diff
@@ -248,11 +249,16 @@ module.exports = kind(
...
+
+/**
+* @type {String}
+* @translated loading message
+* @public
+* @usage in application
+* @var Spinner = require('moonstone/Spinner');
+* @this.$.spinner.set('accessibilityLabel', Spinner.LOADING_STRING);
+*/
+module.exports.LOADING_STRING = loadingString;
```
KunmyonChoi: I think you should use markdown syntax for the sample code and please reference below comment.
```
/**
* Translated loading message.
*
* Usage in application.
* ```
* var Spinner = require('moonstone/Spinner');
* this.$.spinner.set('accessibilityLabel', Spinner.LOADING_STRING);
* ```
*
* @type {String}
* @public
*/
```

KunmyonChoi:
LGTM!

PR #2938, opened 2016-11-18
PLAT-29721: Prevents scrolling to boundary when hidden

aarontam:
### Issue
When a `Scroller` is placed inside a control that can be hidden via a 5-way keypress, the 5-way keypress can trigger a `scrollToBoundary` which executes with a delay, and causes unnecessary scrolling while hidden. In the case of the `*.s` releases, we are muting Spotlight when scrolling, and thus there is a noticeable "hitch" from when the control is hidden to when the next control is spotted, due to the muting.

### Fix
It seems reasonable to prevent the `scrollToBoundary` job from executing if the control (or any of its ancestors) is hidden - we guard with a call to `getAbsoluteShowing`.

Issue: PLAT-29721


lishammel:
closing out as the jira associated with this is closed.

PR #2943, opened 2017-05-11
PLAT-40278 : Add background circle in Checkbox

changgilee:
Add background circle base on `18 GUI guidance in Enyo.

### DIFF
--- a/src/Checkbox/Checkbox.less
+++ b/src/Checkbox/Checkbox.less
@@ -22,6 +22,8 @@
 		padding: 0;
 		width: 30px;
 		height: 30px;
+		line-height: 30px;
+		font-size: 60px;
 		border-radius: 15px;
 		color: transparent;
 		background-color: @moon-dark-gray;
@@ -31,7 +33,7 @@
 		}
 	}
 	&[checked] .moon-icon {
-		color: @moon-white;
+		color: @moon-gray;
 
 		.moon-neutral & {
 			color: @moon-dark-gray;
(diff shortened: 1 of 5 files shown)


src/Checkbox/Checkbox.less
```diff
@@ -2,22 +2,74 @@
...
+		}
+	}
+
 	.moon-icon {
-		visibility: hidden;
 		margin: 0;
 		padding: 0;
+		width: 30px;
+		height: 30px;
+		border-radius: 15px;
```
yeram: Based on `enact` we also need `line-height` as `30px` and `font-size` as `60px`.

src/Checkbox/Checkbox.less
```diff
@@ -2,22 +2,73 @@
...
+			}
+		}
+	}
+
 	.moon-icon {
-		visibility: hidden;
 		margin: 0;
 		padding: 0;
+		width: 30px;
+		height: 30px;
```
Djspaceg: These would probably be better as a `@moon-checkbox-size` variable in the `variables.less` file

src/Checkbox/Checkbox.less
```diff
@@ -2,22 +2,73 @@
...
+		line-height: 30px;
+		font-size: 60px;
```
Djspaceg: This line could become `font-size: (@moon-checkbox-size * 2);`

src/Checkbox/Checkbox.less
```diff
@@ -2,22 +2,73 @@
...
+		border-radius: 15px;
```
Djspaceg: This line could become `font-size: (@moon-checkbox-size / 2);`

src/Checkbox/Checkbox.less
```diff
@@ -2,22 +2,73 @@
...
+		color: transparent;
+		background-color: @moon-dark-gray;
```
Djspaceg: It's best practice to not refer to these variables outside of the `moonstone-dark/light.less` files, and instead assign the `@moon-dark-gray` to a more descriptive color variable name, like `@moon-checkbox-bg-color: @moon-dark-gray;`. This is simpler and easier to maintain over time and abstracts the color value from the component LESS files. Be sure to update both the light and dark color files.

src/Checkbox/Checkbox.less
```diff
@@ -2,22 +2,73 @@
...
+			color: @moon-checkbox-toggle-switch-checked-color;
+		}
+		.moon-icon {
+			color: @moon-checkbox-toggle-switch-color;
+		}
+	}
+
+	.moon-checkbox:not([disabled]) {
+		.moon-icon {
+			background-color: rgba(255,255,255,0.41);
```
Djspaceg: This should be a color variable too. Also, do you think anyone can notice the difference between `0.41` and `0.4` alpha levels? Let's go ahead and round this value down to `0.4`: `rgba(255,255,255,0.4)`.
Alternatively, in the color file, you could use the LESS function to apply the alpha as well: `fade(@moon-white, 40%)`

Djspaceg (changes requested):
Let's make a few organizational changes to this PR.

src/Checkbox/Checkbox.less
```diff
@@ -2,22 +2,73 @@
...
+		}
+	}
+
+	.moon-checkbox:not([disabled]) {
+		.moon-icon {
+			background-color: fade(@moon-white, 40%);
+			color: transparent;
+		}
+		&[checked] .moon-icon {
+			color: @moon-white;
```
Djspaceg: Each of these `@moon-white` variables should be updated to be `@moon-spotlight-text-color`. This one, line 67, and 46.

PR #2946, opened 2017-06-20
PLAT-42382: Japanese multi-line text blocks line-break in the wrong places.

Sangwook-Lee:
### Issue Resolved / Feature Added

Japanese multi-line text blocks line-break in the wrong places.

### Resolution

Added a Japanese language mixin to apply the correct line-break rules to the following multiline components: moonstone/BodyText, moonstone/Dialog, moonstone/Notification, moonstone/Popup, and moonstone/Tooltip

### DIFF
--- a/css/moonstone-text.less
+++ b/css/moonstone-text.less
@@ -66,6 +66,7 @@
 	a:visited {color: @moon-spotlight-color; text-decoration:none;}
 	a:hover {color: @moon-spotlight-color; text-decoration:none;}
 	a:active {color: @moon-spotlight-color; text-decoration:none;}
+	.locale-japanese-line-break;
 }
 .moon-body-large-text {
 	font-size: @moon-body-font-size + 3;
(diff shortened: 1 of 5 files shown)


Djspaceg:
Are you sure `break-word` is a good idea? Wouldn't this cause layout changes in existing apps?

Sangwook-Lee:
I forcibly set the width style and mistook it as necessary. I agree with your opinion they don't need to add `break-word`. But is not it necessary for `Notification` ?

Djspaceg:
I don't think it's "necessary" anywhere. We don't have strict enough UI guidelines to necessitate using it. `break-word` literally breaks lines in the middle of words.

Sangwook-Lee:
As I know, `Notification` have being used `moon-word-break` but I don't know history that added `moon-word-break` to `Notification`. So, we worry that it will be a problem when that is removed.

Djspaceg:
Ah, that's right. It already had it, it should have it again, whether or not that's the "right" decision.

Sangwook-Lee:
Yes. So, I removed `word-break` except for `Notification`. Is other codes OK?

Djspaceg (approved):
Yes, the rest of this looks fine.

Sangwook-Lee:
Thank you for review! :)

PR #2949, opened 2017-07-21
PLAT-43388: Changes made by aria-valuetext

Baekwoo-s:
Updates due to aria-valutext behavior change

Enyo-DCO-1.1-Signed-off-by: Baekwoo Jung <baekwoo.jung@lge.com>


kbs12e:
LGTM

PR #2950, opened 2017-07-26
ENYO-4544: Fix bug of RadioItem

changgilee:
When selected:true is used static, RadioItem is not selected.


src/RadioItem/RadioItem.js
```diff
@@ -61,5 +60,9 @@ module.exports = kind(
 	decorateActivateEvent: function (sender, ev) {
 		ev.toggledControl = this;
 		ev.checked = this.selected;
+	},
+
+	initSelected: function() {
+		if(this.active) this.setSelected(this.active);
```
Sangwook-Lee: It would be nice if there's spaces.

PR #2954, opened 2017-09-18
ENYO-4581 String is cut off for 'sinhala' language in Input component.

Sangwook-Lee:
### Issue
String is cut off for 'sinhala' language in Input component.
We can see this issue when we set locale to latin.

### Fix
It needs to fix for Enyo as Enact was changed. (Migration)
So it's based on https://github.com/enyojs/enact/pull/1134.

ENYO-4581 Input: string is cut off for 'sinhala' language

### DIFF
--- a/src/Input/Input.less
+++ b/src/Input/Input.less
@@ -1,5 +1,8 @@
 /* Input.css */
 .moon-input {
+	line-height: 57px;
+	height: 57px;
+
 	padding: 0;
 	border: 0;
 	cursor: pointer;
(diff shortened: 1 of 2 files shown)


src/Input/Input.less
```diff
@@ -1,5 +1,8 @@
 /* Input.css */
 .moon-input {
+	line-height: 57px;
```
Sangwook-Lee, line 3: I think this 'line-height' can override value from 'moon-body-text'.

Djspaceg: It will depend on the precedence of the rules and the order of the classes, since this is a "top-level" class, just like `moon-body-text`. This code will need to fall after the selectors in `text.less`

PR #2955, opened 2017-09-21
ENYO-4727: Fix undefiend error

Baekwoo-s:
### Issue Resolved / Feature Added
Fixed an undefined error when rendering SimplePicker at runtime (create function)

### Resolution
Force aria-controls to be null with setAttribute at the timing when setAriaAttribute is not available.

### Links
ENYO-4727

### Comments
Enyo-DCO-1.1-Signed-off-by: Baekwoo Jung <baekwoo.jung@lge.com>


kbs12e:
LGTM

PR #2958, opened 2017-10-27
(Enyo) InputHeader: it needs to add exception from 'Input' pre-fix

Sangwook-Lee:
### Issue
InputHeader is cut off because Input height was changed.

### Fix
InputHeader shouldn't be applied changed height from Input.
So, I added hegiht: inherit; to Header.less to maintain original height size.

ENYO-4862 (Enyo) InputHeader: it needs to add exception from 'Input' pre-fix

### DIFF
--- a/src/Header/Header.less
+++ b/src/Header/Header.less
@@ -167,6 +167,7 @@
 				color: inherit;
 				width: 100%;
 				text-overflow: ellipsis;
+				height: inherit;
 
 				.input-placeholder({
 					color: @moon-input-header-placeholder-color;


Djspaceg:
I can't make a reliable estimate of what `height: inherit` will do on an `<input>` field within a non-statically sized element. :\ This will need to be tested to verify the consequences. The consequences should be localizable to the single component though. Just make sure that it doesn't alter the height of `Header` with `titleBelow` and `subTitleBelow` present.

Sangwook-Lee:
@Djspaceg  Thanks for review. We can test it in http://10.186.120.63/~sangwook/InputHeaderSample/dist/.
I think it'd be better to apply 'height' in input.less file except for 'input-header class' if that's possible. Is there any better solutions?

Djspaceg:
I'd need more information or testing to provide alternative ideas.

Sangwook-Lee:
@Djspaceg ,

Or, I think we can set "height" and "line-height" when parent is NOT "headerInput".

**Input.less**
`&:not(.moon-input-header-input-decorator) > .moon-input {
	line-height: 57px;
	height: 57px;
}`

PR #2962, opened 2017-11-17
ENYO-4921 Spinner: it doesn't set transparent property

Sangwook-Lee:
### Issue
I found that this issue occurs since 2.6.4-rc.15 by investigating when this occurred. The reason why this problem occurs is that {{background-color}} isn't changed to {{transparent}} even if transparent property is set.

### Fix
ball-decorator needs to be set background-color to 'transparent' when the spinner has transparent property.

ENYO-4921 Spinner: it doesn't set transparent property


lishammel:
Closing out as the jira is closed as will not fix.

